PR #5, opened 2015-07-24
README: Add build instructions

rnestler:
I'm not sure if qtcreator is really needed. maybe one can get along with qmake and make. Also I stole the screen-shot from http://stackoverflow.com/questions/8860712/setting-default-make-options-for-qt-creator maybe we want to create our own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 ## About LibrePCB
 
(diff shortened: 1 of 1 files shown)


ubruhin:
- building from command line is shown in .travis.yml
- maybe replace the screenshot with the link to stackoverflow?

rnestler:
So I updated and rebased the pull request:
 - show building from command line
 - replace screenshot with link to stackoverflow
Maybe we should also add a note about running the executable and tests?

Issue #21, opened 2015-08-21
Library design review/refactoring


ubruhin:
The actual library design may be a bit too complicated and may not scale very good.

branch to fix this issue: [feature/library_refactoring](https://github.com/LibrePCB/LibrePCB/tree/feature/library_refactoring)

ubruhin:
#### Examples of different library element types:

R/C/L:
- only a few different symbols
- many (!!) different footprints (mostly standard footprints, but also some special footprints)
- for each footprint only a few (or only one) 3D models
- very generic, mostly no exact device (part number) is required

Diodes/Transistors/OpAmps:
- only a few different symbols
- many (!!) different footprints (mostly standard footprints)
- for each footprint only a few (or only one) 3D models
- mostly an exact device (part number) is required (e.g. 1N4007, BC337, LM358N)

LEDs:
- only a few different symbols
- only a few different footprints (mostly standard footprints)
- for each footprint, there should be different 3D models to choose between different colors
- sometimes an exact device (part number) is required, sometimes not

Connectors:
- only a few different symbols
- many (!!) different footprints (some standard footprints, but mostly very special footprints)
- for each footprint only a few 3D models (some connectors are available in different colors)
- often an exact device (part number) is required

Special ICs (FT232, STM32F407, ...):
- mostly a special symbol is required for these devices
- only a few different footprints, sometimes the datasheet describes how they should look like (so specific footprints should be used instead of standard footprints)
- for each footprint only a few (or only one) 3D models

rnestler:
Could you also create a pull-request from the branch? Makes discussing the diff easier.

ubruhin:
see pull request #22

ubruhin:
how to provide / where to save...
- ...suppliers + part numbers required to order components?
- ...prices of components (for each supplier)?
- ...datasheets of components?

ubruhin:
how to support different footprints for different soldering processes (e.g. iron, reflow, wave)?
- use separate footprints/packages with different UUIDs?
- support different soldering processes in one footprint/package?

And how to select the preferred soldering process in the board of a project?

ubruhin:
One major issue is the file size of library elements:
 * XML files are not optimized for small file size
 * 3D models require lot (!!) of disk space
 * SPICE models require lot of disk space
 * Because of the backward compatibility (copies of older versions), the disk space required by library elements will grow continuously with the time

So, the required disk space of a workspace library will explode with (hundreds of) thousands of library elements.

[...]

PR #22, opened 2015-08-25
library refactoring

ubruhin:
Only to discuss, feature is not yet ready to merge

See issue #21


ubruhin:
**At the moment, there are two slightly different proposals for a new library structure:**

|  | Proposal 1 | Proposal 2 |
|-------------------|------------|------------|
| Structure Diagram | branch [library_refactoring](https://github.com/LibrePCB/LibrePCB/blob/feature/library_refactoring/dev/diagrams/svg/library_structure.svg) | branch [library_refactoring_2](https://github.com/LibrePCB/LibrePCB/blob/feature/library_refactoring_2/dev/diagrams/svg/library_structure.svg) |
| Examples Diagram | branch [library_refactoring](https://github.com/LibrePCB/LibrePCB/blob/feature/library_refactoring/dev/diagrams/svg/library_structure_examples.svg) | branch [library_refactoring_2](https://github.com/LibrePCB/LibrePCB/blob/feature/library_refactoring_2/dev/diagrams/svg/library_structure_examples.svg) |
| Description | Devices can have different Packages (see ```Device::PackageVariant```) | Each Device has exactly one Package (no ```Device::PackageVariant```) |

**Advantages of Proposal 1:**
* A device represents (more or less) a whole "device family" - to provide different packages for one component, you only need to create one device (with different packages)
 * Example: A device "LM358" could be available with "DIP-8" or "SOIC-8" package
* As the device represents a whole device family, a SPICE model of that device family needs to be added only to this device (saves disk space)
* In the schematic, it would be possible to specify the device family (e.g. "LM358") without specifying the exact package. The exact package can be different in each board.

**Disadvantages of Proposal 1:**
* More complicated library structure *(too complicated?)*
* Leads to big device XML files as they can contain many different packages
 * Projects will require more disk space as there are many device package variants which are not needed

***Advantages of Proposal 2:***
* The library structure is less complicated!
* Smaller (but more) XML files for devices
 * Example: Instead of one huge "Resistor" device with all possible packages, there would be many small "Resistor" devices (one for each different package)

**Disadvantages of Proposal 2:**
* For each package of a device family (e.g. "LM358"), a separate device must be created, and there is no relation between them
* A SPICE model which is valid for different packages of a device needs to be added to each device separately (so, the same SPICE model exists in different devices --> needs more disk space)

ubruhin:
we will not implement the library design of this branch

PR #32, opened 2015-11-08
Use more restrictive compiler flags

ubruhin:
I'm not yet sure if this would make sense, but I have created a branch where more restrictive compiler flags are used (which should help to increase code quality).

Added compiler flags: `-pedantic` and `-Weffc++` (only known by GCC)

In release mode, `-Werror` is also added (creating releases should not be possible as long as there are compiler warnings). In the Travis-CI config, I have extended the build matrix to always build in debug and release mode. This means that a build is only successful if there are no compiler warnings.

At the moment, `librepcbcommon` is the only project `g++` compiles without warnings. All other projects have to be adjusted before merging this pull request...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ install:
 build_script:
   - mkdir build
   - cd build
-  - qmake -r ..\librepcb.pro
+  - qmake -r ..\librepcb.pro QMAKE_CFLAGS="-Werror" QMAKE_CXXFLAGS="-Werror"
   - mingw32-make -j 4
   - .\generated\windows\qztest.exe # run quazip unit tests
   - .\generated\windows\tests.exe  # run librepcb unit tests
--- a/common.pri
+++ b/common.pri
@@ -43,8 +43,8 @@ CONFIG += c++11
 
 # enable compiler warnings
 CONFIG += warn_on
-QMAKE_CXXFLAGS += -Wextra
-QMAKE_CXXFLAGS_DEBUG += -Wextra
+QMAKE_CXXFLAGS += -Wextra -pedantic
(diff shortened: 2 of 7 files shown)


rnestler:
Setting `-Werror` is bad for packagers. Imagine your code compiles without warnings on your system, but the system of the packager has a newer compiler version which adds a warning. He would have to patch the Makefiles to work around that.

ubruhin:
You're right... So probably it would be better to use ```-Werror``` only for Travis-CI builds...

3rdparty/gmock/gmock.pro
```diff
@@ -13,12 +13,12 @@ GENERATED_DIR = ../../generated
...
 include(../../common.pri)
 
+# compile gmock as static library
 CONFIG -= qt app_bundle
 CONFIG += staticlib thread
 
 # suppress compiler warnings
-QMAKE_CXXFLAGS += -Wno-missing-field-initializers -Wno-unused-variable -Wno-missing-field-initializers
-QMAKE_CXXFLAGS_DEBUG += -Wno-missing-field-initializers -Wno-unused-variable -Wno-missing-field-initializers
+CONFIG += warn_off
```
dns2utf8: Why not start with `-Wall`?

ubruhin: Have you seen the "no" after "-W"? :)
The purpose of these lines is to **disable** some warnings, because it's a 3rd-party library where I won't fix compiler warnings...

dns2utf8: ah, I missed that part :sweat_smile:

ubruhin:
This now much simplified and replaced by #183.

PR #44, opened 2016-01-09
added readme file to each subdirectory

ubruhin:
I think this helps to understand the directory structure of the whole project. In my opinion it's very useful to see the architecture overview diagram when [browsing this repository in the browser](https://github.com/LibrePCB/LibrePCB/tree/add_subdir_readme_files/libs).

@0xB767B I tried to rename the "dev" directory but I didn't like it :wink: With the readme file it should be clear now what's the content of this directory, especially when [watching it in the browser](https://github.com/LibrePCB/LibrePCB/tree/add_subdir_readme_files/dev).


rnestler:
:+1: Looks nice indeed.
Regarding dev... Why is the Doxygen stuff in dev? Shouldn't that belong to doc? Or is doc for user documentation only?

ubruhin:
> Shouldn't that belong to doc? Or is doc for user documentation only?

The [new readme in this directory](https://github.com/LibrePCB/LibrePCB/tree/add_subdir_readme_files/doc) should answer your question :)

PR #46, opened 2016-01-18
Add schematic editor activity diagrams

ubruhin:
@0xB767B Is this the type of diagram you want to have?

diagram

### DIFF
--- a/dev/diagrams/README.md
+++ b/dev/diagrams/README.md
@@ -1,0 +1,7 @@
+# Development Diagrams
+
+All diagrams are created with [Dia](http://live.gnome.org/Dia).
+
+Please save all projects uncompressed (file -> save as -> uncheck "compress file")!
+
+The bash script [convert.sh](./convert.sh) converts all diagrams to SVG and saves them in the directory [svg](./svg).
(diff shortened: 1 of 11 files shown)


0xB767B:
Yes, that's what I had in mind. But I would avoid having "and" in a block but rather use two blocks. It makes the distinct steps better visible. Furthermore it helps structuring the diagram better.

ubruhin:
> But I would avoid having "and" in a block but rather use two blocks.

Ack! I have updated the diagram accordingly.

0xB767B:
Thumbs up!

These activity diagrams build a great foundation for detailed description for activity/flow in tools. Nevertheless I think we cannot avoid having an overall/general specification of the user interface. E.g. your magic right-click ;-) This might be expressed in activity diagrams as well. Or maybe just in pure prose ;-)

0xB767B:
Btw: Title of issue is missleading. We are talking about deinition of workflow/usage. FSM is a implementatio detail.

ubruhin:
> Title of issue is missleading

fixed.

ubruhin:
Now there is also an activity overview diagram/description. Hope this helps...

diagram

ubruhin:
Added new diagram:

diagram

Issue #50, opened 2016-01-26
GDB does not work in the Docker container


ubruhin:
When starting the debugger in qtcreator, following messages appear and the debugger doesn't work:

![warning_010](https://cloud.githubusercontent.com/assets/5374821/12593620/4f0ca7a2-c472-11e5-88cd-4c3d1e162f29.png)

![gdb not responding_006](https://cloud.githubusercontent.com/assets/5374821/12591513/6611b884-c468-11e5-82eb-fe32ed033c7e.png)

ubruhin:
With #51 GDB is now working in the docker container.

The warning "This does not seem to be a Debug build" disappears after disabling the checkbox "Enable QML Debugging" in the project settings. It would be nice to fix that issue someday, but as GDB is now working properly, I still close this issue now.

PR #51, opened 2016-02-04
Fix GDB issues in docker container

ubruhin:
- The flag "--privileged" seems to fix parts of #50 
- Splitting up "docker run" into "docker create" and "docker start" allows to make changes in the container persistent between sessions

@0xB767B What do you think about this?


0xB767B:
Works now. Clear to merge!

PR #54, opened 2016-02-22
Implement board traces & vias

ubruhin:
(not yet ready to merge)


ubruhin:
Adding Vias is now possible! :smiley: 

auswahl_013

Implementing traces is still in progress...

ubruhin:
Drawing traces is now possible! But there is still lot of work to do...

auswahl_017

Issue #55, opened 2016-03-03
Create a scope guard class


ubruhin:
It would be very useful to have a scope guard class which makes it possible to execute multiple methods, which can throw exceptions, atomically. If one method throws an exception, all previously executed methods must be reverted.

How such a scope guard may be used (...not sure if the lambdas are properly used...?):
```C++
void MyClass::doSomethingAtomically(int magicNumber) throw (Exception)
{
    ScopeGuardList sgl;
    sgl.execute([this](){doSomethingOrThrow();},
                [this](){revertSomethingOrThrow();});
    sgl.execute([this, &magicNumber](int n){makeMagicThingsOrThrow(n);},
                [this, &magicNumber](int n){revertMagicThingsOrThrow(n);});
    sgl.execute([this](){fooOrThrow();},
                [this](){barOrThrow();});
    sgl.commit();
}
```

Such a class should be created in the static library [librepcbcommon](https://github.com/LibrePCB/LibrePCB/tree/master/libs/librepcbcommon) (namespace "librepcb") because it would be useful for all other libraries.

@rnestler Interested in implementing such a class? :smiley:

rnestler:
Definitely interested! Feel free to assign me to the issue :smiley:

PR #57, opened 2016-03-07
Add ScopeGuard implementation

rnestler:
- [x] Cleanup commit messages
 - [x] Add ASCII art according to the project style
 - [x] Add some documentation
 - [x] ScopeGuardList should call cleanup actions in reverse order

Inspired by
https://channel9.msdn.com/Shows/Going+Deep/C-and-Beyond-2012-Andrei-Alexandrescu-Systematic-Error-Handling-in-C

### DIFF
--- a/libs/librepcbcommon/scopeguard.h
+++ b/libs/librepcbcommon/scopeguard.h
@@ -1,0 +1,131 @@
+/*
+ * LibrePCB - Professional EDA for everyone!
+ * Copyright (C) 2016 The LibrePCB developers
+ * http://librepcb.org/
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef LIBREPCB_SCOPEGUARD_H
+#define LIBREPCB_SCOPEGUARD_H
+
+/*****************************************************************************************
+ *  Includes
+ ****************************************************************************************/
+#include <QtCore>
+#include <utility>
+
+/*****************************************************************************************
+ *  Namespace / Forward Declarations
+ ****************************************************************************************/
+namespace librepcb {
+
+/*****************************************************************************************
+ *  Class ScopeGuardBase
+ ****************************************************************************************/
+/**
+ * Base class for ScopGuard mainly providing dismiss()
+ */
+class ScopeGuardBase
+{
+    public:
+
+        ScopeGuardBase() noexcept :
+            mActive(true)
+        { }
+
+        ScopeGuardBase(ScopeGuardBase&& rhs) noexcept :
+            mActive(rhs.mActive)
+        {
+            rhs.dismiss();
+        }
+
+        /**
+         * Do not execute cleanup code
+         */
+        void dismiss() noexcept
+        {
+            mActive = false;
+        }
+
+    protected:
+
+        /**
+         * Do not allow to delete a ScopeGuardBase directly. So we can avoid a
+         * virtual destructor.
+         */
+        ~ScopeGuardBase() = default;
+        bool mActive;
+};
+
+/*****************************************************************************************
+ *  Class ScopeGuard
+ ****************************************************************************************/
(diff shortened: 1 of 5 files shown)


libs/librepcbcommon/scopeguard.h
```diff
@@ -0,0 +1,75 @@
...
+
+#ifndef LIBREPCB_SCOPEGUARD_H
+#define LIBREPCB_SCOPEGUARD_H
+
+#include <utility>
+
+namespace librepcb {
+
+template<class Fun>
+class ScopeGuard
```
ubruhin: If no method is declared virtual (not even the destructor), please use the "final" keyword to the class definition

libs/librepcbcommon/scopeguard.h
```diff
@@ -0,0 +1,75 @@
...
+
+        ScopeGuard(Fun f) :
+            mF(std::move(f)),
+            mActive(true)
+        { }
+
+        ~ScopeGuard()
+        {
+            if (mActive) {
+                mF();
```
ubruhin: if mF() throws an exception, we would have a big problem... Maybe we should add a try/catch block, and in the catch block we simple call qFatal("ScopeGuard: Internal Fatal Error") which will let the application crash ;)

ubruhin:
Block comment would be nice :smile: 
```c++
/*****************************************************************************************
 *  Foo
 ****************************************************************************************/
```

libs/librepcbcommon/scopeguardlist.h
```diff
@@ -0,0 +1,66 @@
...
+            mScopeGuards(std::move(rhs.mScopeGuards))
+        { }
+
+        ScopeGuardList(const ScopeGuardList&) = delete;
+        ScopeGuardList& operator=(const ScopeGuardList&) = delete;
+
+        ~ScopeGuardList()
+        {
+            if (mActive) {
+                for (auto& scopeGuard : mScopeGuards) {
```
rnestler: I think the handlers should be called in reverse order, to match the behavior of multiple local `ScopeGuard` variables.

ubruhin:
Btw, you have referenced the wrong issue in the commit message ;) #55 would be the right one...

ubruhin:
Great, thanks! :+1:

ubruhin:
Oooh I've found a bug! When using the ```ScopeGuardList(size_t)``` constructor, the vector gets filled with default constructed ```std::function<void()>``` functions instead of only reserving the requested memory. This will throw a ```bad_function_call``` exception when calling such a default constructed function.

@rnestler Could you fix that? 

And in addition, it would be nice to output the catched exception in the destructor of ```ScopeGuardList``` (and also in the destructor of ```ScopeGuard```):
```C++
try{ (*scopeGuard)(); } catch(std::exception& e) {
    qFatal("Cleanup function threw an exception: %s", e.what());
}
```

rnestler:
> the vector gets filled with default constructed std::function<void()> functions

Ohh sorry my bad. Of course. I only need to call `reserve()`

rnestler:
Hmm btw. qAbort isn't so nice in the unit tests: 
```
[==========] Running 101 tests from 3 test cases.
[----------] Global test environment set-up.
[----------] 9 tests from ScopeGuardTest
[ RUN      ] ScopeGuardTest.testScopeGuard
[       OK ] ScopeGuardTest.testScopeGuard (0 ms)
[ RUN      ] ScopeGuardTest.testScopeGuardDismiss
[       OK ] ScopeGuardTest.testScopeGuardDismiss (0 ms)
[ RUN      ] ScopeGuardTest.testScopeGuardPerformance
Needed 0.0179434s for 10000000 loops
[       OK ] ScopeGuardTest.testScopeGuardPerformance (18 ms)
[ RUN      ] ScopeGuardTest.testScopeGuardList
[       OK ] ScopeGuardTest.testScopeGuardList (0 ms)
[ RUN      ] ScopeGuardTest.testScopeGuardListDismiss
[       OK ] ScopeGuardTest.testScopeGuardListDismiss (0 ms)
[ RUN      ] ScopeGuardTest.testScopeGuardListOrder
[       OK ] ScopeGuardTest.testScopeGuardListOrder (0 ms)
[ RUN      ] ScopeGuardTest.testScopeGuardListSizedConstructor

Program received signal SIGABRT, Aborted.
0x00007ffff6adc2a8 in raise () from /usr/lib/libc.so.6
(gdb) c
Continuing.

Program terminated with signal SIGABRT, Aborted.
The program no longer exists.
```

ubruhin:
> Is every exception used in LibrePCB derived from std::exception?

Yes, they are. Oh now I have found a bug in the implementation of the method ```what()```, I will fix that later ;)

> Hmm btw. qAbort isn't so nice in the unit tests:

:smile: This method is handled in [libs/librepcbcommon/debug.cpp](libs/librepcbcommon/debug.cpp) but we could change its behaviour when running the unit tests...

rnestler:
Btw. I opened #62 and hopefully fixed it in #61.

PR #59, opened 2016-03-13
Improve project core classes

ubruhin:
No change to understand all the things I have done in that huge commit, but it's a very important commit :smile: 

Thanks to many new [high level undo commands](https://github.com/LibrePCB/LibrePCB/tree/improve_project_core_classes/libs/librepcbprojecteditor/cmd), the schematic editor FSM has now much less code and works much better!

@nemofisch @0xB767B 
auswahl_016


rnestler:
Holy crap! That are a lot of changes...

PR #60, opened 2016-03-16
Use the new ScopeGuard/ScopeGuardList classes

ubruhin:
@rnestler It seems that the new classes work great! Thanks to these classes I was able to remove lot of code which is no longer required :smiley:


rnestler:
Nice :smiley: does it work like expected?

PR #61, opened 2016-03-16
ScopeGuardList: Added unit tests and fix bug

rnestler:
The entries should get executed in reverse order.


rnestler:
@ubruhin About the third commit: Don't you think past tense looks a bit silly there? 
> ScopeGuard: Showed which exception caused the abort

ubruhin:
> ScopeGuard: Showed which exception caused the abort

maybe "ScopeGuard: Added exception message to the qFatal() output" ;)

Good Night! :sleeping:

Issue #63, opened 2016-03-19
Travis CI: Mac builds are broken


ubruhin:
No idea why, but mac builds are failing with this error message:
```Project ERROR: Unknown module(s) in QT: webkitwidgets```

ubruhin:
As I don't have a mac, I can't reproduce this error locally. It would be great if a mac user could fix this issue ;)

Until this issue is solved, mac builds are allowed to fail (see #66).

ubruhin:
@HBadertscher Maybe you could fix this issue? :smile:

HBadertscher:
I'll have to come over to Coredump some time, probably next monday. I will then look into this issue, ok?

ubruhin:
Fantastic! :+1:

PR #65, opened 2016-03-20
Add possibility to highlight all items of a specific net signal

ubruhin:
When drawing a wire in the schematic editor, all other lines/pins of the same net signal will now be highlighted.


rnestler:
You did notice that it failed CI before you merged it?

ubruhin:
>You did notice that it failed CI before you merged it?

Yes, but this is caused by #63, not by wrong code...

PR #77, opened 2016-06-12
travis: Fix OS X builds

The-Compiler:
Uses a custom Qt bottle (binary package) built by me for my own project (qutebrowser).


ubruhin:
Very nice, thanks!

Fixes #63

PR #86, opened 2016-06-30
Travis: Update Qt bottle

The-Compiler:
From upstream 5.6.1 to upstream 5.6.1-1

Not 100% sure if it'll work, for qutebrowser I can't load QtWebKit, but I might need to rebuild PyQt5... let's see if it works here!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ install:
       sudo apt-get install -qq libglu1-mesa-dev
     elif [ "${TRAVIS_OS_NAME}" = "osx" ]
     then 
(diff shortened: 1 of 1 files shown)


The-Compiler:
Nope, the bottle is broken...

I won't have time in the near future to take care of this, so I'm considering not testing stuff on OS X for qutebrowser until it works without QtWebKit... I'd suggest the same for LibrePCB, it's just too much of a pain currently :wink:

ubruhin:
OK no problem, thanks anyway ;)

Issue #79, opened 2016-06-12
Get rid of QtWebKit


The-Compiler:
QtWebKit got removed upstream with Qt 5.6 - if you still need a HTML viewer after #72 I'd recommend either QtWebKit or QtWebEngine via a compile-time flag, or using something more minimalistic like [QTextBrowser](http://doc.qt.io/qt-5/qtextbrowser.html) which has some degree of HTML support.

ubruhin:
Merged #87

PR #87, opened 2016-07-03
Use markdown for project descriptions (instead of HTML)

### LINKED ISSUE #72: Use markdown for project description (instead of HTML)
The QWebView in the ControlPanel is overkill and HTML is not very readable. A markdown viewer would be much better and markdown files are more readable...

ubruhin:
- [Hoedown](https://github.com/hoedown/hoedown) is used to convert the README.md into HTML
- [QTextBrowser](http://doc.qt.io/qt-5/qtextbrowser.html) is used to display the HTML in the ControlPanel

Fixes #72 and #79.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
 	path = dev/demo-workspace
 	url = https://github.com/LibrePCB/demo-workspace.git
 	branch = library_refactoring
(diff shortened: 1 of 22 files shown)


dbrgn:
Hoedown is cool.

Issue #17, opened 2015-08-03
Create a "About LibrePCB" dialog


ubruhin:
The current about dialog is very simple and could be improved in many ways:
- [x] Replace the `QMessageBox` with an own class derived from `QDialog`. This new class should be called `AboutDialog` and created in the directory [`libs/librepcb/common/dialogs/`](https://github.com/LibrePCB/LibrePCB/tree/master/libs/librepcb/common/dialogs)
- [x] Add the new about dialog to the "Help" menu of all main windows (ControlPanel, SchematicEditor, BoardEditor)
- [ ] Add more information to the new dialog, maybe splitted up in multiple tabs:
  - [x] A small description about LibrePCB (inclusive Logo)
  - [x] A link to LibrePCB's website (librepcb.org) and to the GitHub project site
  - [x] The application's version / commit hash (determined by `git describe`)
  - [x] Maybe the datetime of the last commit
  - [x] The license (a link to http://www.gnu.org/licenses/gpl-3.0.html, or the whole license text)
  - [x] Options for contribution (Bitcoin address, link to Patreon)
  - [x] Maybe a "how to contribute" section (see also #12)
  - [ ] Maybe the rendered content of the [AUTHORS.md](https://github.com/LibrePCB/LibrePCB/blob/master/AUTHORS.md) file

dbrgn:
Since PR #213 was closed, I plan to give this another attempt with classical UI 🙂

ubruhin:
Fixed in #261

PR #88, opened 2016-07-04
Show git revision in about dialog

rnestler:
See #17.

### DIFF
--- a/librepcb/controlpanel/controlpanel.cpp
+++ b/librepcb/controlpanel/controlpanel.cpp
@@ -294,9 +294,11 @@ void ControlPanel::projectEditorClosed() noexcept
 
 void ControlPanel::on_actionAbout_triggered()
 {
-    QMessageBox::about(this, tr("About"), tr(
-        "LibrePCB is a free & open source schematic/layout-editor."
-        "<br><br>"
+    QMessageBox::about(this, tr("About LibrePCB"), tr(
+        "<h1>About LibrePCB</h1>"
+        "<p>LibrePCB is a free & open source schematic/layout-editor.</p>"
+        "<p>Version: " GIT_VERSION "</p>"
+        "<p>Please see <a href='http://librepcb.org/'>librepcb.org</a> for more information.</p>"
         "You can find the project on GitHub:<br>"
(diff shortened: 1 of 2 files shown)


rnestler:
Is this mergable as is or do I need to change something?

librepcb/controlpanel/controlpanel.cpp
```diff
@@ -294,9 +294,11 @@ void ControlPanel::projectEditorClosed() noexcept
...
 void ControlPanel::on_actionAbout_triggered()
 {
-    QMessageBox::about(this, tr("About"), tr(
-        "LibrePCB is a free & open source schematic/layout-editor."
-        "<br><br>"
+    QMessageBox::about(this, tr("About LibrePCB"), tr(
+        "<h1>About LibrePCB</h1>"
+        "<p>LibrePCB is a free & open source schematic/layout-editor.</p>"
+        "<p>Version: " GIT_VERSION "</p>"
+        "<p>Please see <a href='https://librepcb.org/'>librepcb.org</a> for more information.</p>"
```
ubruhin: Unfortunately librepcb.org does not support https, so the link does not work...

rnestler: Ohh. Then get https working :wink:

ubruhin: Will you pay the SSL certificate? ;)

dbrgn: - Free certificates: https://startssl.com/
- 4$ certificates: https://gogetssl.com/domain-validation/domain-ssl/
- ...and many more.

Github pages doesn't support certificates on custom domains though. I could host the content & certificate for you, but then we wouldn't have auto-updating on push anymore.

ubruhin: I also have a webspace, that's not the problem. But I don't have the time and nerves to set up the certificate, and even less to set up an auto-updating mechanism...

But as the website contains only static content and thus no sensitive data is transmitted, I would say SSL is not that important (at least until now). There are many other issues which are much more important ;)

Issue #93, opened 2016-07-07
Replace private marked copy/default constructors with =delete


rnestler:
In some places in the code the following is found:

``` c++
private:
        // make some methods inaccessible...
        ErcMsgList();
        ErcMsgList(const ErcMsgList& other);
        ErcMsgList& operator=(const ErcMsgList& rhs);
```

Which should be replaced by

``` cpp
public:
        ErcMsgList() = delete;
        ErcMsgList(const ErcMsgList& other) = delete;
        ErcMsgList& operator=(const ErcMsgList& rhs) = delete;
```

The canonical order for the constructors is like follows:

``` cpp
public:
  // Constructors / Destructor
  [Default Constructor]
  [Copy Constructor]
  [Custom Constructors]
  [Destructor]

  // getters, setters, ...

  // Operator Overloadings
  [Assignment Operator]
  [Other Operators]
```

I'm willing to mentor this bug if someone wants to work on it. The following are some of the places were this could be fixed:
- [ ] https://github.com/LibrePCB/LibrePCB/blob/dbc6805ab59c76f2f859ef72f00f24ba047ee2c7/apps/librepcb/controlpanel/controlpanel.h#L124
- [ ] https://github.com/LibrePCB/LibrePCB/blob/340c1bc32e83b2a9852de574c0feec0ff020776d/libs/librepcbworkspace/recentprojectsmodel.h#L59
- [ ] ...

Just search for [`// make some methods inaccessible...`](https://github.com/LibrePCB/LibrePCB/search?utf8=%E2%9C%93&q=make+some+methods+inaccessible&type=Code)

ubruhin:
Yeah, this issue exists in many header files because in the past I have always marked inaccessible methods ```private```. Later I switched to ```delete```, which I always use now when creating new classes...

ali-alsabbah:
I'd like to work on this if mentorship is still available

ubruhin:
> I'd like to work on this if mentorship is still available

Sure, go for it! :smiley:  Just ask either @rnestler or me if you have any questions.

rnestler:
> I'd like to work on this if mentorship is still available

@ali-alsabbah Are you still interested in this? If yes jut tell us how we can support you achieving it :slightly_smiling_face:

rzvncj:
Just browsing here, but wanted to point out that
```c++
public:
        ErcMsgList() = delete;
```
[...]

PR #94, opened 2016-07-07
Cleanup ErcMsgList

rnestler:
See #93

While looking through the code base to get familiar with it I will fix some of those.

### DIFF
--- a/libs/librepcbproject/erc/ercmsglist.h
+++ b/libs/librepcbproject/erc/ercmsglist.h
@@ -96,7 +96,7 @@ class ErcMsgList final : public QObject, public IF_XmlSerializableObject
 
         // File "core/erc.xml"
         FilePath mXmlFilepath;
-        SmartXmlFile* mXmlFile;
+        QScopedPointer<SmartXmlFile> mXmlFile;
 
         // Misc
         QList<ErcMsg*> mItems; ///< contains all visible ERC messages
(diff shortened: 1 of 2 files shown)


ubruhin:
I always use the same order of these methods, so I would consider it as our guidelines:

```c++
public:
  // Constructors / Destructor
  [Default Constructor]
  [Copy Constructor]
  [Custom Constructors]
  [Destructor]

  // getters, setters, ...

  // Operator Overloadings
  [Assignment Operator]
  [Other Operators]
```
Example: [netclass.h](https://github.com/rnestler/LibrePCB/blob/cb13a80110c0516c140bad3ce909cfb65733a3b7/libs/librepcbproject/circuit/netclass.h)

rnestler:
> I always use the same order of these methods, so I would consider it as our guidelines:

I don't like the guideline that much since I consider the copy constructor, assignment operator and destructar as a unit which belongs together. See [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming)).

But consistency is more important.

libs/librepcbproject/erc/ercmsglist.h
```diff
@@ -54,9 +54,12 @@ class ErcMsgList final : public QObject, public IF_XmlSerializableObject
     public:
 
         // Constructors / Destructor
+        ErcMsgList() = delete;
+        ErcMsgList(const ErcMsgList& other) = delete;
         explicit ErcMsgList(Project& project, bool restore, bool readOnly, bool create) throw (Exception);
         ~ErcMsgList() noexcept;
 
+
```
rnestler: Should I remove this empty line?

ubruhin: It would be perfect if you moved this empty line from here to above the "signals:" statement :)

rnestler: Done

PR #104, opened 2016-08-09
Add credits section to readme

ubruhin:
Because [cloudscale.ch](https://www.cloudscale.ch/) is our first sponsor, I think we should now add a "credits" section to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,11 @@ To install LibrePCB on a Linux/UNIX system, just execute the following command a
 sudo make install
 ```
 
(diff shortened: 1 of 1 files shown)


dbrgn:
Looks good! I would put their logo in the README and/or in the library fetching UI though.

ubruhin:
> I would put their logo in the README and/or in the library fetching UI though.

Yes, I think I will put it either on our website and/or the library fetching UI, but in the README I prefer a simple link...

Issue #16, opened 2015-08-03
Add a "new project"-wizard to ControlPanel


ubruhin:
Create new projects with the ControlPanel: 

Instead of only opening a QFileDialog to select the new project's filepath, we should show a simple wizard to choose the following things:
- [x] the project's name
- [x] the project's author
- [x] target directory (where to create the project)
- [ ] target filename (*.lpp) --> automatically generated from the project's name, but editable
- [x] the project's license (optional)
- [ ] some project settings, like language, norm, ... (optional)
- [x] name of the first schematic and board to add (optional)
- [x] which version control system to use (optional)

ubruhin:
I think the `NewProjectWizard` from #106 is pretty nice, so there is no need for further improvements for milestone 0.1 and thus I'll close this issue.

PR #106, opened 2016-08-14
Add a "new project"-wizard

ubruhin:
### Description
This wizard makes it easier to create a new project, and it allows to specify some additional information how the project should be created. This fixes (at least parts of) #16.

### Issues
- The project's directory name and basename of the \*.lpp project file are automatically derived from the project's name (spaces are replaced by underscores, invalid characters are removed). But I'm not sure if this is the best way, because there is no way to choose another directory-/filename. On the other side, this does not confuse the user with too many input fields... **Any thoughts about this?**
- Initializing a Git repository is not yet possible. But as the files .gitignore and .gitattributes will already be copied into the project's directory, I think this is enough for the moment.

### Screenshots
create new project_004
create new project_002
create new project_003

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ install:
   - |
     if [ "${TRAVIS_OS_NAME}" = "linux" ]
     then
-      sudo apt-get install -qq libglu1-mesa-dev
+      sudo apt-get install -qq libglu1-mesa-dev xvfb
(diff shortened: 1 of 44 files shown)


dbrgn:
:tada:

ubruhin:
I think this is ready for merge. The two issues from above may be fixed later (depending on user feedback)...

Issue #14, opened 2015-07-30
Make FileLock working properly


ubruhin:
There is a TODO (documented with doxygen) in the method FileLock::getStatus() which should be fixed: https://github.com/LibrePCB/LibrePCB/blob/master/libs/librepcbcommon/fileio/filelock.cpp

So we need to check if there is another LibrePCB application instance with a specific PID running on the same computer. Maybe this could be done by a new method in the class SystemInfo (resp. 3 different methods for linux, mac and windows).

PR #113, opened 2016-08-21
Refactor/rename common class FileLock

ubruhin:
- Rename class from FileLock to DirectoryLock
- Lock whole directories instead of single files
- Change lock file basename from ".~lock.FILENAME" to ".lock"
- Change separator in lock files from comma (',') to newline ('\n')
- Make private constructors/operators public and use "=delete" instead
- Use helper class "FileUtils" to read/write/remove lock files (removes redundant code)
- Add unit tests


libs/librepcbcommon/fileio/directorylock.cpp
```diff
@@ -34,114 +35,85 @@ namespace librepcb {
...
 DirectoryLock::~DirectoryLock() noexcept
 {
-    if (mLockedByThisObject)
-        try{unlock();} catch (...) {}
+    if (mLockedByThisObject) {
+        try {
+            unlock(); // can throw
+        } catch (const Exception& e) {
+            qCritical() << "Could not remove lock file:" << e.getUserMsg();
+        }
```
rnestler: What happens if another exception than `Exception` gets thrown?

rnestler:
Is there any reason why you don't use [`QLockFile`](https://doc.qt.io/qt-5/qlockfile.html)?

libs/librepcbcommon/fileio/directorylock.cpp
ubruhin: There aren't any other exceptions, all exceptions used in LibrePCB must be derived from `Exception`. If a 3rdparty library would throw other exceptions, we would need to wrap them.

ubruhin:
> Is there any reason why you don't use QLockFile?

There are several small reasons why I decided to implement it by myself:
- `QLockFile` does not provide a method to check whether a lock is stale or not. We would need to get the lock information with `getLockInfo` and compare the PID, hostname and application name with the currently running application. My own class is much simpler to use because this logic is already integrated, the user just calls `getStatus()` and knows if the resource is either locked, unlocked or stale locked.
- The exception handling of `QLockFile` seems to be quite bad. At least the method `unlock()` does not report if it was successful or not. In my own class I use our exceptions like everywhere else in the project.
- I have found no documentation about the content of the lock file, and if it could change with future Qt versions (which could lead into problems when a lock file needs to be accessed with different LibrePCB applications at the same time). When defining the lock file format by myself, I can make sure to keep backward/forward compatibility.

ubruhin:
One big issue of the `DirectoryLock` (formerly `FileLock`) class is that stale locks (locks of crashed application instances) are still not working (because not implemented). To detect stale locks, we need to detect if the application instance, which has created the lock, is still running. But how to implement that?

### Proposal 1
When obtaining a lock, the application's PID and process start time is written into the lock file. To determine whether a lock is stale or not, one could check if a process with the PID and start time of the lock file is still running.

**Advantages:**
- A stale lock will be detected immediately. Right after the application, which holds the lock, crashes, querying process information will immediately show that the lock must be stale. No timeout or something similar.

**Disadvantages:**
- Querying information about running processes is quite painful. Different operating systems provide different APIs to do that, and especially on OSX this seems to be very hard (I can't find any useful information/documentation in the internet).
- In addition, querying information about running processes might be restricted by the operating system. When running LibrePCB in a (kind of) sandbox which hides other running processes on the system, detection of stale file locks will not be possible.
- A stale lock which was created with another username/hostname is not possible, because we cannot query processes on another computer. This might be relevant if the files to lock are located on a network drive, where applications from different computers can access them.

### Proposal 2
The datetime when a lock was obtained is written into the lock file. As long as the lock is active, this datetime will be regularly updated, for example with an interval of 1 minute. A lock is considered as stale if this datetime is older than 1 minute.

**Advantages:**
- Quite simple to implement.
- Very robust, would work properly even on network drives. *Or maybe not --> what if the system time of the computers in the network differs from each other?*

**Disadvantages:**
- Many file accesses (every minute the lock file needs to be overwritten).
- A stale lock cannot be detected immediately. So when LibrePCB crashes, the user needs to wait up to one minute until the stale lock is detected and the user can restore the last backup.

libs/librepcbcommon/fileio/directorylock.cpp
rnestler: What about `std::exception`?

rnestler:
Proposal 3 Just use QLockFile which already implements stale lock handling...

libs/librepcbcommon/fileio/directorylock.cpp
ubruhin: > What about std::exception?

the same :wink:

ubruhin:
> According to the documentation QLockFile handles stale files automatically.

But it does not **report** if a lock was stale or not. And we need to know that to decide whether we have to restore the backup or not.

> Probably because unlock() can't fail: According to the documentation: "Calling unlock() without locking the file first, does nothing."

`unlock()` actually removes a file from the file system, which obviously **can** fail (no permissions, open file handles [windows], disconnected [network] drive, ...).

> Qt guaranties no breaking changes if you stay with the same major version. So this shouldn't be an issue.

I think this is related to the Qt library itself (API/ABI compatibility), but maybe not to files created by Qt (?).

ubruhin:
Detecting stale locks is now implemented too (with comparing username/hostname/pid/process name). I just copied the code from `QLockFile` ;)

Now let's see if the CI succeeds...

ubruhin:
That was a mess....but I think now (after >10 pushes) it should be ready to merge.

Btw, this PR also fixes #14.

PR #114, opened 2016-08-21
Workspace refactoring

ubruhin:
The purpose of this pull request is to implement the new workspace/library/repository concept which is discussed in #109. Just started, not yet ready to merge...

PR #115, opened 2016-08-21
Improve versioning system

ubruhin:
Some general improvements in the versioning system. Parts of this are required for #114.

This pull request changes the file version type from Integer to a semantic version string. So a workspace/library/project can now have version ```0.1``` or ```1.2.3``` instead of only ```0```, ```1```, and so on. This is required to allow making multiple file format upgrades **before** version ```1``` (which indicates a stable version) while keeping the file format version equal to the application version. The file format version needs to be a prefix of the application version, e.g. format version ```0.1``` is used by application version ```0.1.x``` and format version ```1``` is used by application version ```1.x```.

### DIFF
--- a/libs/librepcbcommon/application.cpp
+++ b/libs/librepcbcommon/application.cpp
@@ -33,9 +33,47 @@ namespace librepcb {
  *  Constructors / Destructor
  ****************************************************************************************/
 
-Application::Application(int& argc, char** argv) :
+Application::Application(int& argc, char** argv) noexcept :
     QApplication(argc, argv)
 {
+    // set application version
+    mAppVersion = Version(APP_VERSION);
+    QApplication::setApplicationVersion(mAppVersion.toPrettyStr(2));
+    Q_ASSERT(mAppVersion.isValid());
+
+    // set "git describe" version
+    mGitVersion = QString(GIT_VERSION);
+    if (mGitVersion.isEmpty()) {
+        qWarning() << "Could not determine Git version. Check if Git is added to the PATH.";
+    }
+
+    // set and verify file format version
+    mFileFormatVersion = Version(FILE_FORMAT_VERSION);
+    Q_ASSERT(mFileFormatVersion.isValid());
+    Q_ASSERT_X(mFileFormatVersion.isPrefixOf(mAppVersion),
+               "Application::Application()",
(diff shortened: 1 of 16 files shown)


libs/librepcbcommon/version.cpp
```diff
@@ -127,38 +151,56 @@ Version& Version::operator=(const Version& rhs) noexcept
...
 
 bool Version::operator!=(const Version& rhs) const noexcept
 {
-    if (mNumbers.isEmpty()) return false;
-    return (compare(rhs) != 0);
+    if (isValid() && rhs.isValid()) {
+        return (compare(rhs) != 0);
+    } else {
+        return false;
+    }
```
rnestler: This doesn't make any sense semantically. If any of the values is invalid (`isValid()` returns false) the `!=` will return true which means the objects are equal!

I think it should behave the same as a float with NaN.

ubruhin: Fixed :)

libs/librepcbcommon/version.h
```diff
@@ -172,7 +172,8 @@ class Version final
          *
          * @param rhs   The other object to compare
          *
-         * @return If at least one of both objects is invalid, false will be returned!
+         * @return  If at least one of both objects is invalid, false will be returned
+         *          (except #operator!=() which would return true in this case)!
          */
         Version& operator=(const Version& rhs) noexcept;
```
rnestler: This operator doesn't seem to belong here. *nagnagnag*

ubruhin: :fu:

libs/librepcbcommon/application.cpp
```diff
@@ -33,9 +33,47 @@ namespace librepcb {
  *  Constructors / Destructor
  ****************************************************************************************/
 
-Application::Application(int& argc, char** argv) :
+Application::Application(int& argc, char** argv) noexcept :
     QApplication(argc, argv)
 {
+    // set application version
+    mAppVersion = Version(APP_VERSION);
```
rnestler: Why don't you use the initializer list for those initializations?

ubruhin: Only because this way the initialization of members is grouped together with it's following checks/asserts...

PR #120, opened 2016-09-10
Add Uuid unit tests

ubruhin:
Because the class `Uuid` is very important for LibrePCB, here are some unit tests for it. And I have already fixed some small issues which were found by these tests ;)

### DIFF
--- a/libs/librepcbcommon/uuid.cpp
+++ b/libs/librepcbcommon/uuid.cpp
@@ -34,12 +34,14 @@ namespace librepcb {
 
 bool Uuid::setUuid(const QString& uuid) noexcept
 {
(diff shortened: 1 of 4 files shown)


rnestler:
Nice :+1: haven't taken a deep look at ot, but tests are always good! Maybe writing tests could be on the easy issues list?

ubruhin:
> Maybe writing tests could be on the easy issues list?

I think most classes aren't easy to test because mocks may be required and there is only little documentation about what the classes should exactly do. But especially those classes in `librepcbcommon` are not very complicated, so their unit tests should be (relatively) easy to write. I just opened #121 for that.

PR #125, opened 2016-10-01
Add workspace library manager

ubruhin:
This adds a workspace library manager to the control panel.

Features:
- Create/remove local libraries
- Manually download libraries from the Internet
- Download/update/remove remote libraries (list of libraries fetched from api.librepcb.org)
- Library dependencies are considered when downloading and updating remote libraries

Issues:
- Error handling may not be perfect
- Library dependencies are not considered when removing remote libraries
- Overall user experience is not that awesome...

Even if the library manager is far away from being perfect, I would say that's enough for the moment ;)

Screenshot:
workspace library manager_008


dbrgn:
Cool! :sparkles:

PR #128, opened 2016-10-30
Add bash script to build each commit of a branch

ubruhin:
This script allows to automatically build each single commit of the current branch (up to the master branch).

### DIFF
--- a/dev/build_each_commit.sh
+++ b/dev/build_each_commit.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+mkdir -p ../build && cd ../build
+
+set -euo pipefail
+IFS=$'\n\t'
+
+HEAD=$(git rev-parse --abbrev-ref HEAD)
+if [[ "${HEAD}" == "HEAD" ]]; then
+	HEAD=$(git rev-parse HEAD)
+fi
+echo "HEAD=${HEAD}"
+
+STASH=0
+if [[ -n $(git status --porcelain) ]]; then
+	git stash --include-untracked
+	STASH=1
+fi
+echo "STASH=${STASH}"
+
+START="master"
+END=${HEAD}
+
+trap "git checkout ${HEAD} && test ${STASH} -ne 1 || git stash pop" EXIT
+
+readonly COMMITS=$(git log --pretty=oneline "${START}...${END}" | cut -d' ' -f1 | tac)
+for commit in ${COMMITS[@]}; do
+	git checkout $commit
+	qmake -r ../librepcb.pro > "build_${commit}.log" 2>&1
+	make -j8 >> "build_${commit}.log" 2>&1
+done
+
+echo "SUCCESSFULLY FINISHED!"
+
--- a/dev/build_each_commit.sh
+++ b/dev/build_each_commit.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 mkdir -p ../build && cd ../build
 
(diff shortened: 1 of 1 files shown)


dev/build_each_commit.sh
```diff
@@ -0,0 +1,34 @@
...
+START="master"
+END=${HEAD}
+
+trap "git checkout ${HEAD} && test ${STASH} -ne 1 || git stash pop" EXIT
+
+readonly COMMITS=$(git log --pretty=oneline "${START}...${END}" | cut -d' ' -f1 | tac)
+for commit in ${COMMITS[@]}; do
+	git checkout $commit
+	qmake -r ../librepcb.pro > "build_${commit}.log" 2>&1
+	make -j8 >> "build_${commit}.log" 2>&1
```
dns2utf8: Switch to autodetect:
```
make -j$(nproc)
```

dev/build_each_commit.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
dns2utf8: Use `#!/usr/bin/env bash`

ubruhin:
> cool script, what is your usecase?

I often re-order commits in feature branches (i.e. `git rebase -i master`) to cleanup the history. But sometimes this introduces some errors, so that some commits won't compile anymore. With this script I can easily verify that each commit of the feature branch still compiles, so reordering didn't break something.

dev/build_each_commit.sh
ubruhin: Fixed.

dev/build_each_commit.sh
ubruhin: Fixed.

dns2utf8:
🌈

dev/build_each_commit.sh
```diff
@@ -0,0 +1,34 @@
...
+	STASH=1
+fi
+echo "STASH=${STASH}"
+
+START="master"
+END=${HEAD}
+
+trap "git checkout ${HEAD} && test ${STASH} -ne 1 || git stash pop" EXIT
+
+readonly COMMITS=$(git log --pretty=oneline "${START}...${END}" | cut -d' ' -f1 | tac)
```
dbrgn: Whaat? I've never sean `readonly` :) That's pretty cool.

ubruhin: I also didn't know that, I have stolen it from the Internet xD

ubruhin:
> What happens if your branch is not based on master? Or if you have a detached head?

The command `git log --pretty=oneline "master...HEAD"` may fail in that case, but that's no problem because the `trap` command will always restore all changes made by this script. It doesn't matter whether the script was successful or not.

Because it's only a developer tool (basically for myself), I didn't add too much features and error handling ;)

dbrgn:
Woah, I didn't know `trap` either. @ubruhin is the bash master :wink:

rnestler:
Minor nit-pick: Why add this to LibrePCB if it is a personal developer tool? Or in other words: I would like to use this on other projects and have it directly in `$HOME/bin` :smiley:

ubruhin:
> Why add this to LibrePCB if it is a personal developer tool? Or in other words: I would like to use this on other projects and have it directly in $HOME/bin

This script is explicitly designed to work with LibrePCB, the build steps will not work for other projects. So it makes more sense in this repository than in $HOME. And even if I created it for myself, this does not mean that nobody else can use it too ;)

Or should I change the whole repository into a private project because I'm the only user of it (or better delete it completely, because actually I also don't use it)? :smile:

Issue #47, opened 2016-01-22
Improve project library handling


0xB767B:
Looking at e.g. SES_AddComponents, there's loads of low level magic going on regarding access to the component libray. This must get into to some higher level interface!

ubruhin:
I think you mean especially the process of copying components and symbols to the project's library? This is a thing which is even in the low level modules not yet defined very well, so in the higher level modules (e.g. SES_AddComponents) it's a mess as well.

Before improving SES_AddComponents we have to define a clear strategy for copying library elements to the project's library:
* When adding components to a schematic, should we copy all required files from the workspace library to the project library immediately, or only when the user saves the project the next time?
* And on the other side, when removing components from a schematic, should these components be removed from the project's library immediately, or only when the user saves the project the next time?
* When adding/removing immediately: This would lead into changes inside the project's directory even if the user never saves the project --> unacceptable in my opinion!
* When adding/removing only when saving the project: Simple removing the library elements by deleting their directories is not good because this way we may loose some information of that library elements. That's not acceptable too because with the undo command (Ctrl+Z) the removed components must be re-added again (without loosing some information).

I think the only acceptable solution is to save added/removed components to temporary directories, which are synchronized with the project library when saving the project. But this would need lot of work to implement it...

0xB767B:
Many thoughts here... But I would suggest to split this discussion into two:
* High level interface used e.g. in SES_AddComponents.
* Handling of workspace/project libraries.

ubruhin:
I will fix this issue in the low level modules soon. After that, the SES_AddComponents can be refactored (which is necessary anyway).

ubruhin:
I have improved the project library handling in commit 155ba143a8f1383da08a8015ad37a3b54c6bb9f9, but it's still a little bit provisional.

Now there are some new high level undo commands which are important for the schematic/board editor FSM (see directory [libs/librepcbprojecteditor/cmd/](https://github.com/LibrePCB/LibrePCB/tree/master/libs/librepcbprojecteditor/cmd)).

PR #129, opened 2016-10-31
Add library editor

### LINKED ISSUE #3: Create a library editor
see static library "librepcblibraryeditor"

ubruhin:
Good news: This adds the long-awaited library editor! :smiley: 
Bad news: It's not yet finished... :disappointed: 

Preview:
librepcb base - librepcb library editor_008

Fixes #3


dbrgn:
Looking forward to it! :)

dbrgn:
I just tried it. Downloaded the three recommended libraries (the workflow is SO MUCH BETTER than KiCad!). But when I open them with the library editor they seem empty. Is that not yet implemented? :)

ubruhin:
> Is that not yet implemented? :)

This branch is just outdated because I'm working on [my fork](https://github.com/ubruhin/LibrePCB/tree/add_library_editor) to reduce spam in this PR (I do lot of rebasing / force pushing) ;)

ubruhin:
Good news: I think the library editor is now ready for some testing! :boom:

@dbrgn and @rnestler could you maybe play a little bit around with the library editor and then give some feedback? I know that there are still many small things which are kind of annoying, but at the moment it's most important to fix critical bugs, counterintuitive GUIs or other things which lead to frustrated users ;)

Here are some screenshots:

librepcb base - librepcb library editor_010

librepcb base - librepcb library editor_011

librepcb base - librepcb library editor_012

librepcb base - librepcb library editor_013

librepcb base - librepcb library editor_014

dbrgn:
Awesome! :sparkles: :tada: 

Let's test this next Monday, when you're around. Then you can look at how I try to use the editor, as a usability study :slightly_smiling_face:

rnestler:
My feedback from the testing today:
 * Add new pad:
    When adding new pad, enter should add the pad not just set the name
  package_pads
 * Footprints:
   It confused me that in the package editor there is a list of footprints. Maybe call it "Footprints variants"?
 * Footprint editor:
   * Show pad "name" / "number"
   * Show curser x/y position
   * Make default pad smaller
   * Maybe even add presets for pad sizes?
* New Library Element
  * Show better description on hover
    * example: Package: "Contains footprint variants and 3D model"
  * On the end of the wizard there is an error message if the (optional) author field was not set
package_author

dbrgn:
- [ ] Fix order of input fields when creating a library (tabbing jumps around input fields)
- [ ] Add a default picture for libraries without a picture
- [ ] Make it possible to add a picture when creating a library
- [ ] When creating a library, do an automatic rescan
- [ ] Disable buttons in toolbar that can't be used before editing a part
- [ ] Make rectangles resizeable
- [ ] Tab character in textarea fields should not insert a tab character but switch to next input field
- [ ] While drawing a footprint, the coordinates (absolute and relative) should be visible
- [ ] It would be great to have a real rectangle object in the footprint editor, so that it can be defined by width/height
- [ ] Example for package category is wrong (should probably be DIP, not DIP8)
- [ ] Hide ERC role in component wizard for now (mark all pins as passive)
- [ ] Show preview of symbol in component wizard
- [ ] Changing properties of a component (right hand side form) does not invoke the "changed" star on the tab
- [ ] Middle mouse click to close tabs
- [ ] When grabbing an object in the symbol editor and moving it to the border of the graphics view, start scrolling
- [ ] Author name is not pre-filled with system information
- [ ] When deleting the last instance of a device from a schematic, it should be removed from the project cache

ubruhin:
Wow, thank you very much for your amazing feedback, this was really helpful! :tada: :smiley: 

Now I need some time to fix these issues :stuck_out_tongue_winking_eye:

ubruhin:
Most of these issues are now fixed. But some of them I won't fix at the moment because they have low priority and/or need high effort:

> When adding new pad, enter should add the pad not just set the name

> Fix order of input fields when creating a library (tabbing jumps around input fields)

> Tab character in textarea fields should not insert a tab character but switch to next input field

> Middle mouse click to close tabs

> When grabbing an object in the symbol editor and moving it to the border of the graphics view, start scrolling

Actually all these things aren't specific to the library editor, they also apply to many other windows of LibrePCB. Because of this I have create a new "Project" to track these issues in a more generic manner: https://github.com/LibrePCB/LibrePCB/projects/4

> It could propose 1, 2, 3, ... for the pad names

Good idea, but unfortunately now it looks like the suggested next pad name is an already existing pad:
auswahl_010

As a compromise I leave the text field of the next pad empty (as until now), but when pressing the "+" button while the text field is still empty, the new pad gets a reasonable name (auto incremented number). So If you know that, you just can press the button for example 5 times to get 5 pads numbered from 1 to 5.

> Show pad "name" / "number"

Unfortunately that's not as easy as I hoped. I think It would make sense to slightly redesign the concept of graphics items to make this possible. See also #181.

> Make it possible to add a picture when creating a library

I think this is not really required, or at least not yet. Wizards should only ask for the most important metadata to keep them clean and simple. More data can always be added *after* creating the library (or some other element).

> Make rectangles resizeable

> It would be great to have a real rectangle object in the footprint editor, so that it can be defined by width/height

That's a lot of effort because it requires to introduce a new graphics object type. Most likely I will do that some day, but for the moment polylines must do the trick ;)

> When deleting the last instance of a device from a schematic, it should be removed from the project cache

That's part of #47.

ubruhin:
I have rebased this branch and squashed many commits together to keep the history clean.

Now this should be ready to merge, or does someone want to test the library editor (again)?

PR #127, opened 2016-10-30
Directory and namespace structure refactoring

ubruhin:
There are some issues with the actual directory/namespace structure in this repository:
- Namespaces do not always match the directory structure (for example the static libraries `librepcbproject` and `librepcbprojecteditor` are located in different directories, but both use exactly the same namespace `librepcb::project`).
- The name of the static libraries are quite long (for example `librepcbprojecteditor`), see #52.
- The library manager (which is currently located in `librepcbworkspace`) will depend on `librepcblibraryeditor` in future, but this would break our [architecture design](https://raw.githubusercontent.com/LibrePCB/LibrePCB/8cf2876247d85c8a9d2411a5f8a979edb32d0995/dev/doxygen/images/architecture_overview.png). So we should move the library manager into a separate static library (e.g. `librepcblibrarymanager`).
- Some static libraries are located in the directory `3rdparty`, some in `libs`. We should move all libraries into `libs` to keep the directory structure simple.
- The main application is located in the directory `librepcb` and all other applications in `tools`. To keep things simple, we should move all applications into a directory named `apps`.

This would also fix #52.

## Current directory structure:

```
LibrePCB/
├── 3rdparty/
│   ├── gmock/
│   ├── hoedown/
│   └── quazip/
├── librepcb/
├── libs/
│   ├── librepcbcommon/
│   ├── librepcblibrary/
│   ├── librepcblibraryeditor/
│   ├── librepcbproject/
│   ├── librepcbprojecteditor/
│   └── librepcbworkspace/
└── tools/
    ├── EagleImport/
    ├── ProjectLibraryUpdater/
    ├── UuidGenerator/
    └── WorkspaceLibraryUpdater/
```

## My proposal for a new directory structure:

```
LibrePCB/
├── apps/
│   ├── EagleImport/
│   ├── librepcb/
│   ├── ProjectLibraryUpdater/
│   ├── UuidGenerator/
│   └── WorkspaceLibraryUpdater/
└── libs/
    ├── gmock/
    ├── hoedown/
    ├── quazip/
    └── librepcb/
        ├── common/
        ├── library/
        ├── libraryeditor/
        ├── librarymanager/
        ├── project/
        ├── projecteditor/
        └── workspace/
```

## My proposal for namespace changes:
- LibrePCB Common: `librepcb` → `librepcb`
- LibrePCB Library: `librepcb::library` → `librepcb::library`
- LibrePCB Library Editor: `librepcb::library` → `librepcb::library::editor`
[...]

PR #130, opened 2016-10-31
Replace GoogleMock directory with a submodule

ubruhin:
Because [GoogleTest and GoogleMock have been moved to GitHub last year](https://groups.google.com/forum/#!topic/googletestframework/UD3nRr3RAQk), and they [released version 1.8.0 two months ago](https://github.com/google/googletest/releases/tag/release-1.8.0), it's a good time to update and replace our [copy of gtest/gmock](https://github.com/LibrePCB/LibrePCB/tree/master/3rdparty/gmock) with a submodule.

I have just forked the [google/googletest](https://github.com/google/googletest) repository into [LibrePCB/googletest](https://github.com/LibrePCB/googletest), where I have created a new branch for our own modifications (basically to add a qmake project file).

Because this would make #127 easier to fix, I think we should merge this PR first.


dns2utf8:
Maybe add a hint to the readme for interested devs to rerun `git submodule init && git submodule update`?

ubruhin:
> Maybe add a hint to the readme for interested devs to rerun git submodule init && git submodule update?

A permanent hint in the readme which relates only to a specific commit? I wouldn't do that...

dns2utf8:
Yey, true.
Can you automate it in a configure step?

ubruhin:
In my opinion a configure step is not (yet) required. Then developers would need to know that one have to run `./configure` instead of `git submodule init && git submodule update`, so we have not really won something...

At the moment, I would say that developers just need to know that we use submodules. And because all developers are watching this repository, they all got an email with the title of this PR and thus are automatically informed about the new submodule ;)

dns2utf8:
👍 🌈

PR #133, opened 2016-12-04
Sort entries of HEADERS, SOURCES and FORMS in *.pro files

ubruhin:
The new script `sort_qmake_file_entries.py` sorts the entries of HEADERS, SOURCES and FORMS variables in all qmake project files (*.pro) in alphabetical order (except those in submodules). This reduces the risk of conflicts when merging or rebasing commits which modify *.pro files and makes those files cleaner and thus easier to read.

It may even be used as a pre-commit hook.

See also https://bugreports.qt.io/browse/QTCREATORBUG-553.

### DIFF
--- a/dev/sort_qmake_file_entries.py
+++ b/dev/sort_qmake_file_entries.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from __future__ import print_function
+import os 
+import glob
+from shutil import copyfile
+
+
+"""
+This script sorts the entries of HEADERS, SOURCES and FORMS variables in
+all qmake project files (*.pro) in alphabetical order (except those in 
+submodules). This reduces the risk of conflicts when merging or rebasing
+commits which modify *.pro files and makes those files cleaner and thus 
+easier to read.
+
+This script may be used as a pre-commit hook:
+ln -rsT sort_qmake_file_entries.py ../.git/hooks/pre-commit
+"""
+
+
+def get_repository_root_dir(start_dir):
+    if os.path.isdir(os.path.join(start_dir, '.git')):
+        return start_dir
+    else:
+        parent_dir = os.path.dirname(start_dir)
+        if parent_dir != start_dir:
+            return get_repository_root_dir(parent_dir)
+        else:
+            return None
+
+
+def sort_qmake_file(project_root, filepath):
+    with open(filepath, 'r') as file:
+        old_lines = list(file.readlines())
+        new_lines = list(old_lines)
+    block_start_index = None
+    for line_index, line in enumerate(old_lines):
+        if block_start_index is None:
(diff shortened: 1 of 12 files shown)


dev/sort_qmake_file_entries.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import os 
+import glob
+from shutil import copyfile
```
rnestler: Add `from __future__ import print_function` and use `print()` in the rest of the code to be python3 compatible.

dev/sort_qmake_file_entries.py
```diff
@@ -0,0 +1,95 @@
...
+
+# search all *.pro files and sort their HEADERS, SOURCES and FORMS entries
+print "Sort *.pro file entries..."
+modified_files = sort_qmake_files_in_dir(project_root_dir, project_root_dir)
+print "Finished: {} Files modified.".format(modified_files)
+
+# exit with code != 0 if files were modified
+# -> aborts the commit when used as pre-commit hook!
+exit(modified_files)
+
```
rnestler: It would be nicer to put everything in a `main()` function.

dev/sort_qmake_file_entries.py
```diff
@@ -0,0 +1,100 @@
...
+"""
+
+
+def get_repository_root_dir(start_dir):
+    if os.path.isdir(os.path.join(start_dir, '.git')):
+        return start_dir
+    else:
+        parent_dir = os.path.dirname(start_dir)
+        if parent_dir != start_dir:
+            return get_repository_root_dir(parent_dir)
```
dbrgn: Does this stop at `/`, or will you get infinite recursion?

rnestler: `if parent_dir != start_dir:` will abort the recursion at `/` I guess?

dbrgn: Ah, right.

dev/sort_qmake_file_entries.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from __future__ import print_function
```
ubruhin: The script do not use any kind of division, so I think the `division` module should not change anything. `unicode_literals` may be nice, but I would say now it's too late to change :wink:

dev/sort_qmake_file_entries.py
ubruhin: Yes, as @rnestler noticed it should abort at `/`.

ubruhin:
> A dry-run mode might also be useful for testing :)

Good idea, but as it is already merged (and especially because it's not a critical feature for end users, but only for developers) I won't add this now. In addition, the script already creates backups before modifying files, so things can be reverted if something goes wrong ;)

PR #134, opened 2016-12-12
GraphicsView: Pan with middle mouse button instead of right button

ubruhin:
Panning with right mouse button is quite annoying as it conflicts with context menus :sob: Now the middle mouse button is used instead. Was really hard to implement... :smile:

See also #4.

### DIFF
--- a/libs/librepcb/common/graphics/graphicsview.cpp
+++ b/libs/librepcb/common/graphics/graphicsview.cpp
@@ -201,8 +201,8 @@ bool GraphicsView::eventFilter(QObject* obj, QEvent* event)
         {
             if (!underMouse()) break;
             QGraphicsSceneMouseEvent* e = dynamic_cast<QGraphicsSceneMouseEvent*>(event); Q_ASSERT(e);
-            if (e->buttons().testFlag(Qt::RightButton) && (!mPanningActive)) {
-                QPoint diff = mapFromScene(e->scenePos()) - mapFromScene(e->buttonDownScenePos(Qt::RightButton));
+            if (e->buttons().testFlag(Qt::MiddleButton) && (!mPanningActive)) {
+                QPoint diff = mapFromScene(e->scenePos()) - mapFromScene(e->buttonDownScenePos(Qt::MiddleButton));
                 mPanningActive = true; // avoid recursive calls (=> stack overflow)
                 horizontalScrollBar()->setValue(horizontalScrollBar()->value() - diff.x());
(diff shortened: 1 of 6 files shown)


ubruhin:
Im'm not yet sure if the middle mouse button is better than the right button, because it's less convenient to click. Maybe it would be better to stay with the right button but allow some tolerance before starting to pan. So if the user presses the button, moves the cursor 5px and releases the button, the context menu is shown instead of panning. Or maybe a timeout is better, so if the click was shorter than 200ms for example, the context menu is shown, no matter how far the cursor was moved.

dbrgn:
I like the middle mouse button. It's also what KiCad uses btw, so many users might be accustomed to it.

For laptop users you could also implement some kind of ctrl+LMB. Not sure about that though.

rnestler:
> For laptop users you could also implement some kind of ctrl+LMB. Not sure about that though.

Laptop users typically use their touch-pad to scroll / pan, don't they?

dbrgn:
> Laptop users typically use their touch-pad to scroll / pan, don't they?

You mean horizontal/vertical two-finger scrolling?

rnestler:
> You mean horizontal/vertical two-finger scrolling?

Yes.

ubruhin:
Today I saw that Altium also uses the right mouse button for panning and some of my colleagues also don't like the middle mouse button for panning. So in my opinion we should better stay with the right mouse button.

But then I learned an [interesting detail](https://en.wikipedia.org/wiki/Context_menu#Implementation) of context menus:
> Windows mouse click behavior is such that the context menu doesn't open while the mouse button is pressed, but only opens the menu when the button is released, so the user has to click again (this time with the first mouse button) to select a context menu item. This behavior differs from that of macOS and most free software GUIs.

A short test of myself confirms that context menus under Windows are shown when **releasing** the right mouse button, and under Linux the menu is shown when **pressing** the button. I also saw that Qt emits the event [`QContextMenuEvent`](http://doc.qt.io/qt-5/qcontextmenuevent.html) according to this rule, so the behavior is platform dependent.

This brings the question up: How can we pan with the right mouse button on Linux/OSX if the context menu is *immediately* shown when pressing the button? :wink:

Btw, KiCad seems to break this rule by opening the context menu always immediately when pressing the button, even under Windows. This may be the reason why panning with the right mouse button is not possible at all with KiCad...

ubruhin:
Because there are some unresolved issues with the right mouse button for panning, I merge this PR. So from now on the middle mouse button is used for panning.

Maybe some time we add a configuration option for using the right mouse button for panning again.

PR #136, opened 2017-01-03
Travis-CI: Cleanup configuration

ubruhin:
- Move logic into separate shell scripts to make the whole configuration more clear.
- Remove secure variable from `.travis.yml`, store it in the web interface instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,13 @@
 language: cpp
-
 sudo: required
-
 cache: ccache
+notifications:
+  email:
+    on_success: change
+    on_failure: change
 
 matrix:
+  fast_finish: true
   include:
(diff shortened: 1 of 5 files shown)


rnestler:
:+1:

ubruhin:
> It's just moving most parts into shell scripts, is it?

Yes, hopefully this makes the CI configuration more clear.

> What do you mean by that?

At the moment, an access token from GitHub is defined in `.travis.yml` as an encrypted variable `env.global.secure`. But this is kind of "spam" (and can't be decrypted in builds from forks and pull requests anyway), so I decided to remove it from `.travis.yml` and store it in the Travis-CI webinterface instead. It seems that the behavior is exactly the same as before, but the `.travis.yml` is kept more clear. See [Travis docs](https://docs.travis-ci.com/user/environment-variables#Defining-Variables-in-Repository-Settings) for more information.

Issue #92, opened 2016-07-07
Use std::unique_ptr<T> instead of raw pointer where applicable


rnestler:
There are a lot of places in the code where raw pointers which could be replaced straight forward by `std::unique_ptr`.

ubruhin:
Until now, Qt's smart pointers are used instead of the C++11 ones. The only disadvantage seems to be that QScopedPointer has no move semantics. As a workaround we could use QSharedPointer when passing pointers to outside of function scope.

The other way would be to replace all Qt smart pointers with C++11 ones and don't use Qt pointers any more. 

C++11 smart pointers never throw std::exceptions by themself, do they? If they do, we would have to wrap them with our own exception classes...

ubruhin:
I consider this as closed for following reasons:
- C++ smart pointers can throw exceptions in many situations. So we would have to wrap all `std::exception` with our own `librepcb::Exception` class. This is PITA.
- Qt's smart pointers are already widely used in LibrePCB. And in every new class I'm creating or refactoring, I use smart pointers instead of raw pointers where applicable.
- Replacing all old raw pointers with smart pointers at once would require lot of effort. I think it's better to replace them step by step when refactoring old classes. Most of the classes will need some refactoring anyway at some point...
- [Using smart pointers is now recommended in the developers documentation](https://github.com/LibrePCB/LibrePCB/commit/97cdea86516fcb4278a511f16737718dbe90ac74#diff-ffaf3dcdbd556e0df700be213ecb89d8R22).
- Maybe Qt provides a unique pointer with move semantics someday ;) Until then we can use `QSharedPointer` instead.

PR #138, opened 2017-01-16
Replace manual memory management with std::unique_ptr

rnestler:
To be able to use features like std::make_unique() we need C++14.

### DIFF
--- a/common.pri
+++ b/common.pri
@@ -33,8 +33,8 @@ lessThan(QT_MAJOR_VERSION, 5) {
     }
 }
 
-# c++14 is obligatory!
-CONFIG += c++14
+# c++11 is obligatory!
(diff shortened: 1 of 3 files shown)


ubruhin:
C++14 seems to require GCC 5.0 or later, as described [here](https://gcc.gnu.org/projects/cxx-status.html#cxx14). Unfortunately Ubuntu 14.04 ships with GCC 4.8 (I guess), so for Ubuntu 14.04 users it won't be possible to contribute to LibrePCB unless manually updating GCC.

Because of this, in my opinion it's not (yet) worth it to switch to C++14.

ubruhin:
I'm sorry but I can't merge this because of the same reason as in #92 ;)

PR #140, opened 2017-01-19
Build and deploy nightly AppImage

ubruhin:
Having [AppImage](http://appimage.org/) for nightly builds but also for stable releases would be pretty nice. With this PR I try to let them build on Travis-CI and upload to [Bintray](https://bintray.com/librepcb).

Basically it already works, but there are still some issues:
- [x] Runtime resources do not work inside the AppImage (see implementation in 5d80d1b96e89259b12092d35c54189fd19ac212b).
- [ ] ~~The icon of the *.AppImage file seems to be square instead of round (at least in Nemo). Maybe the background of the *.png is not transparent? Could we even use SVG instead?~~ Unfortunately the icon is no longer shown anyway (see https://github.com/probonopd/AppImageKit/issues/346).
- [x] At least on some systems the look of the LibrePCB application from an AppImage is not the same as without the AppImage. It seems that the system theme is not used properly.
- [ ] The file [bintray.json](https://github.com/LibrePCB/LibrePCB/blob/f57e907c1b6a2a8420f3e9ed2e67263ccaf044d4/dev/travis/bintray.json) contains only static information. How could we also use some dynamically determined information (e.g. datetime, version number, commit hash)?
- [ ] ~~At the moment Travis-CI downloads [`linuxdeployqt`](https://github.com/probonopd/linuxdeployqt) with every build directly from GitHub. I don't like to download and execute binaries this way because of possible security issues (a compromised binary could read secure environment variables). Is there another ("more official") way to get `linuxdeployqt`? Maybe building from source would be the best...~~ Actually the AppImages of `linuxdeployqt` are easy to use and work pretty good, and now I also know its source code and have more trust into `linuxdeployqt`, so I don't see this as a problem any more.
- [ ] Maybe some automatic testing of the generated packages would be useful.
- [x] As soon as everything is working properly, we should stop uploading packages to Bintray from branches other than `master`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,12 @@ matrix:
     - os: linux
       dist: trusty
       compiler: gcc
-      env: BUILD_DOXYGEN=true QT_BASE="trusty"
-    # Ubuntu 14.04 + Clang + Latest Qt from beineri PPA (https://launchpad.net/~beineri)
+      env: BUILD_DOXYGEN=true DEPLOY_APPIMAGE=false QT_BASE="trusty"
+    # Ubuntu 14.04 + Clang + Qt from PPA (https://launchpad.net/~beineri) + AppImage deployment
     - os: linux
       dist: trusty
       compiler: clang
-      env: BUILD_DOXYGEN=false QT_BASE="qt58" QT_PPA="ppa:beineri/opt-qt58-trusty"
+      env: BUILD_DOXYGEN=false DEPLOY_APPIMAGE=true QT_BASE="qt58" QT_PPA="ppa:beineri/opt-qt58-trusty"
     # OS X + GCC
     - os: osx
       compiler: gcc
@@ -35,4 +35,5 @@ script:
   - ./dev/travis/build.sh
   - ./dev/travis/run_tests.sh
   - ./dev/travis/update_doxygen.sh
+  - ./dev/travis/build_appimage.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,3 +37,13 @@ script:
   - ./dev/travis/update_doxygen.sh
   - ./dev/travis/build_appimage.sh
 
+deploy:
+  provider: bintray
+  file: ./dev/travis/bintray.json
+  user: "${BINTRAY_USER}"
+  key: "${BINTRAY_API_KEY}"
+  skip_cleanup: true
+  on:
+    branch: master
+    condition: ${DEPLOY_APPIMAGE} = true
+
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ LibrePCB by yourself. Nevertheless there are easier ways to get LibrePCB running
 
 - For Windows there are nightly builds available to download: 
   [librepcb-nightly.zip](https://ci.appveyor.com/api/projects/librepcb/librepcb/artifacts/build/librepcb-nightly.zip?branch=master)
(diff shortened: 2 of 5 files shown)


dbrgn:
At first, I was really sceptical of this concept of standalone binaries. But after watching https://www.youtube.com/watch?v=mVVP77jC8Fc halfway through, I think it's a great fit for nightlies! And our final goal should be packaging by distributions themselves anyways :)

probonopd:
> At the moment Travis-CI downloads linuxdeployqt with every build directly from GitHub.

What do you propose? You could build it yourself from a "known good" commit...

ubruhin:
> Great work @ubruhin!

Thanks! But actually I have to thank you for trying to build AppImages for LibrePCB in your fork of LibrePCB, because otherwise I might never have started to build AppImages :smile: 

> Check what I do in
https://github.com/probonopd/AppImages/blob/master/bintray.sh

Wow, looks like lot of work! I will check it out as soon as I continue working on this feature.

> What do you propose? You could build it yourself from a "known good" commit...

Of course it would be great if there were a package in the official Debian/Ubuntu repositories ;) But as this is not the case, I think I will just continue using your binaries...

> Maybe upload the other builds to transfer.sh instead, for testing. transfer.sh keeps them for 14 days and deletes them automatically thereafter.

Good idea, maybe I will try that out!

ubruhin:
There is still lot of work until our AppImage reaches the quality required for official releases. But I think just for nightly builds, the current AppImage is ready to use :smiley: 

It would be great if someone could quickly test the AppImage: [LibrePCB-Nightly-Linux-x86_64.AppImage](https://bintray.com/librepcb/LibrePCB-Nightly/download_file?file_path=LibrePCB-Nightly-Linux-x86_64.AppImage)

I think I would then merge this PR regardless of the [high number of open issues](https://github.com/LibrePCB/LibrePCB/pull/140#issue-202003080) :wink:

probonopd:
Works for me @ubruhin

dbrgn:
Seems to work for me.

Something I noticed (unrelated to AppImage): If the current workspace is not readable or has the wrong library version number, the application shows an error message and quits instead of offering to choose/create a different workspace.

ubruhin:
Thanks for testing @probonopd and @dbrgn!

> In my experience, Qt 5.8 is much improved in this regard.

Awesome! Just tested with Qt 5.8 and in fact it looks much better :smiley: 

> What would you like to have? You could mirror your own copy.

Actually now I'm quite happy with your AppImage releases :wink: 

> the application shows an error message and quits instead of offering to choose/create a different workspace.

That's true, I also noticed that some time ago --> #154

ubruhin:
The AppImage now contains Qt 5.8 which improved the look of LibrePCB.

probonopd:
You could consider something like [uploadtool](https://github.com/probonopd/uploadtool) to upload AppImages to your Releases page, therefore increasing their visibility.

ubruhin:
> You could consider something like uploadtool to upload AppImages to your Releases page, therefore increasing their visibility.

To be honestly, I don't like having nightly builds on GitHub releases because every time a release is updated (i.e. on every push to master) all watchers of the project gets notified. This is too much spam for me, and I couldn't find a way to get rid of that. I noticed that problem with your project `linuxdeployqt` :wink:

probonopd:
>  every time a release is updated (i.e. on every push to master) all watchers of the project gets notified

Indeed, interesting aspect, didn't think of this. Thanks.

PR #141, opened 2017-01-21
Scan workspace library on every program startup

ubruhin:
Until now, the SQLite database of the workspace library is [put under version control](https://github.com/LibrePCB/demo-workspace/blob/36974bc36e875a28fb5affa3b291b699d5123dc1/v0.1/metadata/library_cache.sqlite) because it was not automatically generated.

This PR will force to (re)scan the workspace library on every program startup to ensure that it is always up to date. Because this is done completely asynchronously in a separate thread and can be interrupted at any time (e.g. on program exit), it should not have any negative impact on user experience (except maybe little decreased performance due to higher CPU load for a few seconds).

Thanks to this change I could [remove the SQLite database from version control](https://github.com/LibrePCB/demo-workspace/commit/eee0d43904412292418758c12e2c4867fedc3b82).

### DIFF
--- a/apps/librepcb/controlpanel/controlpanel.cpp
+++ b/apps/librepcb/controlpanel/controlpanel.cpp
@@ -120,6 +120,9 @@ ControlPanel::ControlPanel(Workspace& workspace) :
         if ((filepath.isExistingFile()) && (filepath.getSuffix() == "lpp"))
             openProject(filepath);
     }
+
+    // start scanning the workspace library (asynchronously)
+    mWorkspace.getLibraryDb().startLibraryRescan();
 }
 
 ControlPanel::~ControlPanel()
--- a/libs/librepcb/common/sqlitedatabase.cpp
+++ b/libs/librepcb/common/sqlitedatabase.cpp
@@ -81,11 +81,13 @@ SQLiteDatabase::SQLiteDatabase(const FilePath& filepath) throw (Exception) :
 
     // set SQLite options
     exec("PRAGMA foreign_keys = ON"); // can throw
+    enableSqliteWriteAheadLogging(); // can throw
 
     // check if all required features are available
     Q_ASSERT(mDb.driver() && mDb.driver()->hasFeature(QSqlDriver::Transactions));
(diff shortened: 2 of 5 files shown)


dbrgn:
The database is used like an index/cache, right? What happens if it already exists, is it re-created?

ubruhin:
> The database is used like an index/cache, right? What happens if it already exists, is it re-created?

Yes and Yes :)

The best would be to rescan the library only if the library files/directories were modified since the last scan. But this requires us to store the hash of every library file/directory into the SQLite database (or something like that). Then we could skip files which were not modified since the last scan to speed up the scanning process.

But for the moment I think it should also work without this optimization...

dbrgn:
Something I forgot, could there be a race condition if generating the database takes long, but another part of the code wants to access it before it's done?

ubruhin:
> Something I forgot, could there be a race condition if generating the database takes long, but another part of the code wants to access it before it's done?

Because the same database file is opened multiple times (one time for normal use, and one time for the rescan), the SQLite database driver is responsible for synchronization. As described [here](https://sqlite.org/threadsafe.html) (mode 2), this should be save. And because I use transactions, the whole rescan should be one single atomic operation on the database.

dbrgn:
> Actually the thread safety of SQLite depends on conditions. Maybe I should check the SQLITE_THREADSAFE option to make sure that multithreading is really supported ;)

Yes, that's probably a good idea.

ubruhin:
The SQLite driver of Qt seems to be compiled with `SQLITE_THREADSAFE=1`, so it should be safe to access the database from different threads. I have now also added an assert which checks the value of this compile time option.

But there was effectively another multithreading issue: As long as the library was scanned, all other accesses to the library database were completely blocked! :open_mouth: Now with d90b82d8752ebae90b69d2f7bb028dfb2c175d18 I have enabled [Write-Ahead Logging](http://www.sqlite.org/wal.html) to avoid this behavior.

dbrgn:
Cool, the workspace scan acts as a canary :wink:

ubruhin:
Now I also have added some unit tests for the `SQLiteDatabase` class, which also tests concurrent reading/writing from/to the database with multiple threads. I think if the CI passes, this PR is ready to merge...

PR #143, opened 2017-02-02
README: General refactoring, extending and cleanup

ubruhin:
Refactoring time! :smile: 

[Click here for a preview.](https://github.com/LibrePCB/LibrePCB/blob/refactor_readme/README.md)


README.md
```diff
@@ -17,17 +17,22 @@ development stage (no stable release available).
...
 - All-In-One: project management + schematic editor + board editor
-- Intuitive, modern and easy-to-use schematic/board editors
-- Very powerful library design
-- Uses XML as primary file format (allows version control of libraries and projects)
+- Intuitive, modern and easy-to-use graphical user interface
+- Very powerful library design with some innovative concepts
+- Uses human-readable file formats for both libraries and projects
 - Multi-PCB feature (different PCB variants of the same schematic)
-- Automatic netlist synchronisation between schematic and PCB (no manual export/import)
-- Up to 100 copper layers :-)
```
dbrgn: I think the missing artificial layer limit is obvious in an open source tool, right?

README.md
```diff
@@ -17,17 +17,22 @@ development stage (no stable release available).
...
+- Automatic netlist synchronisation between schematic and board
 
-### Planned Features
-- Integrated 3D PCB viewer + 3D file export
-- Automatic SPICE netlist export
-- Reverse engineering mode (from PCB to schematic)
```
dbrgn: Maybe re-add that section as "ideas for the future"? It would show that these things are not non-goals.

README.md
```diff
@@ -17,17 +17,22 @@ development stage (no stable release available).
...
+
+## Installation
+
+There are no stable releases available yet, so basically you need to compile
+LibrePCB by yourself. Nevertheless there are easier ways to get LibrePCB running:
+- On Arch Linux you can just install the package
```
dbrgn: I think depending on the markdown parser you might need an empty line between a paragaraph and a list. (Same thing for headlines and paragraphs.)

README.md
```diff
@@ -36,52 +41,46 @@ development stage (no stable release available).
...
 
 ```bash
-sudo apt-get install g++ qt5-default qttools5-dev-tools qt5-doc qtcreator libglu1-mesa-dev zlib1g zlib1g-dev openssl
+sudo apt-get install g++ qt5-default qttools5-dev-tools qt5-doc qtcreator \
+libglu1-mesa-dev zlib1g zlib1g-dev openssl
 ```
 
-#### Installation on ArchLinux
-
-You can install [librepcb-git](https://aur.archlinux.org/packages/librepcb-git/) from the AUR.
```
dbrgn: If you just want a list of the build dependencies on Arch (to keep it in this list):

```
sudo pacman -S qt5-webkit qt5-tools desktop-file-utils shared-mime-info git qconf
```

README.md
```diff
@@ -36,52 +41,46 @@ development stage (no stable release available).
...
+It's important to clone the repository recursively to get all submodules too:
 
-        open ./build/generated/mac/librepcb.app
-
-#### Installation in a docker container
-
-To build and run LibrePCB in a [docker](https://www.docker.com/) container (which is pretty cool!), check out [these instructions](https://github.com/LibrePCB/LibrePCB/tree/master/dev/docker).
+```bash
+git clone --recursive https://github.com/LibrePCB/LibrePCB.git && cd LibrePCB
+```
```
dbrgn: Maybe add another section:

> ...or if you have already cloned the repository:
> 
> cd LibrePCB
> git submodule init
> git submodule update

README.md
ubruhin: In my opinion, yes. Of course we should document the layer limit somewhere, but I think the readme is the wrong place for such a obvious information.

README.md
ubruhin: I've moved them to the [wishlist](https://github.com/LibrePCB/LibrePCB/wiki/Wishlist) because this is our new place for documenting long-term goals.

README.md
ubruhin: That seems to be true because ReText actually renders it wrong ;) I will fix that...

README.md
ubruhin: Thanks, I will add it!

README.md
ubruhin: I'd like to keep the readme short, and because this information is redundant (same can be done with removing and re-cloning the repository) it feels like spam... ;)

README.md
ubruhin: I think this list is quite outdated, isn't it?

- Webkit is no longer a dependency of LibrePCB
- g++, zlib and openssl are missing (or are they all installed per default on Arch?)
- Which package includes the Qt5 library itself?

Actually it would be great if a CI was trying to install `librepcb-git` from the AUR to see whether it is still working properly.

dbrgn: g++ is definitely a prerequisite on Arch Linux. Not sure about zlib and openssl, I probably need to add them to the list. Didn't know about the webkit thing.

I'll update the AUR package :)

dbrgn: Oh, I already removed qt5-webkit on July 18th last year - I simply pasted the outdated dependency list to Github :)

Regarding g++:

> Note: The group base-devel is assumed to be already installed when building with makepkg. Members of this group should not be included in makedepends array.

I assume zlib and openssl are both real dependencies, not just build dependencies, right? Added them both to `depends=`.

ubruhin: zlib is required for both compile-time and runtime, so headers and shared libraries are needed. OpenSSL is only required as shared library at runtime.

Btw, `qt5-tools` is only required for building, it's no runtime dependency. But `qt-base` is both, a runtime and compile-time dependency, see [bitcoin-qt](https://www.archlinux.org/packages/community/x86_64/bitcoin-qt/) for example.

And why do we need `qconf`?

dbrgn: Ah that's a good point, why is `qt5-base` missing? I thought that was in the list previously.

I'm not sure regarding qconf.

I think I should test these dependencies in a docker container that only has the standard system installed :)

ubruhin: I just tried it out on an Arch Linux in a docker container, and it turned out that not even `gcc` and `make` were installed. So I also added them to the list. This worked for me:
```bash
sudo pacman -S git make gcc qt5-base qt5-tools desktop-file-utils shared-mime-info openssl zlib
```

dbrgn: In that case, just add `base-devel` instead.

To quote the packaging guidelines:

> Note: The group base-devel is assumed to be already installed when building with makepkg. Members of this group should not be included in makedepends array.

rnestler: Definitely just add `base-devel`. Every Arch user which wants to compile something just installs it.

ubruhin: OK, I replaced `gcc` and `make` by `base-devel`.

> To quote the packaging guidelines:
> > Note: The group base-devel is assumed to be already installed when building with makepkg.

That may be true, but it's not relevant for the dependency list in our readme, because when building LibrePCB "by hand" (not by the `librepcb-git` package from AUR) one needs to install `base-devel` by himself.

So actually the dependency list of the AUR package is different by the one we provide in our readme because these are two different use cases.

dbrgn: Yes, that is true.

PR #144, opened 2017-02-05
AppVeyor: Provide nightly builds for Windows

ubruhin:
Surprisingly, uploading build artifacts is quite easy with AppVeyor ;) So I started to use this feature to provide nightly builds for Windows. They can now be downloaded [here](https://ci.appveyor.com/api/projects/librepcb/librepcb/artifacts/build/librepcb-nightly.zip?branch=appveyor_nightly_builds). 

Just extract the downloaded ZIP archive and execute `bin\librepcb.exe`. Isn't this awesome?! :smiley: 

@nemofisch Maybe you would like to try if it works properly on your machine?

The only thing which is not yet working is to load runtime resources of LibrePCB (e.g. license template files). I think the resource system needs to be adjusted slightly...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: 0.1.{build}-{branch}
+version: '{build}'
 os: Windows Server 2012
 configuration: Debug
 build: off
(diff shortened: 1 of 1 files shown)


dbrgn:
2017-02-05-232516_410x284_scrot

Doesn't work with wine :cry:

ubruhin:
> Doesn't work with wine

:sob: Would be interesting to analyze why it doesn't run...

ubruhin:
Update:
- Runtime resources are now loaded correctly (thanks to #146).
- I have just successfully tested the nightly build on a fresh Windows 10 installation.

=> I'll merge this PR.

PR #146, opened 2017-02-06
Refactor resources system

### LINKED ISSUE #116: Do not embed translation files into the executable
Until now, all compiled translation files (`*.qm`) are embedded into the executable by using the [Qt Resource System](http://doc.qt.io/qt-5/resources.html). This is not very meaningful because only a small subset of all translation files are really used at runtime.

I think the new resource file system which was created in 5d80d1b96e89259b12092d35c54189fd19ac212b would be better suited for storing translation files, because only the required files will be loaded at runtime. Unused files do not consume memory or processor time.

ubruhin:
- The filepath to runtime resource files is no longer configurable by preprocessor defines because this was too error-prone and made it difficult to create packages for distributing the application (see issues in #140 and #144). Runtime resources must now be located at `../share/librepcb` relative to the executable, which should be much more portable as there are no longer any absolute paths used.
- The directories `./res/` and `./packaging/` were replaced by the new directory `./share/`, which contains all resource files in a directory structure which can be simply copied to `/usr/share/` or `/usr/local/share/` to install them on a Unix/Linux system.
- Qt translation files (*.qm) are no longer statically compiled into the librepcb executable, but just copied to the resources directory `./share/` instead. This simplifies and speeds up deploying of translation files (if `make` detects no changes, building takes now ~300ms instead of ~2s on my machine!). Fixes #116.


libs/librepcb/workspace/settings/items/wsi_applocale.cpp
```diff
@@ -45,27 +46,29 @@ WSI_AppLocale::WSI_AppLocale(const QString& xmlTagName, XmlDomElement* xmlElemen
         mAppLocaleTmp = mAppLocale;
     }
 
+    QString i18nDir = qApp->getResourcesFilePath("i18n").toStr();
```
rnestler: This could be const.

apps/librepcb/main.cpp
```diff
@@ -157,18 +157,19 @@ static void installTranslations() noexcept
     qApp->installTranslator(qtTranslator);
 
     // Install system language translations (all system languages defined in the system settings, in the defined order)
+    QString dir = qApp->getResourcesFilePath("i18n").toStr();
```
rnestler: dir could be const, couldn't it?

libs/librepcb/workspace/settings/items/wsi_applocale.cpp
ubruhin: arghh... fixed :wink:

apps/librepcb/main.cpp
ubruhin: Fixed...

PR #149, opened 2017-02-15
AppVeyor: Ship MS Visual C++ DLLs with nightly builds

ubruhin:
This fixes runtime errors because of missing DLLs (e.g. msvcr120.dll) on
Windows without the Visual C++ Redistributable Package. See also
https://msdn.microsoft.com/en-us/library/dd293565(v=vs.110).aspx


ubruhin:
Seems to work...

PR #156, opened 2017-03-08
Travis: Make ccache also working with clang and OS X

### LINKED ISSUE #90: Improve CI build time by using ccache
According to https://docs.travis-ci.com/user/caching/#ccache-cache travis supports using ccache, which could speed up builds.

ubruhin:
Build times of the first run (caches for clang and osx not yet available):
auswahl_010

Build times after triggering a rebuild (caches from the first run are now available):
auswahl_011

Fixes #90


dbrgn:
Nice results :)

PR #159, opened 2017-03-11
Refactor/improve attributes system

ubruhin:
For the new library editor it was required to refactor the attributes system and add some helper classes to work with attributes.

**Changes:**

- Remove multilingual support for component attributes (it was complicated and not very useful).
- Make XML representation of attributes identical for projects and libraries.
- Projects can now have user-defined attributes.
- Create a new widget for modifying attributes, which will also be used in the new library editor.

Old widget:
auswahl_010

New widget:
auswahl_011

This branch needs to be rebased to `master` as soon as #158 is merged, because it depends on the new class `UndoStackTransaction`.


libs/librepcb/common/attributes/attributetype.cpp
```diff
@@ -69,6 +69,15 @@ const AttributeUnit* AttributeType::getUnitFromString(const QString& unit) const
         QString(tr("Unknown unit of attribute type \"%1\": \"%2\"")).arg(mTypeName, unit));
 }
 
+bool AttributeType::isUnitAvailable(const AttributeUnit* unit) const noexcept
+{
+    if (mAvailableUnits.isEmpty()) {
+        return (unit == nullptr);
+    } else {
+        return mAvailableUnits.contains(unit);
+    }
```
rnestler: The following would be clearer for me:
```C++
return (mAvailableUnits.isEmpty() && (unit == nullptr)) || mAvailableUnits.contains(unit);
```
But it would change the behavior slightly. Also maybe explain *why* it return true when the list is empty and the Unit is `nullptr`.

libs/librepcb/common/attributes/attributelist.cpp
```diff
@@ -0,0 +1,142 @@
...
+ ****************************************************************************************/
+namespace librepcb {
+
+/*****************************************************************************************
+ *  Constructors / Destructor
+ ****************************************************************************************/
+
+AttributeList::AttributeList() noexcept
+{
+}
```
rnestler: What's the point of an empty constructor? Maybe use `= default` in the header?

libs/librepcb/common/attributes/attributelist.cpp
```diff
@@ -0,0 +1,142 @@
...
+                QString(tr("The attribute \"%1\" exists multiple times in \"%2\"."))
+                .arg(attr->getKey(), domElement.getDocFilePath().toNative()));
+        }
+        mAttributes.append(attr);
+    }
+}
+
+AttributeList::~AttributeList() noexcept
+{
+}
```
rnestler: Same as above: What's the point of an empty destructor?

ubruhin: Just to be consistent over the whole project. You could see this as a design guideline ;)

libs/librepcb/common/fileio/filepath.cpp
```diff
@@ -265,10 +265,11 @@ QString FilePath::cleanFileName(const QString& userInput, CleanFileNameOptions o
     // replace spaces with underscore
     if (options.testFlag(CleanFileNameOption::ReplaceSpaces)) ret.replace(' ', '_');
     // remove all invalid characters
-    QString validChars("_\\-.0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz");
+    QString validChars(" _\\-.0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz");
```
rnestler: This change seems not really related to the commit. Also: This allows `\` but not `/`? Where is it used? This won't work with UNIX paths.

ubruhin: Oh that's quite strange ;)

libs/librepcb/common/attributes/attributetype.cpp
ubruhin: Actually the class `AttributeType` requires some refactoring anyway, so I think i will keep the current implementation and postpone the proposed improvements. I'm sorry ;)

libs/librepcb/common/fileio/filepath.cpp
ubruhin: Now I got it: The backslash is required to escape the minus sign because it is a reserved regex character ;)

Nevertheless I have created a separate PR to fix and improve this function: #160

PR #160, opened 2017-03-15
FilePath: Fix/improve method cleanFileName()

ubruhin:
- Spaces were replaced even if CleanFileNameOption::KeepSpaces was used
- Simplify regex: use character ranges, remove escaping for minus sign
- Add CleanFileNameOption::ToUpperCase
- Improve comments
- Add test

### DIFF
--- a/libs/librepcb/common/fileio/filepath.cpp
+++ b/libs/librepcb/common/fileio/filepath.cpp
@@ -260,16 +260,16 @@ QString FilePath::cleanFileName(const QString& userInput, CleanFileNameOptions o
 {
     // perform compatibility decomposition (NFKD)
     QString ret = userInput.normalized(QString::NormalizationForm_KD);
+    // remove all invalid characters
+    ret.remove(QRegularExpression("[^-._ 0-9A-Za-z]"));
     // remove leading and trailing spaces
     ret = ret.trimmed();
-    // replace spaces with underscore
+    // replace remaining spaces with underscore (if corresponding option set)
     if (options.testFlag(CleanFileNameOption::ReplaceSpaces)) ret.replace(' ', '_');
-    // remove all invalid characters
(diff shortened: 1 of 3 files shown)


libs/librepcb/common/fileio/filepath.cpp
```diff
@@ -260,16 +260,16 @@ QString FilePath::cleanFileName(const QString& userInput, CleanFileNameOptions o
 {
     // perform compatibility decomposition (NFKD)
     QString ret = userInput.normalized(QString::NormalizationForm_KD);
+    // remove all invalid characters (note: backslashes are required to escape minus sign)
+    ret.remove(QRegularExpression("[^ _\\-.0-9A-Za-z]"));
```
rnestler: Ahhh so its "\\-" because you need to escape the "\" for the C-String and then to escape the '-' in the regex :laughing:

rnestler: Don't you need to escape the '.' as well? '.' means any character, doesn't it? Or is this different inside a `[]` group?

ubruhin: [Wikipedia says](https://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended):
> Within POSIX bracket expressions, the dot character matches a literal dot. For example, a.c matches "abc", etc., but [a.c] matches only "a", ".", or "c".

I also found another interesting fact:
> The - character is treated as a literal character if it is the last or the first (after the ^, if present) character within the brackets: [abc-], [-abc]. Note that backslash escapes are not allowed.

If I understand correctly, backslash escapes are only allowed in [POSIX extended](https://en.wikipedia.org/wiki/Regular_expression#POSIX_extended) (which Qt seems to support, because the escaping works as expected).

But actually it may still be better to use following code:
```
ret.remove(QRegularExpression("[^-._ 0-9A-Za-z]"));
```

ubruhin: > But actually it may still be better to remove the backslashes:

Done.

Issue #157, opened 2017-03-09
Decide how to deal with invalid objects


ubruhin:
### Invalid attributes

There are many possibilities of invalid attributes of objects:
- Empty/invalid strings (for example forbidden characters in library element names)
- Negative numbers for line widths, drill diameters, pad sizes, and many more
- Invalid UUIDs, version numbers, ...
- And many more...

### Invalid combination of attributes

In addition, there are also *combinations* of attributes which I consider as (more or less) invalid:
- Footprints without any visible items (pads, polygons, ...)
- Components without any symbols
- Components which have multiple signals with the same UUID and/or name
- Pin-Signal-Mappings and Pad-Signal-Mappings which contain non-existing pins, pads or signals
- And many more...

Of course some of these invalid states can be avoided by validating new values before applying them, but some things are not really avoidable. For example when creating a new component, it won't have assigned any symbols yet, so there is a component without any symbols. Same for footprints, they are just empty right after creation.

Some things would be possible to avoid, but quite hard to implement. For example a component holds a list of its signals, where all signals must have unique names. But signals do not know each other and thus their method `setName(QString newName)` can't check whether there is already a signal with the same name.

These things are currently validated while saving the elements to their XML file. This was easy to implement, but it's quite annoying for users because they notice the errors only when trying to save the element. And even more annoying, it's currently impossible to save an invalid element, so if you want to stop working on a library element, you either need to fix all errors or discard all changes before you're able to close the library editor.

### Invalid attributes

I think it's important to avoid all smaller issues (negative numbers, invalid names, ...) as soon as possible:
1. Whenever possible, initialize attributes with valid, reasonable default values.
2. If 1. is not possible (for example there isn't a reasonable, valid default name for a newly created library element), these attributes must be passed to the constructor of the object to create. The constructor can then validate them and throw an exception if something is invalid.
3. All affected setter methods (for example `setLineWidth(int width)`) must not accept invalid values. They must throw an exception instead, with an expressive message (e.g. "The line width cannot be negative").

### Invalid combination of attributes

As many of these issues cannot be avoided with reasonable effort, maybe it's best to just accept them ;) So saving an invalid object (like a component which contains multiple signals with the same name) is possible. But there should be a way to validate such an element, for example add methods like `bool isValid()` and `QStringList getErrors()`, so the UI is able to display all issues of an invalid element.

@rnestler @dbrgn @0xB767B Any other ideas?

ubruhin:
[...]

PR #161, opened 2017-04-14
Refactor serialization

ubruhin:
The concept of serialization was inconsistent until now:
- To deserialize elements, the XML element tag names were defined outside the classes to be deserialized (e.g. the string "pin" was defined in the `SymbolPin`'s outer class `Symbol`).
- To serialize elements, the classes to be serialized defined their XML element tag names in their own function `serializeToXmlDomElement()` (e.g. the string "pin" was defined in the class `SymbolPin`).

So nearly every XML tag name was defined in two different classes. To make this more consistent, XML tag names are now always defined in the outer classes (as described in the first point above) and are passed to the inner classes as argument of `serializeToXmlDomElement()`. This way it is also possible to use different XML tag names for the same class, for example if a common/generic class is used in very different contexts.

Other changes:
- Replace the interface `IF_XmlSerializableObject` with the new class `SerializableObject`.
- Make `serializeToXmlDomElement()` no longer pure virtual and implement it in `SerializableObject` instead. From now on this is only a convenience method.
- Add the pure virtual method `serialize(XmlDomElement& root)`. In contrast to `serializeToXmlDomElement()` this new method allows to serialize multiple objects into the same DOM element (this is useful in some cases, e.g. to serialize a list of objects of mixed types).
- Remove `checkAttributesValidity()` because it was bad design (see #157).


dbrgn:
I can't judge the code, but from your description, it sounds good. And if more code is removed than added, then it *must* be good :wink:

ubruhin:
> And if more code is removed than added, then it must be good

Absolutely! :smile: 

But now I have added the new class template [`SerializableObjectList`](https://github.com/LibrePCB/LibrePCB/blob/refactor_serialization/libs/librepcb/common/fileio/serializableobjectlist.h) :wink:  This class is still work-in-progress, and my idea is to add even more such classes (see commit message of 18cbb6c1945d07fee0664ba1a0e05288637aca48).

At the moment, these classes increase the total LOC, but if they are widely used in future, the total LOC will be reduced for sure. But I'm still not sure if these classes can be designed generic enough to make them usable enough...

@dbrgn @rnestler Any thoughts on my first draft of [`SerializableObjectList`](https://github.com/LibrePCB/LibrePCB/blob/refactor_serialization/libs/librepcb/common/fileio/serializableobjectlist.h)?

ubruhin:
Unfortunately I was not as successful as hoped for with the new template class `SerializableObjectList`, so I decided to remove it again :unamused: 

Instead, I added some helper functions to easily serialize whole containers: https://github.com/LibrePCB/LibrePCB/blob/refactor_serialization/libs/librepcb/common/fileio/serializableobject.h#L88-L122

Now the overall balance between added and removed lines of code is even better :smiley:

ubruhin:
Ok the class `SerializableObjectList` is there again :see_no_evil: Now it seems that it works quite good (it is heavily used on the branch `add_library_editor`)....

PR #162, opened 2017-04-25
Fix non-deterministic content in XML files

ubruhin:
`QDomDocument` was not able to produce deterministic XML output because it stores attributes in a `QHash` and thus [their order is random](https://geidav.wordpress.com/2013/02/27/deterministic-attribute-order-in-xml-documents-in-qt-5/). This is very bad for version control systems because the diffs on these XML files is unnecessary large.

After some experiments with other XML libraries I noticed that `QXmlStreamWriter` does not have this issue and thus the generated XML files are deterministic. So I changed our serialization to use `QXmlStreamWriter` instead of `QDomDocument`.

Even if this PR does not introduce canonical XML as requested by #2, I think it should solve the issues with version control systems and thus #2 can be closed anyway.

### DIFF
--- a/libs/librepcb/common/fileio/xmldomdocument.cpp
+++ b/libs/librepcb/common/fileio/xmldomdocument.cpp
@@ -89,13 +89,18 @@ XmlDomElement& XmlDomDocument::getRoot(const QString& expectedName) const throw 
  *  General Methods
  ****************************************************************************************/
 
-QByteArray XmlDomDocument::toByteArray() const noexcept
+QByteArray XmlDomDocument::toByteArray() const throw (Exception)
 {
-    QDomDocument doc;
-    doc.implementation().setInvalidDataPolicy(QDomImplementation::ReturnNullNode);
-    doc.setContent(QString("<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>"));
-    doc.appendChild(mRootElement->toQDomElement(doc));
-    return doc.toByteArray(1); // indent only 1 space to save disk space
+    QByteArray data;
+    QXmlStreamWriter writer(&data);
+    writer.setAutoFormatting(true);
+    writer.setAutoFormattingIndent(1); // indent only 1 space to save disk space
+    writer.setCodec("UTF-8");
+    writer.writeStartDocument("1.0", true);
+    mRootElement->writeToQXmlStreamWriter(writer);
+    writer.writeEndDocument();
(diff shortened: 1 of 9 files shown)


dbrgn:
:sparkles:

ubruhin:
Now I also updated the whole [demo-workspace](https://github.com/LibrePCB/demo-workspace) submodule: https://github.com/LibrePCB/demo-workspace/commit/6fc3ab855f5445c1ab261d8f3e0b5d2072fd6744

The only XML node which still changes every time a project is saved, is the [last modified](https://github.com/LibrePCB/demo-workspace/blob/master/projects/Demo%20Brushless%20Controller/Demo%20Brushless%20Controller.lpp#L8) attribute in LibrePCB's project files :wink: Actually It would be great if this information is dynamically fetched from VCS, file attributes or something like that. But until this is implemented, I see two solutions:

- Keep going with the current solution (project files will be modified every time a project is saved).
- Remove the "last modified" attribute from project files, and in schematics just display the datetime of the time when the project was opened, and update it every time the project is saved.

dbrgn:
I think making "last modified" to be the date when that file was actually last modified (= on save) sounds useful  to me :)

ubruhin:
> I think making "last modified" to be the date when that file was actually last modified (= on save) sounds useful to me :)

So this would be the current solution which changes the project file on every save? The second solution from above does *not* use the correct "last modified" date until the project is saved again, because this attribute is stored nowhere. So when opening an old project, schematics show the current datetime instead of the datetime when the project was really last modified...

So we can choose between:
- Show correct "last modified" date time in schematics: Content of project files changes on every save, because this datetime needs to be stored and even version controlled.
- Show current datetime instead of "last modified" in schematics until the project is saved the first time after opening it: Project file does not change on save (if there were no explicit changes made).

dbrgn:
Ah, now I get it.

What about making the date a static field that can be updated with the current date easily? KiCad does it that way.

ubruhin:
> What about making the date a static field that can be updated with the current date easily?

Mmmh then it's very likely that this is forgotten sometimes. The result would be that exported schematic PDFs contain old datetimes (same one as from last export). IMO it would be less "dangerous" if a wrong datetime is too new instead of too old.

> KiCad does it that way.

That's a good reason to do it the other way :smile:

ubruhin:
I decided to remove the "last modified" datetime from project files and updated this branch accordingly. As soon as CI is successful, this should be ready to merge.

PR #170, opened 2017-07-21
Add Ratio class to enable support for single precision FPUs

ubruhin:
Until now, `qreal` floating point numbers were used to represent ratios (e.g. 25%) in the board design rules (for example to specify clearances). Unfortunately there were two issues with that concept:

- The precision of a `qreal` depends on the available FPU (either single precision or double precision). So the design rules were slightly platform dependent.
- The serialization and deserialization of a `qreal` in the class `XmlDomElement` was implemented only for double precision FPUs. Compiling for single precision FPUs was explicitly prevented by a static assert. 

This PR adds a new class `Ratio` which is used to represent, serialize and deserialize ratios in a platform independent way (the underlaying type is `qint32` instead of `qreal`). The precision is always 1ppm.

Thanks to this change, LibrePCB works now also with single precision FPUs, which was [successfully tested on the Open Pandora](https://repo.openpandora.org/?page=detail&app=librepcb). This is the first time LibrePCB runs on an ARM processor! :smiley:


dbrgn:
🎉

PR #183, opened 2017-09-04
Travis/AppVeyor: Compile with "-Werror" flag

ubruhin:
This ensures that pull requests do not create new compiler warnings.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ install:
 build_script:
   - mkdir build
   - cd build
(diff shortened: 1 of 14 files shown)


ubruhin:
Hooray, finally it works! :sparkles: 

Was not easy to fix all warnings because they highly depend on the used compiler, compiler version and Qt version. Some warnings occur only on older tools, some only on newer tools. No chance to reproduce them all locally :unamused:

PR #184, opened 2017-09-07
Refactor attributes system

### LINKED ISSUE #171: Review/Refactor Attributes System
LibrePCB has a quite sophisticated attribute system which allows to display texts in schematics and boards which are dynamically determined. For example in a symbol one can place a label with the text "${SYM::NAME}". As soon as such a symbol is added to a schematic, the text is replaced by the symbol's name (e.g. "R5").

But actually I'm not sure if this concept is really good:
- Maybe the syntax "${NAMESPACE::ATTRIBUTE}" could be simplified (IMO it looks ugly)?
- Do we really need to separate namespace (e.g. "SYM") and attribute (e.g. "NAME")? Maybe we could completely remove the namespace. If still required, the namespace could be integrated into the attribute name instead (e.g. "${SYM_NAME}").
- Should we introduce a strict separation between built-in attributes and user defined attributes? For example the attribute "${SYM::NAME}" is a built-in attribute which is implicitly defined by symbols, it's not listed in any attribute list which the user can edit. If the user would add an attribute "NAME" to a symbol, we actually would have a name conflict as there is already a built-in attribute with the same name! But how could a strict separation look like?

ubruhin:
Fixes #171.


ubruhin:
I think this should be ready to merge soon. Here the current state of the new attributes system:

- Added a [documentation about the attributes system](http://librepcb.org/LibrePCB-Doxygen/refactor_attributes_system/de/df4/doc_attributes_system.html)
- For simplicity, attribute namespaces (e.g. `${NAMESPACE::ATTRIBUTE}`) are no longer available
- Simplified syntax from `${NAMESPACE::ATTRIBUTE}` to `#ATTRIBUTE`
- Added support for [escaping the `#` character](http://librepcb.org/LibrePCB-Doxygen/refactor_attributes_system/de/df4/doc_attributes_system.html#doc_attributes_system_escaping)
- Added support for [multiple key substitution](http://librepcb.org/LibrePCB-Doxygen/refactor_attributes_system/de/df4/doc_attributes_system.html#doc_attributes_system_multiple_key_substitution)
- Updated the [lookup order](http://librepcb.org/LibrePCB-Doxygen/refactor_attributes_system/de/df4/doc_attributes_system.html#doc_attributes_system_lookup_order)
- Added user-defined attributes to the file format of device library elements and devices in boards. They are not yet usable in the GUI, so basically they are completely useless ;) I first want to work without them because I'm not 100% sure if the concept is good. But I already want to have them in the file format specified, so no new file format is needed if we enable them in the GUI.
- Added unit tests for the attribute substitution in strings. Some tests are currently disabled because of bugs in some edge cases, these should be fixed some time but for now it should be good enough ;)

PR #193, opened 2017-10-22
Switch from XML to S-Expressions file format

### LINKED ISSUE #192: Switch from XML to S-Expressions file format
In #150 we decided to switch from XML to S-Expressions as primary file format. This issue can be used for discussions about the details of the new file format.

I already have a first question to discuss: When should S-Expression values be quoted and when not?

According to [this IETF draft](http://people.csail.mit.edu/rivest/Sexp.txt), double quotes can optionally be omitted if the value does not contain strings. So following two lines are considered as equivalent:
```
(key value)
(key "value")
```

But IMO it would be better to always use double quotes if the represented value is a string, and omit the quotes only if the represented type is a boolean or number. Examples:
```
(number -12.34)
(boolean true)
(string "foobar")
```

This way one can "see" the data type of the values in an S-Expressions file. This makes it also easier to support syntax highlighting in editors, because the syntax highlighter can distinguish between strings, numbers and booleans. This is how syntax highlighting could look like for our files:

auswahl_015

Actually we can even go one step further: Should we define a set of basic data types used in S-Expression files? For example:
- Integer: A number without decimal point (e.g. `-42`)
- Float: A number with decimal point (e.g. `-42.0`)
- Boolean: Either `true` or `false`
- Enum item: A string without spaces, but not `true` or `false` (e.g. `align_top`)
- String: A string in double quotes (e.g. `"Hello World!"`)

A syntax highlighter could then easily distinguish between all of these data types, so a LibrePCB S-Expressions file would look really nice in editors ;) In addition, such a file is even more expressive because the data type of every value is visible, like this:

auswahl_014

ubruhin:
Example files:

- [Symbol](https://github.com/LibrePCB/demo-workspace/blob/switch_from_xml_to_sexpr/v0.1/libraries/local/LibrePCB_Base.lplib/sym/193ef70d-8dab-4a6c-a672-274c5bf09b68/symbol.lp)
- [Component](https://github.com/LibrePCB/demo-workspace/blob/switch_from_xml_to_sexpr/v0.1/libraries/local/LibrePCB_Base.lplib/cmp/d167e0e3-6a92-4b76-b013-77b9c230e5f1/component.lp)
- [Schematic](https://github.com/LibrePCB/demo-workspace/blob/switch_from_xml_to_sexpr/projects/Demo%20Brushless%20Controller/schematics/main.lp)
- [Circuit](https://github.com/LibrePCB/demo-workspace/blob/switch_from_xml_to_sexpr/projects/Demo%20Brushless%20Controller/core/circuit.lp)

Fixes #192


dbrgn:
Shouldn't dates be quoted? E.g. see [here](https://github.com/LibrePCB/demo-workspace/blob/switch_from_xml_to_sexpr/v0.1/libraries/local/LibrePCB_Base.lplib/sym/193ef70d-8dab-4a6c-a672-274c5bf09b68/symbol.lp#L5)

dbrgn:
Some other remarks:

```sexp
(name (locale "") "R-US")
(description (locale "") "Resistor American (IEEE 315)")
(keywords (locale "") "resistor,resistance")
```

I don't quite get that. Shouldn't that be `(keywords "" "resistor,resistance")`? Or if you want to make the locale optional, `(keywords "resistor,resistance")`?

I also wonder whether in some cases more newlines would be better:

```sexp
(signal 6d776f4d-2a7c-4128-a98a-dbb1dd861411 (name "2") (role passive)
  (required true) (negated false) (clock false) (forced_net "")
)
```

Maybe this would be better like this:

```sexp
(signal 6d776f4d-2a7c-4128-a98a-dbb1dd861411
  (name "2")
  (role passive)
  (required true)
  (negated false)
  (clock false)
  (forced_net "")
)
```

It would also help when doing diffs.

Regarding this one:

```sexp
(disp none)
```

`none` is an enum variant, not `null`, right?

What about flags?

```sexp
(signal 1c1c7abc-7b40-4f92-b533-f65604644db7 (name "1") (role passive)
  (required true) (negated false) (clock false) (forced_net "")
)
```

Maybe `required` could become a flag?

```sexp
(signal 1c1c7abc-7b40-4f92-b533-f65604644db7 required (name "1") (role passive)
  (negated false) (clock false) (forced_net "")
)
```

On the other hand it might make the format less generic and harder to parse.

And a general comment, I'd use two spaces for indentation, one is a bit hard to see when dealing with larger structures.

ubruhin:
> Shouldn't dates be quoted?

I saw no reason to quote them because the format of dates is well known (contains only letters, digits, :, - and +) and thus is always a valid S-Expression token.

ubruhin:
> I don't quite get that. Shouldn't that be (keywords "" "resistor,resistance")? Or if you want to make the locale optional, (keywords "resistor,resistance")?

I expected that question :D I already spent some time to improve the format of localized strings (name, description, keywords), but unfortunately I didn't found a good solution.

Actually the main question is: How should the localizing of these attributes work? My current idea is that for each library element there are fields for "name", "description" and "keywords" without a specific locale. So if you create libraries only for yourself, you can for example fill these fields with German words. In addition, translated strings can optionally be added, e.g. for your libraries someone could add English translations.

For that system, the most reasonable file format would probably look like this:
```
(name "Hallo Welt")
(name (locale "en_US") "Hello World")
```

But this is really ugly to parse, because the text is not always at the same index of the node (index 0 without locale, index 1 with locale). Of course it would be possible to either take the count of child nodes, or the type of each child node into account, but it feels like a workaround to do that.

If the order of the child nodes is swapped, parsing gets easier, but IMO the file looks less intuitive:
```
(name "Hallo Welt")
(name "Hello World" (locale "en_US"))
```

Another solution is to always write down the locale, but without an extra child node, but this also looks strange:
```
(name "" "Hallo Welt")
(name "en_US" "Hello World")
```

Maybe it would make sense to make the locale mandatory, so every string has a locale defined?
```
(name (locale "de_CH") "Hallo Welt")
(name (locale "en_US") "Hello World")
```
But this would bring up a new question: Which locale should be used if the user's locale is not found? So if you have set your default locale to "de_DE", which of the two available locales "de_CH" and "en_US" should be displayed? When having default/fallback name/description/keywords, these will be used in such cases.

Do you have a good answer for these questions? :)

> I also wonder whether in some cases more newlines would be better:
> It would also help when doing diffs

It helps doing diffs on short files. But when adding more newlines, many files get *much* longer (e.g. >1000 lines), which is also not nice. I think creating nice looking files is a trade-off between length of lines and count of lines.

> And a general comment, I'd use two spaces for indentation

Of course this would improve readability, but here I also see a trade-off between file size and readability. Optimal for file size would be zero indentation (and even no newlines), optimal for readability ~2 spaces. So for me, 1 space seems to be a good compromise (keep in mind that you will probably *never* write such files by hand, and even reading such files should not be daily business).

> none is an enum variant, not null, right?

Right.

> What about flags?

Looks interesting, need to think about it... ;)

ubruhin:
Ok now I updated the format of localized strings by making the "locale" node optional:

```
(name "Hallo Welt")
(name (locale "en_US") "Hello World")
```

IMO the files look much better this way, but of course parsing feels a bit crappy now...

What do you think @dbrgn?

Btw, the links from the first post above always point to the latest file format, just check out these files again to see the difference.

dbrgn:
Don't you initially parse the S-Expression into a generic nested key-value tree? In that case parsing would not make a difference.

In general I like the approach with the optional locale node.

ubruhin:
> Don't you initially parse the S-Expression into a generic nested key-value tree?

Yes, I do.

> In that case parsing would not make a difference.

"Low-level" parsing does not make a difference, deserializing of objects is a bit ugly. See 3717032dc7664866994e7641030bdfc20b444323, this is the change needed to make the "locale" node optional.

libs/librepcb/common/fileio/sexpression.cpp
```diff
@@ -0,0 +1,310 @@
...
+{
+    if (isList()) {
+        return mValue;
+    } else {
+        throw FileParseError(__FILE__, __LINE__, mFilePath, -1, -1, QString(),
+                             tr("Node is not a list."));
+    }
+}
+
+QList<SExpression> SExpression::getChilds(const QString& name) const noexcept
```
dbrgn: Shouldn't that be `getChildren`?

dbrgn:
> this is the change needed to make the "locale" node optional

That isn't too bad though :)

libs/librepcb/common/fileio/sexpression.cpp
dbrgn: Welcome to the brave world of stemming and pluralization 😆

ubruhin:
I think this should be ready now! :sparkles: 

I have even added the new mime type [`application/x-librepcb-file`](https://github.com/LibrePCB/LibrePCB/blob/switch_from_xml_to_sexpr/share/mime/packages/librepcb.xml) and a [language specification](https://github.com/LibrePCB/LibrePCB/blob/switch_from_xml_to_sexpr/dev/gtksourceview-3.0/language-specs/librepcb-file.lang) to enable syntax highlighting of *.lp files in GtkSourceView based editors.

@dbrgn @rnestler Someone interested in providing a syntax highlighting config for Vim? Or does Vim also support the GtkSourceView language specs?

PR #194, opened 2017-10-27
Add more functionality to the component chooser dialog

ubruhin:
When adding a component to the schematic, it was not yet possible to choose a specific device. Devices were not even displayed in the dialog, so the user had no idea which devices (resp. packages) are available for the chosen component.

With this PR, all devices (and their packages) of every listed component are now also shown in the dialog. The user can now even choose a specific device (instead of a component), which is then automatically used when adding the component to the board later. But this is optional, the user can still only choose a component and then choose the exact device some time later.

In addition, this PR also adds a very powerful search field to the dialog, which allows to search in the whole workspace library for components and devices! This way the user is able to find components and devices very quickly.

add component_010

### DIFF
--- a/libs/librepcb/common/common.pro
+++ b/libs/librepcb/common/common.pro
@@ -71,6 +71,7 @@ SOURCES += \
     geometry/hole.cpp \
     geometry/polygon.cpp \
     geometry/text.cpp \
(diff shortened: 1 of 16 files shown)


ubruhin:
Should now be ready.

PR #195, opened 2017-11-07
Rename "item" to "gate" in components and schematics

ubruhin:
A component (resp. a symbol variant of it) consists of one or more independent symbols (for example an OpAmp could have one power symbol and one amplifier symbol). Such a symbol is currently called `ComponentSymbolVariantItem` in the source code, and `item` in files (resp. `lib_item` in schematic files).

But actually I wonder why I didn't name it `gate`, like other tools do :thinking:

I think `gate` is more expressive than `item` (?), so I would just rename that now...

### DIFF
--- a/libs/librepcb/library/cmp/componentsymbolvariant.cpp
+++ b/libs/librepcb/library/cmp/componentsymbolvariant.cpp
@@ -62,6 +62,11 @@ ComponentSymbolVariant::ComponentSymbolVariant(const SExpression& node) :
     // Load all symbol variant items
     mSymbolItems.loadFromDomElement(node);
 
+    // backward compatibility, remove this some time!
+    foreach (const SExpression& node, node.getChildren("item")) {
+        mSymbolItems.append(std::make_shared<ComponentSymbolVariantItem>(node)); // can throw
+    }
+
     if (!checkAttributesValidity()) throw LogicError(__FILE__, __LINE__);
 }
 
--- a/libs/librepcb/library/cmp/componentsymbolvariantitem.h
+++ b/libs/librepcb/library/cmp/componentsymbolvariantitem.h
@@ -51,6 +51,8 @@ namespace library {
  *  - UUID
  *  - Symbol UUID
  *  - Pin-signal-mapping
+ *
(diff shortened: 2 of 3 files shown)


dbrgn:
Hm, I like this in general ("put the thing in the item and do some stuff with the object"), but why the word gate? (Maybe I just don't understand the context though.)

ubruhin:
> but why the word gate?

Probably it comes from [logic gate](https://en.wikipedia.org/wiki/Logic_gate) where for example a 4-channel AND gate consists of 4 single channel gates. 

Even if today such a symbol does not always represent a **logic** gate, I still like the word "gate" more than "item" because many EDA users probably already know this term.

ubruhin:
In Altium this [seems to be called "part"](http://www.altium.com/documentation/17.1/display/ADES/((Understanding+Models,+Components+and+Libraries))_AD#!UnderstandingModelsComponentsandLibraries-BreakingtheComponentintoMultipleParts) but I don't like that term because I plan to use it in another context...

Now I struggle whether we should really change it to "gate" or to stay with "item" ;)

@rnestler Any opinion?

0xB767B:
Just for the sake of another view to the issue... Thinking in terms of software design, could we think of it as composite? If I understand you correctly, you mean an opamp is composed of several parts. Wikipedia states about the composite pattern:

"The intent of a composite is to "compose" objects into tree structures to represent part-whole hierarchies."

But anyway, I think you should stick to common terminology used in EDA programs. Like that you are talking to users in already known terms.

sebgonzo:
Personally I don't like `gate` name. It is OK only for logic gates, but for opamps, microcontrolers and other multi part symbols is misleading. What about 'subsymbol' or 'subpart' name?

ubruhin:
OK now I'm completely unsure which term to use :confused: So maybe just stay with `item`...

ubruhin:
So, I decided to merge this PR because LibrePCB already uses Eagle terminology for many things, and with using "gate" (like Eagle) instead of "item", this should make users switching from Eagle to LibrePCB even more easy. Even if "gate" may not be absolutely perfect, it's still *much* better than "item"...

PR #211, opened 2017-12-01
Device editor: Sort pad names numerically, then alphanumerically

### LINKED ISSUE #210: Library editor: First sort tables numerically, then alphanumerically
plese change the sort of the list

it's so easy to make mistake to set 1-9 with insert 10-12
plese fix it, thx much more

right with the correction
correct

dbrgn:
I have no idea what I'm doing 😂 

screenshot

Fixes #210

### DIFF
--- a/tests/common/toolboxtest.cpp
+++ b/tests/common/toolboxtest.cpp
@@ -1,0 +1,74 @@
+/*
+ * LibrePCB - Professional EDA for everyone!
+ * Copyright (C) 2017 LibrePCB Developers, see AUTHORS.md for contributors.
+ * http://librepcb.org/
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*****************************************************************************************
+ *  Includes
+ ****************************************************************************************/
+#include <QtCore>
+#include <gtest/gtest.h>
+#include <librepcb/common/toolbox.h>
+
+/*****************************************************************************************
+ *  Namespace
+ ****************************************************************************************/
+namespace librepcb {
+namespace tests {
+
+/*****************************************************************************************
+ *  Test Class
+ ****************************************************************************************/
+
+class ToolboxTest : public ::testing::Test
+{
+};
+
+/*****************************************************************************************
+ *  Test Methods
+ ****************************************************************************************/
+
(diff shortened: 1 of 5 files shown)


ubruhin:
Cool! :smiley: 

Actually there would also be another solution, without the need for creating a new class :wink: 
http://www.archivum.info/qt-interest@trolltech.com/2007-02/00076/Re-Sort-numerically-QTableWidget.html

For me, both ways are acceptable - you can decide which way to choose :)

But I think it would be reasonable to move this logic to [libs/librepcb/common/](https://github.com/LibrePCB/LibrePCB/tree/master/libs/librepcb/common) because it may also be helpful for some table widgets in the project editor.

ubruhin:
> So we'd have to try to convert it to an integer and conditionally add it using setData instead of setText

We can always use `setData` instead if `setText`. The only difference is that in one case we need to pass a `QVariant` which represents a string, and in the other case we need to pass a `QVariant` which represents an integer.

We could create a helper function in [`Toolbox`](https://github.com/LibrePCB/LibrePCB/blob/master/libs/librepcb/common/toolbox.h) which takes a `QString` and returns a `QVariant` which represents either a string or an integer, depending on whether the passed string was a number or not. The caller code could then look like this:

```
item->setData(Qt::DisplayRole, Toolbox::stringOrNumberToQVariant(padName));
```

dbrgn:
Sounds good!

ubruhin:
Btw, there are some more tables which could be improved the same way:
- `ComponentSignalListEditorWidget`
- `FootprintListEditorWidget`
- `PackagePadListEditorWidget`

@dbrgn It would be awesome if could adjust them too :heart_eyes:

dbrgn:
@ubruhin like this? If it's OK, I can remove the first two commits by rebasing and also add the code to the other tables you mentioned.

tests/common/toolboxtest.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+class ToolboxTest : public ::testing::Test
+{
+};
+
+/*****************************************************************************************
+ *  Test Methods
+ ****************************************************************************************/
+
+TEST(ToolboxTest, testNumeric)
```
ubruhin: Unfortunately in the test output one can't see that this test is related to the method `stringOrNumberToQVariant`:

```
[----------] 3 tests from ToolboxTest
[ RUN      ] ToolboxTest.testNumeric
[       OK ] ToolboxTest.testNumeric (0 ms)
[ RUN      ] ToolboxTest.testNonNumeric
[       OK ] ToolboxTest.testNonNumeric (0 ms)
[ RUN      ] ToolboxTest.testMixed
[       OK ] ToolboxTest.testMixed (0 ms)
[----------] 3 tests from ToolboxTest (0 ms total)
```

I see different ways how to fix that:
1. Change the name of the test cases to:
  - `testStringOrNumberToQVariant_numeric`
  - `testStringOrNumberToQVariant_nonNumeric`
  - `testStringOrNumberToQVariant_mixed`
2. Put all three tests into one test with the name `testStringOrNumberToQVariant`
3. Rename the whole test (and its file) to `ToolboxStringOrNumberToQVariantTest`

dbrgn:
Ok, tests are updated and the PR is rebased.

I did not change the three widgets you mentioned as discussed, because they are editable and sorting is tricky.

ubruhin:
Great, thanks! :smiley: 

I will merge it as soon as the CI succeeded...

PR #213, opened 2017-12-11
Convert about dialog to QML

dbrgn:
This is a first attempt at integrating QML based views into LibrePCB (#17).

The QML files are included as Qt resources, so they don't need to be distributed separately.

Data is passed to the `QQuickView` using `setContextProperty`. Maybe the version stuff could be wrapped into a separate `QObject`. I think QML objects can also expose their own slots and signals for full interaction with C++ code.

Layout is not very pretty yet, I haven't figured out how to do left/top margin yet.

Some initial feedback would be nice 🙂

### DIFF
--- a/apps/librepcb/controlpanel/controlpanel.cpp
+++ b/apps/librepcb/controlpanel/controlpanel.cpp
@@ -23,6 +23,8 @@
 #include <QtCore>
 #include <QtWidgets>
 #include <QFileDialog>
+#include <QQuickView>
+#include <QQmlContext>
 #include "controlpanel.h"
 #include "ui_controlpanel.h"
 #include <librepcb/workspace/workspace.h>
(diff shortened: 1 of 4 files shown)


dbrgn:
What I gathered so far in general about QtWidgets vs QtQuick 2 (note that this is just gathering resources, I'm not proposing to rewrite the LibrePCB UI):

### Pro QtQuick

- Modern, maintained
- Can be ported to mobile (this will be quite hard for QtWidgets)
- Support for HighDPI screens (I don't think that works well with QtWidgets)
- Works fine on Wayland (I read somewhere that QtWidgets is buggy on Wayland)
- Declarative UI, so easier to maintain
- Support for transitions / animations (can aid usability in some cases)
- Resources are loaded at runtime, this can make development easier (UI changes without recompile) and enables dynamic UIs

### Cons QtQuick

- Some controls might look a bit less native on certain platforms (the [Fusion Style](https://doc.qt.io/qt-5/qtquickcontrols2-styles.html) should be enough native-looking though)
- Not available on Qt 5.2 (Xenial has 5.5 though)
- If we want to stick with Xenial, then we can only use [Quick Controls 1, not 2, which was introduced in Qt 5.7](https://doc.qt.io/qt-5/qtquickcontrols2-differences.html). QQC2 is more efficient, especially on embedded (not that we care much about that).
- Resources are loaded at runtime, so this might be a slight performance difference (they're cached though)

### Resources

- https://www.reddit.com/r/kde/comments/6a46rx/qt_widgets_vs_qt_quick_for_developing_an_app_for/
- https://doc.qt.io/qt-5/qtquickcontrols2-differences.html

---

List might be expanded later.

dbrgn:
@ubruhin any decision about this yet? :)

ubruhin:
At the moment I think QtQuick/QML would introduce more problems than it solves for LibrePCB. It would probably require high effort to solve problems which we didn't have with QtWidgets. And because our highest priority task is to bring out a first stable release as soon as possible, we shouldn't introduce new technologies which aren't really required (and even no developer of LibrePCB is familiar with).

Maybe we can take a deeper look at QtQuick/QML some time in future, after bringing out the first official release of LibrePCB. Especially the [Qt Quick Scene Graph](http://doc.qt.io/qt-5/qtquick-visualcanvas-scenegraph.html) looks very interesting (as a replacement for QGraphicsScene/QGraphicsView), maybe it would be useful for us to create better schematic/board editors. See also this related blog post: [Should you still be using QGraphicsView?](https://blog.qt.io/blog/2017/01/19/should-you-be-using-qgraphicsview/)

So, thanks @dbrgn for taking a first look at QML, but I hope you understand that I close this PR now :wink:

dbrgn:
Sure! :)

PR #227, opened 2018-01-25
README: Clarify configuration of Qt Creator

dbrgn:
I think this makes it a bit clearer for people new to Qt Creator.

Kliment (in IRC) was a bit confused at first, and I remember that I also had some issues when I first tried to open LibrePCB.


kliment:
Looks good to me

PR #228, opened 2018-01-25
Disable unimplemented menu items and icons in the schematic and board editor

### LINKED ISSUE #189: Remove, hide or disable all useless GUI elements
In some GUIs there are widgets which do not yet have a functionality (because not yet implemented). All these widgets should be removed, hidden or disabled before publishing the first release of LibrePCB because they can be very annoying for users (looks like LibrePCB is buggy :wink:).

Here an incomplete list of such widgets:
- Schematic/BoardEditor: Toolbuttons for clipboard functionality (cut/copy/paste)
- Schematic/BoardEditor: Toolbutton and menu item "Print" (--> try to implement it!)
- Schematic/BoardEditor: Menu item "Help"
- SchematicEditor: Toolbutton to open the layer stack
- SchematicEditor: Buttons to reorder schematic pages (--> try to implement it!)
- SchematicEditor: Checkbox for mirroring symbols
- SchematicEditor: Tab "General" in project settings dialog
- BoardEditor: Menu item "PDF export"
- LibraryEditor: Checkboxes of component signal properties "negated" and "clock"

kliment:
Fixes #189 
Also makes the board layer stack setup dialog in the board editor resizeable on all platforms by explicitly enabling the size grip

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/boardeditor.ui
+++ b/libs/librepcb/projecteditor/boardeditor/boardeditor.ui
@@ -109,6 +109,9 @@
     <addaction name="actionToolDrawPolygon"/>
    </widget>
    <widget class="QMenu" name="menuHelp">
+    <property name="enabled">
+     <bool>true</bool>
+    </property>
     <property name="title">
      <string>Help</string>
     </property>
@@ -255,6 +258,9 @@
    </property>
   </action>
   <action name="actionPrint">
+   <property name="enabled">
+    <bool>false</bool>
+   </property>
    <property name="icon">
     <iconset resource="../../../../img/images.qrc">
      <normaloff>:/img/actions/print.png</normaloff>:/img/actions/print.png</iconset>
@@ -264,6 +270,9 @@
    </property>
    <property name="shortcut">
     <string>Ctrl+P</string>
+   </property>
+   <property name="visible">
+    <bool>false</bool>
    </property>
   </action>
   <action name="actionQuit">
@@ -285,6 +294,9 @@
    </property>
    <property name="text">
     <string>PDF Export</string>
+   </property>
+   <property name="visible">
+    <bool>false</bool>
    </property>
   </action>
   <action name="actionShowControlPanel">
@@ -340,6 +352,9 @@
    <property name="shortcut">
     <string>F1</string>
    </property>
+   <property name="visible">
+    <bool>false</bool>
+   </property>
   </action>
   <action name="actionRotate_CCW">
    <property name="icon">
@@ -376,6 +391,9 @@
    <property name="shortcut">
     <string>Ctrl+C</string>
    </property>
+   <property name="visible">
+    <bool>false</bool>
+   </property>
   </action>
   <action name="actionCut">
    <property name="icon">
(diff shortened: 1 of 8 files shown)


libs/librepcb/projecteditor/boardeditor/boardeditor.ui
```diff
@@ -109,6 +109,9 @@
     <addaction name="actionToolDrawPolygon"/>
    </widget>
    <widget class="QMenu" name="menuHelp">
+    <property name="enabled">
```
ubruhin: I would change "visible" to false instead of "enabled", because IMO it looks strange to have a disabled menu item.

libs/librepcb/projecteditor/dialogs/projectsettingsdialog.ui
```diff
@@ -19,12 +19,24 @@
...
+     <property name="tabsClosable">
+      <bool>false</bool>
+     </property>
+     <property name="tabBarAutoHide">
+      <bool>true</bool>
+     </property>
      <widget class="QWidget" name="tabGeneral">
       <attribute name="title">
-       <string>General</string>
+       <string/>
```
ubruhin: There is still an empty tab, but now even with an empty name:
project settings_010

So, just completely remove this tab, it's easy to add it again once we need it.

libs/librepcb/projecteditor/schematiceditor/schematiceditor.ui
```diff
@@ -61,6 +61,9 @@
     <addaction name="actionQuit"/>
    </widget>
    <widget class="QMenu" name="menuHelp">
+    <property name="enabled">
```
ubruhin: Same here -> replace set "visible" to false instead of "enabled".

libs/librepcb/projecteditor/schematiceditor/schematiceditor.ui
```diff
@@ -313,6 +319,9 @@
    <property name="text">
     <string>PDF Export</string>
    </property>
+   <property name="visible">
+    <bool>true</bool>
+   </property>
```
ubruhin: Just a note: When modifying widget properties with QtCreator, it leads to changes in *.ui files even if the property was changed back to the default value (like here, `visible=true` is actually the default). To avoid cluttering up commits with such useless changes, this can be reverted by clicking on the button at the right of the property field:
auswahl_014

libs/librepcb/projecteditor/schematiceditor/symbolinstancepropertiesdialog.ui
```diff
@@ -186,6 +186,9 @@
         </item>
         <item row="5" column="0">
          <widget class="QLabel" name="label_9">
+          <property name="enabled">
+           <bool>false</bool>
+          </property>
```
ubruhin: Could you maybe just completely remove the this label and its corresponding checkbox? I think it may confuse users if there is a disabled checkbox. And adding it again later is quite easy...

kliment:
Reverting changes to default values in qtcreator doesn't seem to work for me, but other than that I believe I've addressed all issues here

ubruhin:
Thanks for the update! I have permitted myself to rebase your commits, to do some small cleanup in code and commit messages and to fix the CI error ;) I hope you agree with these changes.

I think this is ready to merge as soon as the CI succeeds.

Issue #230, opened 2018-02-07
Support for compiling for Qt > 5.9 


NZSmartie:
I've successfully compiled LibrePCB using Qt 5.7. But, as of writing this; Qt 5.10 is available which is the toolchain I installed initially on Windows. 

When building for Qt 5.10 Mingw, the build fails with the following error:
```
error: specialization of 'template<class T>QtPrivate::QForeachContainer' class in different namespace [-fpermissive]
```

The source of the issue is https://github.com/LibrePCB/LibrePCB/blob/3de6686162b204ce7296847a8d29447ce35705d4/libs/librepcb/common/fileio/serializableobjectlist.h#L376-L394

There isn't much information about `QForeachContainer`  but the issue was about the namespace being different. 

My my proposed fix was to surround the block with the new `QtPrivte` namespace instead of prefixing QForeachContainer:

```c++
/*****************************************************************************************
 * Prevent from using SerializableObjectList in a foreach loop because it always would
 * create a deep copy of the list! You should use C++11 range based for loops instead.
 ****************************************************************************************/

#if (QT_VERSION > QT_VERSION_CHECK(5, 9, 0))
namespace QtPrivate {
#endif

template <typename T, typename P>
class QForeachContainer<librepcb::SerializableObjectList<T, P>> { public:
    ~QForeachContainer() = delete;
};
template <typename T, typename P>
class QForeachContainer<const librepcb::SerializableObjectList<T, P>> { public:
    ~QForeachContainer() = delete;
};


#if (QT_VERSION > QT_VERSION_CHECK(5, 9, 0))
} // namespace QtPrivate
#endif
```

Since i'm new to LibrePCB, and not sure of the scope of this change, can someone else verify this compiles with Qt > 5.9 and functionality isn't broken?

NZSmartie:
Further investigation shows that this issue appears to be specific with Windows. as macOS (OSX) builds are already happening with Qt 5.10. I've yet to verify if this also happens on Debian builds

NZSmartie:
I used docker to create a Qt 5.10 build environment and observed the same compile error.  mentioned above and resolved with the fix as well.

`Dockerfile`
```Dockerfile
FROM rabits/qt:5.10-desktop

LABEL description="LibrePCB debian build environment for Qt 5.10"

RUN sudo apt-get update -qq && sudo apt-get install -y \
    git \
    build-essential \
    libglu1-mesa-dev \
    openssl \
[...]

PR #231, opened 2018-02-07
Support Qt 5.10

NZSmartie:
See #230

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,7 +8,7 @@ deploy: off
 environment:
   global:
     MINGW: C:\Qt\Tools\mingw530_32
-    QTDIR: C:\Qt\5.7\mingw53_32
+    QTDIR: C:\Qt\5.9.3\mingw53_32
     VCRT_DIR: 'C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\Redist\x86\Microsoft.VC120.CRT'
 
 init:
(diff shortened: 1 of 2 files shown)


NZSmartie:
So, the PPA that provides up to date builds for Qt Toolchain only provides Qt 5.10 for Ubuntu Xenial (16.04) https://launchpad.net/~beineri/+archive/ubuntu/opt-qt-5.10.0-xenial. 

However, Travis CI do not provide a build environment for Xenial because of maintenance costs. https://github.com/travis-ci/travis-ci/issues/5821

`.travis.yml` has a build matrix setup for multiple Qt builds, but I wont be able to add one for Qt 5.10

NZSmartie:
Atleast Qt 5.9 is available for Trusty and already part of travis-ci builds. AppVeyor was at 5.7 which I had bumped that up to 5.10 but could bump it back down to 5.9 to keep the version numbers consistent across platform builds. 
That way if a bug occurs or there's a API issue, we can rule out Qt being different for different platforms

ubruhin:
Great, thanks! Looks good for me, but it seems that on AppVeyor the test `QLiteDatabaseTest.testConcurrentAccessFromMultipleThreads` now takes massively longer (~14s instead of ~10s) and thus the test fails. I retriggered the build to see if it fails again...

ubruhin:
OK, now with Qt 5.9.3 it works also on AppVeyor :+1:

PR #232, opened 2018-02-08
Fix compile error on single precision FPUs

Luz:
This fix makes the code compilable on the openpandora again.


ubruhin:
Thanks :smiley:

PR #234, opened 2018-02-12
Support building Windows targets on Linux build host

NZSmartie:
A few minor issues when trying to build Windows binaries from a Linux Host (i,e. Qt with MinGW on Ubuntu).

1. `#include <Windows.h>` should be `#include <windows.h>` as Linux paths are case sensitive and is all lowercase in MinGW.
1. `lrelease.exe` command not found is caused by https://github.com/LibrePCB/LibrePCB/blob/a0a45e188242a05480c80b505726b85a0351a06b/apps/librepcb/librepcb.pro#L111-L115
   where it is incorrectly assumed the target is also the build host. Instead, checking `QMAKE_HOST.os` to compare host platform to `Windows` or `Linux`.


apps/librepcb/librepcb.pro
```diff
@@ -110,8 +110,11 @@ FORMS += \
 
 # Custom compiler "lrelease" for qm generation
 isEmpty(QMAKE_LRELEASE) {
-    win32: QMAKE_LRELEASE = $$[QT_INSTALL_BINS]\lrelease.exe
-    else: QMAKE_LRELEASE = $$[QT_INSTALL_BINS]/lrelease
+    contains(QMAKE_HOST.os,"Windows"){
```
ubruhin: I'm not sure, but maybe we could even completely remove the windows specific configuration of `QMAKE_LRELEASE`, because under Windows the `.exe` suffix is often not required. So this may even work for both Windows and Linux:

```
isEmpty(QMAKE_LRELEASE): QMAKE_LRELEASE = $$[QT_INSTALL_BINS]/lrelease
```

Could you maybe try this? Travis and AppVeyor can tell us if it breaks something, but they don't check if it also works with cross-compiling...

NZSmartie:
Updated and looks good. Heh overlooked how simple the fix actually was

ubruhin:
Thanks @NZSmartie for the fix! 😃 

For me it looks good, do you also think it's ready to rebase (on to the current `master`) and merge?

PR #239, opened 2018-02-18
Add initial support for copper planes in boards

### LINKED ISSUE #196: Add support for copper planes in boards
Copper planes are actually quite important, but not yet implemented in LibrePCB. Probably the most complicated part of planes is to calculate all segments of the plane, while every segment must satisfy various requirements:

- Keep clearance to other nets or objects (according to design rules)
- Ensure a minimum width at any point in the plane (remove plane segments which are too thin)
- Optionally connect pads by thermal relief
- Optionally remove segments which are not connected to any net (orphans)

I think this can basically be done with following steps:
1. Take plane outline polygon:
auswahl_012
2. Take the outline of every object which must not be connected to the plane, and grow these outlines by their required clearance width:
auswahl_015
3. Subtract these objects from the outline:
auswahl_016
4. Shrink (offsetting) the resulting segments by the minimum width value:
auswahl_017
5. Expand (offsetting) the resulting segments by the minimum width value:
auswahl_018
6. Fill the resulting segments:
auswahl_019
7. Result (No thermal pads, keep orphans):
auswahl_020

I used the [clipper library](http://angusj.com/delphi/clipper.php) for these operations, seems to work pretty good. The only disadvantage is that it doesn't support arcs in polygons, only straight lines are allowed. So we need to convert arcs to many small segments of straight lines :( Unfortunately I didn't find a library which can handle arcs...

Btw, I found some interesting papers about polygon offsetting:
- https://seant23.files.wordpress.com/2010/11/anoffsetalgorithm.pdf
- http://www.me.berkeley.edu/~mcmains/pubs/DAC05OffsetPolygon.pdf

ubruhin:
This adds an initial implementation for copper planes:

auswahl_010

There are still some limitations:
- No clearance values from design rules (because we don't yet have design rules for that)
- No thermal pads (because it seems to be quite complicated to calculate them, so I'm not yet sure how to implement it exactly)

But I think even with these limitations, planes are quite usable for many PCB projects, so I think we can fix them some time later ;)

Fixes #196


dbrgn:
Woohoo 🙂

kliment:
The algorithm for calculating thermals is:
1. Take the set of pads, and enlarge them by the isolation size of the pour
2. Subtract the enlarged shape from the pour
3. For each of the four sides of the pad, add a line from the pad to the pour if there is a pour area where the line would join

Issue #132, opened 2016-11-07
Docker: Simplify X11 Setup


0xB767B:
It seems as the ugly xauth/sed hack in build_container.sh can be simplified using xhost +local:

See: https://github.com/0xB767B/dockerfiles/tree/master/qtcreator

It might as well not be the safest solution, but at least more understandable what the script is doing.

I'll try it on LibrePCB at some point.

ubruhin:
> I'll try it on LibrePCB at some point.

:+1:

ubruhin:
See also #152.

ubruhin:
FYI, the Dockerfile will be removed in #647.

PR #240, opened 2018-03-04
Docker: Add "troubleshooting" section to instructions

trdenton:
added some steps I had to perform to get the container working on my system.  May be helpful for other docker newbies

### DIFF
--- a/dev/docker/README.md
+++ b/dev/docker/README.md
@@ -36,3 +36,9 @@ Different examples to run the docker container "librepcb:git":
 ./run_container.sh dia                      # run Dia (diagram editor)
 ```
 
+#### Troubleshooting
+
+* "QXcbConnection: Could not connect to display :0"
+    * Give docker the rights to access the X-Server with: `xhost +local:docker`
+* "error gathering device information while adding custom device "/dev/dri/card0"
+    * install a newer version of docker (appears to be working as of 17.12.1-ce)


dev/docker/README.md
```diff
@@ -29,10 +29,17 @@ Build the docker container "librepcb:git":
...
 ./run_container.sh                          # enter the bash
 ./run_container.sh qtcreator librepcb.pro   # open Qt Creator
 ./run_container.sh librepcb                 # run LibrePCB
 ./run_container.sh dia                      # run Dia (diagram editor)
 ```
 
+#### Troubleshooting
+
+* "QXcbConnection: Could not connect to display :0"
+    * Give docker the rights to access the X-Server with: `xhost +local:docker`
```
ubruhin: Hmm, AFAIK this shouldn't be required if the docker container is created with the `--privileged` flag, which is the case in the `build_container.sh` script. For me it works properly when just executing the `build_container.sh` and `run_container.sh` scripts, doesn't that also work for you?

In addition, isn't the `xhost +local:docker` only valid during one X session, so after a reboot you need to execute this command again? Then it would maybe make sense to add it to the `run_container.sh` script. See also https://github.com/LibrePCB/LibrePCB/issues/132.

trdenton: it did not work for me, but this may have been because i was on an earlier version of docker - I will retry from scratch now that I am on the latest version

trdenton: Ah, I get some xauth related error when the image builds:
`xauth:  unable to link authority file /tmp/.docker.xauth, use /tmp/.docker.xauth-n`.  This is then resolvable via `xhost +local:docker`

ubruhin:
I rebased your branch to `master` and made the commit message more expressive :)

Will be merged as soon as the CI succeeds. Thanks for the pull request @trdenton!

PR #242, opened 2018-03-04
Docker: set USER env var

### LINKED ISSUE #241: SystemInfoTest.testGetUsername fails in docker container
Just tested on eee604fe2932b97e55faf88f1c0df6cab2a5fd23 with a fresh docker image install.  Running the unit tests through qt creatore, I see one failure:

[  FAILED  ] SystemInfoTest.testGetUsername

It appears the the environment vars `USER` / `USERNAME` are not set in the container, which is what getUsername is relying on.  It seems this is more of a 'docker configuration'-type issues than a 'software correction' -type issue

trdenton:
this fixes SystemInfoTest.testGetUsername in the docker container

fixes #241 

inspired by https://vsupalov.com/docker-build-time-env-values/


ubruhin:
Makes sense, thanks! :smiley:

PR #243, opened 2018-03-19
Increase max length of library directory paths

dbrgn:
Fixes LibrePCB/librepcb-rfcs#9.


ubruhin:
Thanks, but it seems that this is not really necessary (see https://github.com/LibrePCB/librepcb-rfcs/issues/9#issuecomment-374241944), so I'll close this. Sorry :wink:

PR #246, opened 2018-04-05
Implement stroke texts for footprints and boards

### LINKED ISSUE #165: Decide how to implement fonts in footprints and boards
One big problem is still to draw texts in boards in a way which also makes Gerber export possible. So the text needs to be convertible into simple polygons.

I think the text should not be converted into polygons just for the Gerber export, but even for displaying the text in the board editor. This way we can guarantee that the Gerber output is exactly the same as what is displayed in the board editor (WYSIWYG).

But which system/library should we use for converting text into simple polygons (a series of straight lines and arcs)? Most important would be to have a simple vector font (like many EDA tools), but it would be absolutely awesome if any TrueType font could be used (like Altium)!

From [Gerber specifications](https://www.ucamco.com/files/downloads/file/81/the_gerber_file_format_specification.pdf):
> 4.17 Text in the Image
Gerber has no native font support – this adds too much complication in relation to the modest
text requirements in PCB fabrication data. Text must be represented as pure image data, best
as regions (contours), see 4.12. Painting the text area is anathema!
Font definitions often contain splines and Gerber does has linear and circular segments but no
splines. The splines must therefore be approximated to either linear or circular segments.
Circular is more precise with less objects than linear, but mathematically more complicated.

See also #153 which is about fonts in schematics. Maybe we could find a solution which is fine for schematics (where "pretty" fonts are needed) and boards (where Gerber export is needed).

ubruhin:
Note: There are still some [open questions](https://github.com/LibrePCB/LibrePCB/issues/165#issuecomment-378917465) to clarify before this is ready to merge. But here already an explanation about this new feature:

board

## Stroke Font

Until now we used TrueType fonts in footprints/boards, but they cannot be exported into the Gerber format without extremely high effort. So we need stroke fonts (like most CAD/EDA applications). But because we were not happy with existing stroke font technologies, we started to develop specifications for a new stroke font: https://github.com/fontobene/fontobene/blob/master/SPECIFICATION.md

As these specifications will be ready soon, I started to implement them in LibrePCB. They can now be used in footprints and boards. Schematics still use TrueType fonts as no Gerber export is required.

## Application Font Pool

Now the application has a [directory](https://github.com/LibrePCB/LibrePCB/tree/fontobene/share/librepcb/fontobene) where FontoBene files are located. They are loaded at runtime, so they need to be deployed together with the application. The library editor uses the default application font when editing footprints.

## Project Font Pool

To remove the dependency of projects/boards to the application-wide fonts (and thus make projects even more self-contained), projects copy the application-wide fonts into the directory `resources/fontobene` inside the project's directory. This is done only once, afterwards boards of a project will only use these project-wide fonts. Changing application-wide fonts will thus have no effect on all existing projects. That's important to not break boards when updating the application (and its fonts) to a newer version.

In addition, this even allows users to use other fonts for their boards :) Other EDA tools do not provide this flexibility as fonts are often strongly bundled with the application.

## Default Board Font

Currently there is only one font available, but in future we may provide various fonts and allow the user to specify what font should be used for every text object placed on a board. So we need to specify which one is the default font. That's now a property of every board, so you can choose the default font for every board separately.

## Board Text Objects

Users can now place standalone text objects on boards:

demo brushless controller lpp - librepcb board editor_021

To modify these text objects, there's a property dialog:

demo brushless controller lpp - librepcb board editor_022

## Footprint Text Objects

Footprints (in the library) typically contain text objects for name and value. When adding a footprint to a board, all its contained text objects are now *copied* as separate text objects (*not* just referenced), decoupled from the footprint. The original footprint text objects are *not* shown in a board. This way the decoupled text objects can be modified (e.g. moved/rotated) separately.

auswahl_023

Footprint texts can also be removed from a board if not used. Maybe some time we can add a "lock text objects" property to footprints to avoid accidentally removing or modifying text objects.

## Board File Changes

In the demo workspace you can see what has changed in the board file: https://github.com/LibrePCB/demo-workspace/commit/94942f5bf5d46dd1f6fbf0914178297193020a3c#diff-8482b174ff3f28dcc855f119f2294404

Of course a board file is now a lot larger. But I think that's worth it...

## Feedback?

Feedback is very welcome! You can already check this branch out to test the new features. It's not too late to make some changes :wink: 

Fixes #165.


dbrgn:
Great work @ubruhin! Looking good.

I like this font btw. Kudos to the LibreCAD developers, upon whose work the FontoBene format (and the "unicode" font) is based (I think mostly @Rallaz, maybe also @dxli or @lordofbikes).

ubruhin:
Update:
- Rebased to `master`
- Use absolute stroke width instead of a ratio (e.g. `0.2mm` instead of `20%`)
- Add stroke width to letter spacing (adjust letter spacing dynamically to stroke width)

ubruhin:
Update:
- Rebased to `master`
- Fixed several issues
- Use automatic line spacing (line spacing specified by font plus stroke width)
- Allow to manually override letter spacing and line spacing (ignoring values provided by font): 
stroke text properties_010

Now we "only" need a font with a suitable license to finish this feature...

ubruhin:
Update: Switch font to "FontoStroke" (converted NewStroke, see https://github.com/fontobene/fontobene-fonts/pull/4) because of the unknown license situation of the LFF fonts. NewStroke has CC0 license, so it can be used for any purpose.

ubruhin:
Fixed some issues encountered in the new font (https://github.com/fontobene/fontobene-converters/pull/3) and renamed "FontoStroke" to "NewStroke Bene" (https://github.com/fontobene/fontobene-fonts/pull/4).

I think now it should be ready to merge.

dbrgn:
I can do some testing tomorrow.

dbrgn:
Seems to work great, unless you add too many characters :) Probably some algorithm with exponential runtime. Would be worth profiling.

ubruhin:
Thanks for testing! Yeah performance improvements could be done later if really needed...

PR #249, opened 2018-04-06
BoardEditor: Save device properties on enter

dbrgn:
The class was missing an `accept` override.

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/deviceinstancepropertiesdialog.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/deviceinstancepropertiesdialog.cpp
@@ -109,9 +109,7 @@ void DeviceInstancePropertiesDialog::buttonBoxClicked(QAbstractButton* button) n
             applyChanges();
             break;
         case QDialogButtonBox::AcceptRole:
-            if (applyChanges()) {
-                accept();
-            }
+            accept();
             break;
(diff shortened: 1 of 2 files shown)


libs/librepcb/projecteditor/boardeditor/deviceinstancepropertiesdialog.h
```diff
@@ -73,6 +73,7 @@ class DeviceInstancePropertiesDialog final : public QDialog
     private: // Methods
         void buttonBoxClicked(QAbstractButton* button) noexcept;
         void keyPressEvent(QKeyEvent* e);
+        void accept();
```
ubruhin: Could you please add the keywords `noexcept override`?

libs/librepcb/projecteditor/boardeditor/deviceinstancepropertiesdialog.cpp
```diff
@@ -135,6 +135,13 @@ void DeviceInstancePropertiesDialog::keyPressEvent(QKeyEvent* e)
     }
 }
 
+void DeviceInstancePropertiesDialog::accept()
+{
+    if (applyChanges()) {
```
ubruhin: This is good so far. But in `buttonBoxClicked()` above, the `if (applyChanges())` can now be removed to avoid executing `applyChanges()` twice.

Issue #215, opened 2017-12-18
Add link to official Gerber reference viewer


dbrgn:
In the docs and in the gerber export dialog you recommend gerbv. However, I've had experiences with gerbv where it would render gerbers on a wrong way (even with examples directly from the reference).

I'd rather recommend Gerbview (part of KiCad, docs [here](http://docs.kicad-pcb.org/stable/en/gerbview.html)) and/or the [online reference viewer](https://gerber.ucamco.com/).

ubruhin:
Hmm okay, actually I was happy with gerbv until now (more than with Gerbview). Do you have example files which gerbv doesn't render properly? But yes, the online reference viewer is probably a good candidate to recommend.

marmei:
What do you think about kayout? https://github.com/klayoutmatthias/klayout. It is native Qt.

ubruhin:
> What do you think about kayout?

I have never heard about it, so I spent a few minutes to give it a try but unfortunately I was not even able to open a Gerber file with it :wink: Did you try it too?

marmei:
Yes Indeed, Klayout is a professional tool for chip design. I used it for MEMS design and it is much better suited than all commercial tools I have tried/provided by my university.
However I do not know if it is straight forward to combine it directly with librePCB.

ubruhin:
OK but I don't see any reason to "combine" it with LibrePCB, it's something completely different. This issue is just about recommending LibrePCB users a Gerber viewer which they could use to review the generated Gerber files. And until now I never saw a better Gerber viewer than `gerbv`...

marmei:
BTW - excellent talk at fosdem! 
In my opinion you are are paving the most promising way to create an open source PCB editor in commercial grade.
Just wanted add this link for viewing gerber files in Klayout. http://klayout.de/klayout_screenshot5_pcb.png

ubruhin:
> BTW - excellent talk at fosdem!
In my opinion you are are paving the most promising way to create an open source PCB editor in commercial grade.

Great to hear, thank you! :)

> Just wanted add this link for viewing gerber files in Klayout. http://klayout.de/klayout_screenshot5_pcb.png

Sure that this screenshot shows an imported Gerber file?

[...]

PR #255, opened 2018-04-22
Improve PCB fabrication output generator

### LINKED ISSUE #73: Improve fabrication output (gerber/excellon) generator
- [x] Make the whole fabrication output generator more generic/configurable (until now, many things are hardcoded)
- [ ] ~Add more Gerber X2 attributes to the output files (e.g. attributes for layers, pads, ...)~
 *postponed*

ubruhin:
~~*Note: Depends on #254, so this one must not yet be merged.*~~

- Make Gerber/Excellon export settings configurable (output path, file suffixes, ...)
- Save these settings in board files
- Add support for inner copper layers
- Allow to generate separate drill files for NPTH and PTH
- Fix wrong export of NPTH drills (they appeared on all layers)
- Make output files reproducible by sorting all objects by their UUID
- Add link to official reference Gerber viewer

This is how the fabrication output generator dialog now looks like:

fabrication output generator_010

Fixes #73, #215.

### DIFF
--- a/libs/librepcb/project/boards/board.cpp
+++ b/libs/librepcb/project/boards/board.cpp
@@ -51,6 +51,7 @@
 #include <librepcb/library/pkg/footprint.h>
 #include "items/bi_polygon.h"
 #include "boardlayerstack.h"
+#include "boardfabricationoutputsettings.h"
 #include "boardusersettings.h"
 #include "boardselectionquery.h"
 #include "../circuit/netsignal.h"
@@ -89,6 +90,10 @@ Board::Board(const Board& other, const FilePath& filepath, const QString& name) 
 
         // copy design rules
         mDesignRules.reset(new BoardDesignRules(*other.mDesignRules));
+
+        // copy fabrication output settings
+        mFabricationOutputSettings.reset(new BoardFabricationOutputSettings(
+                                             *other.mFabricationOutputSettings));
 
         // copy user settings
         mUserSettings.reset(new BoardUserSettings(*this, *other.mUserSettings));
@@ -149,6 +154,7 @@ Board::Board(const Board& other, const FilePath& filepath, const QString& name) 
         qDeleteAll(mNetSegments);       mNetSegments.clear();
         qDeleteAll(mDeviceInstances);   mDeviceInstances.clear();
         mUserSettings.reset();
+        mFabricationOutputSettings.reset();
         mDesignRules.reset();
         mGridProperties.reset();
         mLayerStack.reset();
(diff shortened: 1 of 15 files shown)


sebgonzo:
@dbrgn 

I think better instead adding additional checkbox use new Jinja syntax? E.g.

- layer-in-name: `{{PCB_LAYER_NAME}}.gbr`
- layer-in-extension: `{{PROJECT_NAME}}.{{GERBER_EXT}}`

ubruhin:
> I think the intro text in the dialog should mention the Gerber version ("Extended Gerber X2" or "RS-274X X2").

:+1:

> Also, since we don't yet have support for manufacturer profiles, maybe we can just add a toggle button where you can switch between encode-layer-in-name (XXX.gbr) and encode-layer-in-extension (project.XXX)? Would be very useful since with that you could already cover most manufacturers without manually changing every layer filename.

An alternative would be to just use Protel file extensions as default, which probably most PCB manufacturers accept. Then no additional GUI widgets are needed...

> I think better instead adding additional checkbox use new Jinja syntax.
> E.g.
> 
>    layer-in-name: {{PCB_LAYER_NAME}}.gbr
>    layer-in-extension: {{PROJECT_NAME}}.{{GERBER_EXT}}

@sebgonzo I don't understand how you mean that, where would PCB_LAYER_NAME and GERBER_EXT be defined?

sebgonzo:
@ubruhin 
My idea is to add built-in attributes:
- `{{PCB_LAYER_NAME}}` - it will substitute with layer name taken from Board Editor layer names (e.g. "Top Copper", "Bottom Copper" etc.)
- `{{GERBER_EXT}}` - it will substitute with Gerber files extensions like in Protel format ("GTL", "GBL" etc.)

Using this attributes in textboxes user can construct it's own Gerber filename template.
For example:
`{{PCB_LAYER_NAME}}.gbr` will translate to `Top Copper.gbr`, `Bottom Copper.gbr`
`{{PROJECT}}.{{GERBER_EXT}}` will translate to `My Project.GTL`, `My Project.GBL`

dbrgn:
@sebgonzo if I'm not mistaken this might already be possible. At least @ubruhin certainly intended to implement support for placeholders when I talked to him yesterday.

ubruhin:
@sebgonzo I think this doesn't really work, because the board layer names do not always correspond to gerber layer names. For example in the drill file there are vias and pads (so, two different layers in one gerber file), or on silkscreen you can configure which of the layers "Top/Bot Placement", "Top/Bot Names" and "Top/Bot Values" should be exported.

@dbrgn Yes, placeholders like "{{PROJECT}}" or "{{LAYER}}" (for the index of inner copper layers) are already supported (see screenshot above), but I don't plan to support more placeholders for the gerber export.

ubruhin:
## Update
- Rebased to master (now we have support for Jinja attributes syntax)
- Add support for solder paste layers (creating Gerber files for stencils)
- Add buttons to easily switch between two predefined configurations (*.gbr vs. Protel file extensions)
  - It's not the cleanest solution, but I think it should be enough until we create a *real* output generator...

## Default configuration
fabrication output generator_010

## Protel configuration
fabrication output generator_011

PR #258, opened 2018-04-29
Add ability to place NPTH drills on boards

### LINKED ISSUE #253: Add ability to place NPTH drills on boards
Currently only vias (PTH) can be placed, but no NPTH drills.

ubruhin:
~~*Note: Depends on #255, so this one must not yet be merged.*~~

Now it's possible to place NPTH drills directly on a board.
- They are subtracted from copper planes (with clearance)
- They are exported into the NPTH drills Excellon file

demo brushless controller lpp - librepcb board editor_014

Fixes #253.


dbrgn:
Nice. Is the clearance configurable?

Issue #12, opened 2015-07-28
Add "Contribution" section to readme


ubruhin:
List some different ways how to contribute to LibrePCB, for example:
- coding: fix issues (starting with issues labeled with "easy")
- design/coding: create a website? design icons?
- donate: [Gratipay](https://gratipay.com/)?
- donate: PayPal?
- donate: Bitcoin?
- other ways to contribute?

also an IRC channel would be nice...

rnestler:
You probably want to also crate a file [`CONTRIBUTING.md`](https://github.com/blog/1184-contributing-guidelines) with contributing guidelines.

PR #261, opened 2018-05-13
About Dialog

dbrgn:
This is a new about dialog for LibrePCB. Fixes most of #17.

Checklist from that issue:

- [x] Replace the `QMessageBox` with an own class derived from `QDialog`. This new class should be called `AboutDialog` and created in the directory [`libs/librepcb/common/dialogs/`](https://github.com/LibrePCB/LibrePCB/tree/master/libs/librepcb/common/dialogs)
- [x] Add the new about dialog to the "Help" menu of all main windows
  - [x] ControlPanel
  - [x] LibraryEditor
  - [x] SchematicEditor
  - [x] BoardEditor
- [x] Add more information to the new dialog, maybe splitted up in multiple tabs:
  - [x] A small description about LibrePCB (inclusive Logo)
  - [x] A link to LibrePCB's website (librepcb.org) and to the GitHub project site
  - [x] The application's version / commit hash (determined by `git describe`)
  - [x] Maybe the datetime of the last commit (-> build datetime)
  - [x] The license (a link to http://www.gnu.org/licenses/gpl-3.0.html, or the whole license text)
  - [x] Options for contribution (Bitcoin address, link to Patreon)
  - [x] Maybe a "how to contribute" section (see also #12)
  - [ ] ~~Maybe the rendered content of the [AUTHORS.md](https://github.com/LibrePCB/LibrePCB/blob/master/AUTHORS.md) file~~

Here's a screenshot:

screenshot

(The missing window decorations are due to my window manager.)

@ubruhin do you think we should add tabs, for things like "contributing" and the author list?b

### DIFF
--- a/apps/librepcb/controlpanel/controlpanel.cpp
+++ b/apps/librepcb/controlpanel/controlpanel.cpp
@@ -105,6 +105,8 @@ ControlPanel::ControlPanel(Workspace& workspace) :
     //        &Workspace::instance(), &Workspace::openLibraryEditor);
     connect(mUi->actionAbout_Qt, &QAction::triggered,
             qApp, &QApplication::aboutQt);
+    connect(mUi->actionAbout, &QAction::triggered,
+            qApp, &Application::about);
     connect(mUi->actionWorkspace_Settings, &QAction::triggered,
             &mWorkspace.getSettings(), &WorkspaceSettings::showSettingsDialog);
     connect(mLibraryManager.data(), &LibraryManager::openLibraryEditorTriggered,
@@ -422,12 +424,6 @@ void ControlPanel::projectEditorClosed() noexcept
 /*****************************************************************************************
  *  Actions
  ****************************************************************************************/
-
-void ControlPanel::on_actionAbout_triggered()
-{
-    AboutDialog* aboutDialog = new AboutDialog(this);
-    aboutDialog->exec();
-}
 
 void ControlPanel::on_actionNew_Project_triggered()
 {
--- a/apps/librepcb/controlpanel/controlpanel.cpp
+++ b/apps/librepcb/controlpanel/controlpanel.cpp
@@ -38,6 +38,7 @@
 #include <librepcb/projecteditor/projecteditor.h>
 #include <librepcb/projecteditor/newprojectwizard/newprojectwizard.h>
 #include <librepcb/common/application.h>
+#include <librepcb/common/dialogs/aboutdialog.h>
 #include <librepcb/common/fileio/fileutils.h>
 #include "../markdown/markdownconverter.h"
 #include "projectlibraryupdater/projectlibraryupdater.h"
@@ -424,14 +425,8 @@ void ControlPanel::projectEditorClosed() noexcept
 
 void ControlPanel::on_actionAbout_triggered()
 {
-    QMessageBox::about(this, tr("About LibrePCB"), QString(tr(
-        "<h1>About LibrePCB</h1>"
-        "<p>LibrePCB is a free & open source schematic/layout-editor.</p>"
-        "<p>Version: %1 (%2)</p>"
-        "<p>Please see <a href='http://librepcb.org/'>librepcb.org</a> for more information.</p>"
-        "You can find the project on GitHub:<br>"
-        "<a href='https://github.com/LibrePCB/LibrePCB'>https://github.com/LibrePCB/LibrePCB</a>"))
-        .arg(qApp->getAppVersion().toPrettyStr(3), qApp->getGitVersion()));
+    AboutDialog* aboutDialog = new AboutDialog(this);
+    aboutDialog->exec();
 }
 
 void ControlPanel::on_actionNew_Project_triggered()
--- a/libs/librepcb/common/application.h
+++ b/libs/librepcb/common/application.h
@@ -91,6 +91,9 @@ class Application final : public QApplication
         static Application* instance() noexcept;
 
 
+    public slots:
+        static void about() noexcept;
+
     private: // Data
         Version mAppVersion;
         QString mGitVersion;
--- a/libs/librepcb/common/application.h
+++ b/libs/librepcb/common/application.h
@@ -50,7 +50,7 @@ class StrokeFontPool;
  ****************************************************************************************/
 
 /**
- * @brief The Application class extends the QApplication with the exception-save method
+ * @brief The Application class extends the QApplication with the exception-safe method
  *        #notify()
  *
  * @author ubruhin
@@ -71,6 +71,7 @@ class Application final : public QApplication
         // Getters
         const Version& getAppVersion() const noexcept {return mAppVersion;}
         const QString& getGitVersion() const noexcept {return mGitVersion;}
+        const QDateTime& getBuildDate() const noexcept {return mBuildDate;}
         const Version& getFileFormatVersion() const noexcept {return mFileFormatVersion;}
         const FilePath& getResourcesDir() const noexcept {return mResourcesDir;}
         FilePath getResourcesFilePath(const QString& filepath) const noexcept;
@@ -93,6 +94,7 @@ class Application final : public QApplication
     private: // Data
         Version mAppVersion;
         QString mGitVersion;
+        QDateTime mBuildDate;
         Version mFileFormatVersion;
         FilePath mResourcesDir;
         QScopedPointer<StrokeFontPool> mStrokeFontPool; ///< all application stroke fonts
--- a/libs/librepcb/common/dialogs/aboutdialog.cpp
+++ b/libs/librepcb/common/dialogs/aboutdialog.cpp
@@ -1,0 +1,114 @@
+/*
+ * LibrePCB - Professional EDA for everyone!
+ * Copyright (C) 2018 LibrePCB Developers, see AUTHORS.md for contributors.
+ * http://librepcb.org/
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*****************************************************************************************
+ *  Includes
+ ****************************************************************************************/
+#include "aboutdialog.h"
+#include "ui_aboutdialog.h"
+#include <librepcb/common/application.h>
+
+/*****************************************************************************************
+ *  Namespace
+ ****************************************************************************************/
+namespace librepcb {
+
+AboutDialog::AboutDialog(QWidget* parent) noexcept :
+    QDialog(parent),
+    mUi(new Ui::AboutDialog)
+{
+    // Set up base dialog
+    mUi->setupUi(this);
+
+    // Event handlers
+    connect(mUi->buttonBox, &QDialogButtonBox::clicked,
+            this, &AboutDialog::close);
+
+    // Layout
+    mUi->tabWidget->setCurrentIndex(0);
+
+    // Get some version information
+    const Version& appVersion = qApp->getAppVersion();
+    const QString& gitVersion = qApp->getGitVersion();
+    const QString& buildDate = qApp->getBuildDate().toString("yyyy-MM-dd hh:mm:ss (t)");
+
+    // Dynamic text
+    mUi->textVersion->setText(QString("Version: %1 (%2)<br>Build date: %3").arg(appVersion.toPrettyStr(3), gitVersion, buildDate));
+    mUi->textLinks->setText(tr("For more information, please check out <a href='%1'>librepcb.org</a> or our <a href='%2'>GitHub repository</a>.").arg("http://librepcb.org/", "https://github.com/LibrePCB/LibrePCB"));
+    mUi->textContributeFinancially->setText(
+        tr("Support sustainable development of LibrePCB by donating financially, either via <a href='%1'>Patreon</a> or via <a href='%2'>Bitcoin</a>!")
+        .arg("https://www.patreon.com/librepcb", "bitcoin:1FiXZxoXe3px1nNuNygRb1NwcYr6U8AvG8")
+    );
+    mUi->textContributeCode->setText(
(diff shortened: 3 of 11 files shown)


ubruhin:
Nice!

Regarding the UI:
- I would put version number and build date on separate lines to increase readability.
- The version string (incl. git hash) should be selectable by mouse, so it can be copy&pasted (helpful for example to write down the exact version number when opening an issue).

> @ubruhin do you think we should add tabs, for things like "contributing" and the author list?

IMHO for contributing guidelines and authors it should be enough to add links to CONTRIBUTING.md and AUTHORS.md of this repository, then we probably don't need tabs...

apps/librepcb/controlpanel/controlpanel.cpp
```diff
@@ -411,14 +412,11 @@ void ControlPanel::projectEditorClosed() noexcept
...
-        "<p>LibrePCB is a free & open source schematic/layout-editor.</p>"
-        "<p>Version: %1 (%2)</p>"
-        "<p>Please see <a href='http://librepcb.org/'>librepcb.org</a> for more information.</p>"
-        "You can find the project on GitHub:<br>"
-        "<a href='https://github.com/LibrePCB/LibrePCB'>https://github.com/LibrePCB/LibrePCB</a>"))
-        .arg(qApp->getAppVersion().toPrettyStr(3), qApp->getGitVersion()));
+    const Version& appVersion = qApp->getAppVersion();
+    const QString& gitVersion = qApp->getGitVersion();
+    const QString& buildDate = qApp->getBuildDate();
+    AboutDialog* aboutDialog = new AboutDialog(this, appVersion, gitVersion, buildDate);
```
ubruhin: - Why passing version information to the about dialog? If the dialog is also used in the other main windows (schematic/board/library editors), this code needs to be copied each time.
- The AboutDialog can be created on the stack, no need to use heap for that.

img/images.qrc
```diff
@@ -83,6 +83,7 @@
         <file>actions/flip_horizontal.png</file>
         <file>actions/flip_vertical.png</file>
         <file>app/librepcb.png</file>
+        <file>app/librepcb64.png</file>
```
ubruhin: Why a new image, isn't it the same as librepcb.png?

libs/librepcb/common/application.h
```diff
@@ -90,6 +91,7 @@ class Application final : public QApplication
     private: // Data
         Version mAppVersion;
         QString mGitVersion;
+        QString mBuildDate;
```
ubruhin: As it is a date/time, I would use QDateTime for that.

libs/librepcb/common/dialogs/aboutdialog.cpp
```diff
@@ -0,0 +1,46 @@
...
+    headerLicense->setContentsMargins(0, 16, 0, 0);
+    textLinks->setOpenExternalLinks(true);
+    textLicense->setOpenExternalLinks(true);
+    ui->aboutContentLayout->addWidget(textIntro);
+    ui->aboutContentLayout->addWidget(headerVersion);
+    ui->aboutContentLayout->addWidget(textVersion);
+    ui->aboutContentLayout->addWidget(headerLinks);
+    ui->aboutContentLayout->addWidget(textLinks);
+    ui->aboutContentLayout->addWidget(headerLicense);
+    ui->aboutContentLayout->addWidget(textLicense);
```
ubruhin: Why creating all widgets by code instead of using the designer?

libs/librepcb/common/dialogs/aboutdialog.h
```diff
@@ -0,0 +1,27 @@
...
+    ~AboutDialog();
+
+private slots:
+    void on_buttonBox_clicked(QAbstractButton *button);
+
+private:
+    Ui::AboutDialog *ui;
+};
+
+#endif // ABOUTDIALOG_H
```
ubruhin: For the whole header file: Please take a look at similar headers, for example `stroketextpropertiesdialog.h`.
- Missing license header
- Missing namespace
- Missing default/copy constructors and assignment operator
- Coding style in general (indentation, `noexcept` keyword, ...)
- Member `ui` could be a `QScopedPointer`

sorry ;)

dbrgn:
> I would put version number and build date on separate lines to increase readability.

Can do that.

> The version string (incl. git hash) should be selectable by mouse, so it can be copy&pasted (helpful for example to write down the exact version number when opening an issue).

Then ideally we should make all text selectable for consistency, right?

> IMHO for contributing guidelines and authors it should be enough to add links to CONTRIBUTING.md and AUTHORS.md of this repository, then we probably don't need tabs...

Maybe I'll give it a try to check out how it looks, otherwise I'll just add the links.

apps/librepcb/controlpanel/controlpanel.cpp
dbrgn: > Why passing version information to the about dialog?

I could not access `qApp` inside the dialog even when including the header file. Maybe you can try it? I probably just overlooked something.

img/images.qrc
dbrgn: It's scaled down to 64x64px. Qt apparently cannot scale background images.

libs/librepcb/common/dialogs/aboutdialog.cpp
dbrgn: Because of the dynamic content in the strings. How would I do that with the designer?

libs/librepcb/common/dialogs/aboutdialog.h
dbrgn: Yeah, I used the generator and did not *yet* add the license header etc. But it was 4:30 in the morning and I was tired, so I decided to go to sleep instead :stuck_out_tongue:

dbrgn:
I just noticed that I forgot to mention that this is a WIP pull request :) I'm not done yet with the checklist.

ubruhin:
> Then ideally we should make all text selectable for consistency, right?

Yeah, why not (doesn't matter for me).

> I just noticed that I forgot to mention that this is a WIP pull request :)

Good to know xD

apps/librepcb/controlpanel/controlpanel.cpp
ubruhin: I guess the problem is that `AboutDialog` is not (yet) in the `librepcb` namespace.

dbrgn: Ahh, that could be :)

img/images.qrc
ubruhin: Normally I use a QLabel (not QWidget) for displaying images. Just set the "pixmap" and "scaledContents" properties and set the label to a fixed size.

libs/librepcb/common/dialogs/aboutdialog.cpp
ubruhin: All the widgets can be created and configured in the designer. Then set reasonable widget names (for example `lblLicenseText`). Dynamic texts can then be set in code like that:

    ui->lblLicenseText->setText(...);

Of course all static texts (like your `textIntro`) should be filled directly in the designer, no need to do this by code.

libs/librepcb/common/dialogs/aboutdialog.h
ubruhin: Ah, I see :)

img/images.qrc
dbrgn: Done

libs/librepcb/common/dialogs/aboutdialog.cpp
dbrgn: If you look at the current version, I used two helper methods to style headings and text. Is there some way of adding "styles" to widgets to set parameters like `contentsMargins` or `openExternalLinks` without repeating yourself?

libs/librepcb/common/application.h
dbrgn: Done

libs/librepcb/common/dialogs/aboutdialog.h
dbrgn: Style guide should be applied now.

dbrgn:
Ok, this is mostly done now. New screenshot:

tab1

tab2

Still to do:

- Add the new about dialog to the "Help" menu of all main windows (ControlPanel, SchematicEditor, BoardEditor)
- Maybe add the rendered content of the AUTHORS.md file

Open questions:

- Should we move the version number to the always-visible top, right below the title ("About LibrePCB")?

ubruhin:
> Ok, this is mostly done now. New screenshot:

Nice! :+1: 

> Should we move the version number to the always-visible top, right below the title ("About LibrePCB")?

Yeah, maybe this would make sense. At least work a try.

libs/librepcb/common/dialogs/aboutdialog.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+/*****************************************************************************************
+ *  Private Methods
+ ****************************************************************************************/
+
+void AboutDialog::on_buttonBox_clicked(QAbstractButton *button) noexcept
+{
+    Q_UNUSED(button);
+    this->close();
+}
```
ubruhin: This whole method could be removed by adding that to the constructor:

    connect(mUi->buttonBox, &QDialogButtonBox::clicked,
                this, &AboutDialog::close);

libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
```diff
@@ -501,6 +503,12 @@ void BoardEditor::on_actionRebuildPlanes_triggered()
     }
 }
 
+void BoardEditor::on_actionAbout_triggered() noexcept
```
ubruhin: This method is called two times when clicking on the menu item, so the dialog appears two times:
- One time from the `connect()` in the constructor
- One time from the implicit connection created by Qt because the method name has the pattern "on_WIDGETNAME_SIGNALNAME()".

I don't like the implicit connection because it is error prone. So better rename the method.

Another option would be to completely remove this method and provide a slot `about()` in the class `Application`, just like [`QApplication::aboutQt`](https://doc.qt.io/qt-5/qapplication.html#aboutQt). The pointer to the parent widget can be obtained with [`QApplication::activeWindow()`](http://doc.qt.io/qt-5/qapplication.html#activeWindow). This would simplify opening the about dialog to this:

    connect(mUi->actionAbout, &QAction::triggered, qApp, &Application::about);

libs/librepcb/common/dialogs/aboutdialog.ui
```diff
@@ -0,0 +1,315 @@
...
+           <item>
+            <widget class="QLabel" name="headerContributeFinancially">
+             <property name="font">
+              <font>
+               <weight>75</weight>
+               <bold>true</bold>
+              </font>
+             </property>
+             <property name="text">
+              <string>Contribute Financially</string>
```
rnestler: "Contribute Financially" sounds a bit strange to me. Maybe just "Donate"?

dbrgn: Changed.

dbrgn:
> Wouldn't it look better if the logo is at the top of the dialog? The tab widget could then be resized to the full width of the dialog.

I put it in the top center instead:

img

dbrgn:
Ah, I missed some comments above. I addressed most of them in fixup commits, except one. Will do that later.

libs/librepcb/common/dialogs/aboutdialog.cpp
```diff
@@ -0,0 +1,114 @@
...
+    mUi->tabWidget->setCurrentIndex(0);
+
+    // Get some version information
+    const Version& appVersion = qApp->getAppVersion();
+    const QString& gitVersion = qApp->getGitVersion();
+    const QString& buildDate = qApp->getBuildDate().toString("yyyy-MM-dd hh:mm:ss (t)");
+
+    // Dynamic text
+    mUi->textVersion->setText(QString("%1 (%2)<br>Build date: %3").arg(appVersion.toPrettyStr(3), gitVersion, buildDate));
+    mUi->textLinks->setText(tr("For more information, please check out <a href='%1'>librepcb.org</a><br>or our <a href='%2'>GitHub repository</a>.").arg("http://librepcb.org/", "https://github.com/LibrePCB/LibrePCB"));
```
ubruhin: Could you remove the `<br>` in this label? All other labels also don't have manual line breaks.

libs/librepcb/common/dialogs/aboutdialog.cpp
```diff
@@ -0,0 +1,114 @@
...
+    // Layout
+    mUi->tabWidget->setCurrentIndex(0);
+
+    // Get some version information
+    const Version& appVersion = qApp->getAppVersion();
+    const QString& gitVersion = qApp->getGitVersion();
+    const QString& buildDate = qApp->getBuildDate().toString("yyyy-MM-dd hh:mm:ss (t)");
+
+    // Dynamic text
+    mUi->textVersion->setText(QString("%1 (%2)<br>Build date: %3").arg(appVersion.toPrettyStr(3), gitVersion, buildDate));
```
ubruhin: Adding the prefix "Version:" would improve consistency as the build date is also prefixed with "Build date:".

dbrgn:
Fixed and rebased. I think now it's finally ready to merge (unless you see something else that should get fixed).

Oh god, I think this was probably the second-worst rebase I ever did in my life... Just moving a commit that changes three parts of a file across refactorings and style guide applications is the worst, because for every conflicting move you have to manually resolve merge conflicts twice :weary:

ubruhin:
Hm I guess you have still missed two of my comments above. One of them is even a bug.

> Oh god, I think this was probably the second-worst rebase I ever did in my life...

I would squash these commits anyway - a maximum of 3 commits should really be enough for the new about dialog ;) for example:

1. Application: Add build datetime
2. Add AboutDialog
3. Use AboutDialog in all main windows

dbrgn:
> I would squash these commits anyway

Would have been good to know that beforehand, that's why I wrote...

> Let me know if I should squash some of the commits to reduce the number

...13 days ago :smile:

libs/librepcb/common/dialogs/aboutdialog.cpp
dbrgn: Done.

dbrgn:
(So annoying that Github hides the comments automatically, without explicitly resolving the discussion :confused:)

ubruhin:
> What's the second one you mentioned?

I meant the "Version:" prefix. So as soon as the bug with the twice opened dialog is fixed, I think this is ready to merge.

dbrgn:
> I meant the "Version:" prefix.

Ugh, I fixed that already, but apparently it got lost during the rebase :confused:

dbrgn:
Alright, should be fixed & rebased.

ubruhin:
Perfect, thanks! :smiley: 

Fixes #17.

Issue #179, opened 2017-08-14
BinTray: Files can only be published within 180 days from the version publish date.


ubruhin:
Travis uploads a new AppImage after every build on `master`. But 180 days after the first upload, the file cannot be overwritten or even removed again:

```
[Bintray Upload] Publishing version 'nightly' of package 'LibrePCB-Nightly-Linux-x86_64'...
[Bintray Upload] Bintray response: 403 Forbidden. Files from version 'nightly' can't be published. Files can only be published within 180 days from the version publish date.
```

This is unusable for nightly builds!

So: Is there any workaround for this? Or is there another hosting service which is better suited? Actually I don't like BinTray anyway because it is far too "complicated" just for hosting some files. A much more simple hosting service may be better suited for our use-case.

probonopd:
Bintray is meant for released versions, not for continuous or nightly builds (they have a paid version for that). As a possible alternative, you might check out uploading to GitHub Releases. I have a little [tool](https://github.com/probonopd/uploadtool) for it.

dbrgn:
You could also just upload the files to your own cloudscale server :) As long as LibrePCB is not (yet) incredibly popular, that will be just fine.

ubruhin:
> As a possible alternative, you might check out uploading to GitHub Releases.

The disadvantage of this solution is that all watchers of the repository will receive an email every time a release gets updated. For me that's too much spam ;) Or do you know a way to avoid the spam?

> You could also just upload the files to your own cloudscale server :)

Would be possible, but I think as soon as the server is really used as API server, that's probably not the best solution because of security and performance reasons ;)

I think @probonopd needs to introduce an AppImage hosting platform :stuck_out_tongue_winking_eye:

probonopd:
I think watchers actually welcome this kind of "spam" (otherwise they wouldn't be watching)?

> I think @probonopd needs to introduce an AppImage hosting platform

Hey, I am not Canonical ;-)

ubruhin:
> I think watchers actually welcome this kind of "spam" (otherwise they wouldn't be watching)?

For "normal" releases (which are created/updated very rarely), I agree because that's an important information. But automatic builds for every commit on `master` are not really important and occur too often, so IMO that's not worth to send emails for that ;)

dbrgn:
> but I think **as soon** as the server is really used as API server

...that might take a while. Once that happens, we can think about the hosting again.

ubruhin:
[...]

Issue #142, opened 2017-01-28
GitHub Pages fails to publish doxygen output because of too large repository


ubruhin:
Travis builds the doxygen documentation for each branch and pushes it to the [LibrePCB-Doxygen](https://github.com/LibrePCB/LibrePCB-Doxygen) repository. For details have a look at the implementation in [dev/travis/update_doxygen.sh](https://github.com/LibrePCB/LibrePCB/blob/master/dev/travis/update_doxygen.sh). The HTML output is then [published](http://librepcb.org/LibrePCB-Doxygen/master/) using GitHub Pages.

But the problem is that the doxygen documentation of deleted branches stays there forever, and thus the LibrePCB-Doxygen repository grows more and more. And after reaching a specific size, GitHub Pages refuses to update the published page. This did already happen a few times. Because of this, I [changed the implementation](https://github.com/LibrePCB/LibrePCB/commit/e02b325ecc1d0d2239ca70632b45c0b521fc4baa) to always append changes to the last commit (so there are always only two commits in the LibrePCB-Doxygen repository). Now the repository size grows slower, but it still grows.

I think it should be enough to change the implementation to let it remove the generated documentation of removed branches. But maybe there would be a completely different approach which works better? In fact, we don't even need a Git repository for hosting the doxygen output, this was just the easiest solution for publishing the documentation ;)

ubruhin:
Now even the builds on Travis have failed because of a timeout when cloning the LibrePCB-Doxygen repository :scream: 

So I quickly added a workaround in 5d6076a08d014d02c2ee6b4addc9a2f004b1ee38: Only update Doxygen documentation on master

I close this issue because the problem should be solved now (although with the drawback that the documentation for other branches than `master` is no longer available)...

Issue #256, opened 2018-04-22
Create an installer for Windows


ubruhin:
We really need to provide an installer for Windows before we can publish the first official release of LibrePCB.

There are many tools available to create installers. For me the most interesting one is the [Qt Installer Framework](https://doc.qt.io/qtinstallerframework/index.html), especially because of the ability to create online installers/updaters, **including the server-side components**.

Would be great to give it a try...

PR #265, opened 2018-05-21
Refactor/cleanup CI configuration and upload artifacts to our own server

ubruhin:
Refactored CI configuration:
- Support "make install" on all platforms -> simplifies packaging also on mac and Windows
- Travis: Also create application bundle (*.dmg) for mac (untested)
- Travis: Refactor helper scripts, make them also working on Windows (in MSYS2 bash)
- AppVeyor: Use MSYS2 bash to run the same helper scripts as on Travis

In addition, I extended our server infrastructure (thanks to [cloudscale.ch](https://www.cloudscale.ch/) for sponsoring the server!). Now we have two new subdomains:
- https://download.librepcb.org/ - Nightly builds are automatically uploaded to this page, so now we have them all at the same place. No more need for BinTray (see also #179) and AppVeyor for file hosting. In future we can also host releases (and maybe more resources) on this page.
- https://doxygen.librepcb.org/ - Doxygen documentation. No more need to host them on GitHub Pages (see also #142).

Both nightly builds and doxygen documentation are now available for all branches! So now we can even download build artifacts or read the documentation from other branches than `master`.

Tasks to do after merging this:
- Delete (or disable) BinTray account https://bintray.com/librepcb
- Delete repository https://github.com/LibrePCB/LibrePCB-Doxygen
- Update installation instructions in https://github.com/LibrePCB/librepcb-doc


ubruhin:
I can't decide which identifier to use in filenames to specify the architecture :sob: 
- `x86` / `x86_64` (nice because well known)
- `x86` / `x64` (nice because short)
- `i386` / `amd64` (don't like this)
- `32bit` / `64bit` (don't like this)
- ...

sebgonzo:
I vote for `x86` / `x86_64`. It's easier to distinguish than `x86` / `x64` and reffers better to x86 architecture with 64bit extensions.

dbrgn:
Cool!

I agree with @sebgonzo.

Furthermore, a README in the root would be good. What does the `installer_packages` directory mean? What do the three directories inside `nightly_builds` mean?

ubruhin:
OK, let's go with `x86` / `x86_64`.

> Furthermore, a README in the root would be good.

Yeah I can give it a try some time. But generally download.librepcb.org is intended only for file hosting, not for things like installation instructions or something like that.

> What does the installer_packages directory mean?

I'm just experimenting with Qt's installer framework (see #256), where this directory provides packages which will be automatically fetched by the installer.

> What do the three directories inside nightly_builds mean?

They are branch names. Every branch has its own directory here.

dbrgn:
> But generally download.librepcb.org is intended only for file hosting, not for things like installation instructions or something like that.

Yes, but packagers will go to that domain and take a look at what files are available. The information needs to be for them.

PR #266, opened 2018-05-23
Zooming shortcuts for libraryeditor, boardeditor and schematiceditor

Petross404:
I used plain plus and minus for zoom in / out but what is troubling me is that designer did put "&amp;" in every menu. Why is that?

### DIFF
--- a/libs/librepcb/libraryeditor/libraryeditor.ui
+++ b/libs/librepcb/libraryeditor/libraryeditor.ui
@@ -76,12 +76,12 @@
      <x>0</x>
      <y>0</y>
      <width>611</width>
-     <height>22</height>
+     <height>32</height>
     </rect>
    </property>
    <widget class="QMenu" name="menuFile">
     <property name="title">
-     <string>File</string>
+     <string>&amp;File</string>
     </property>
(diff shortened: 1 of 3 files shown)


ubruhin:
Cool, thanks! :smiley: 

> but what is troubling me is that designer did put "&" in every menu. Why is that?

Okay, my qtcreator doesn't do that, but actually that's a nice thing because now we automatically have keyboard shortcuts for all menu entries :smiley: The "&" sign is not displayed in the GUI, but the character following the "&" gets an underscore which means that this menu item can be triggered by pressing Alt+\<character\> (e.g. Alt+S for a menu item called "&Save").

For me the PR looks good, only the commit message could be better (see [CONTRIBUTING.md](https://github.com/LibrePCB/LibrePCB/blob/master/CONTRIBUTING.md) for commit message guidelines). But don't worry, GitHub allows me to adjust the message when merging ;)

So from your side this PR is ready to merge, right?

Petross404:
Thank you. FYI I am willing to edit my commit message.

Other than that I don't have anything else to add or tweak. :wink:

PR #267, opened 2018-05-23
Provide installer for nightly builds

### LINKED ISSUE #256: Create an installer for Windows

ubruhin:
With this PR the CI now also builds an installer for Linux, Mac and Windows. The installer can be used to install and even to update LibrePCB to the latest version. This is especially useful for Windows which doesn't have a package manager.

The installers are built with the [Qt Installer Framework](https://doc.qt.io/qtinstallerframework/index.html) which is pretty cool. On Windows the installer also (optionally) registers LibrePCB as handler for *.lpp files, so project files can be opened with a double click in the explorer. For Linux and Mac I did not (yet) implement it, maybe I'll also give it a try...

The installers can currently be downloaded from here: https://download.librepcb.org/nightly_builds/installer/

Updates are hosted separately, the installer fetches them automatically from here: https://download.librepcb.org/installer_packages/

This is an important step for LibrePCB because now it's dead simple to install, update and use LibrePCB on Windows!

Fixes #256.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,20 +13,20 @@ matrix:
     - os: linux
       dist: trusty
       compiler: gcc
-      env: BUILD_DOXYGEN=true DEPLOY_APPIMAGE=false QT_BASE="trusty"
+      env: BUILD_DOXYGEN=true DEPLOY_APPIMAGE=false DEPLOY_INSTALLER=false QT_BASE="trusty"
     # Ubuntu 14.04 + Clang + Qt from PPA (https://launchpad.net/~beineri) + AppImage deployment
     - os: linux
       dist: trusty
       compiler: clang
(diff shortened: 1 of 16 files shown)


ubruhin:
Now the *.lpp mime type registration should also work on Linux. I think the installer is now ready to deploy...

Symbian9:
> The installers can currently be downloaded from here:
> https://download.librepcb.org/nightly_builds/installer/

It is incorrect link!

Here is actual link:

- https://download.librepcb.org/nightly_builds/master/

PR #268, opened 2018-05-26
Mirror support for schematic

eqvinox:
This enables flipping around symbol instances in the schematic editor.

(note: vertical flipping = horizontal flip + 180° rotate)

### DIFF
--- a/libs/librepcb/project/schematics/cmd/cmdsymbolinstanceedit.cpp
+++ b/libs/librepcb/project/schematics/cmd/cmdsymbolinstanceedit.cpp
@@ -37,7 +37,8 @@ namespace project {
 CmdSymbolInstanceEdit::CmdSymbolInstanceEdit(SI_Symbol& symbol) noexcept :
     UndoCommand(tr("Edit symbol instance")), mSymbol(symbol),
     mOldPos(symbol.getPosition()), mNewPos(mOldPos),
-    mOldRotation(symbol.getRotation()), mNewRotation(mOldRotation)
+    mOldRotation(symbol.getRotation()), mNewRotation(mOldRotation),
+    mOldMirrored(symbol.getMirrored()), mNewMirrored(mOldMirrored)
 {
 }
 
@@ -46,6 +47,7 @@ CmdSymbolInstanceEdit::~CmdSymbolInstanceEdit() noexcept
     if (!wasEverExecuted()) {
         mSymbol.setPosition(mOldPos);
         mSymbol.setRotation(mOldRotation);
+        mSymbol.setMirrored(mOldMirrored);
     }
 }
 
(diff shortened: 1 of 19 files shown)


ubruhin:
Wow, very nice work @eqvinox, thanks! :smiley: 

After a quick review, here some feedback:
1. I would rename the new S-Expr attribute from "mir" to "mirror", then it's consistent with devices (class `BI_Device`). Yes, I know that it's still not consistent with "pos" and "rot", actually I wonder if we should also rename them to "position" resp. "rotation", but that's another topic ;)
2. I don't like the change in `sexpression.h` because the thrown exceptions cause many error messages in the console output. I would move this logic to `SI_Symbol` instead. Just check if the new mirror attribute exists before accessing it. If not, load its default value. See for example the symbol attribute "lib_gate" which also implements backward compatibility that way.
3. The handling of mirroring in `CmdSymbolInstanceEdit` seems to be different from the one in `CmdDeviceInstanceEdit`. Maybe you could adjust that so both classes implement mirroring the same way? Otherwise I need to take a deeper look if mirroring of symbols really works the same way as of devices (it's very easy to make mistakes here)...
4. The placement of the mirror checkbox in `SymbolInstancePropertiesDialog` is not consistent with the same checkbox in `DeviceInstancePropertiesDialog`. Maybe you could adjust this to make it consistent?
5. It would be helpful to have a keyboard shortcut for mirroring. You could just set e.g. "M" as keyboard shortcut for the corresponding QAction in the designer.

Btw, :+1: for applying our coding guidelines very well :)

eqvinox:
Pushed updated version (same PR/here), adressed everything except 3.

The reason the mirroring handling is different is that I haven't found the time to implement vertical mirroring yet. I only occasionally have the time to hack on this... I will probably get to it at some point, your choice whether you want to take it as it is or wait ;)

ubruhin:
> The reason the mirroring handling is different is that I haven't found the time to implement vertical mirroring yet.

Hm, that's actually the thing you could copy&paste from `CmdDeviceInstanceEdit`. It "emulates" vertical mirroring by horizontal mirroring + 180° rotation too, so no need to change something in `SymbolInstance`.

eqvinox:
Yes it's easy, but unfortunately that doesn't mean the work does itself so I still need to find some time to get to it :D

ubruhin:
I see :) So if it's OK for you, I'll wait a bit with merging this, maybe you can continue working on this some time...

Btw. there  seems to be a small bug with mirroring of text items and netlabels. For example this:

grafik

gets mirrored to this:

grafik

Note the VSS, GND and V_BATTERY texts.

ubruhin:
@eqvinox Do you see a chance to at least fix the bug I mentioned above?

Unfortunately rebasing this branch to master will cause some merge conflicts, but I could do the rebase for you if you like ;)

ubruhin:
I rebased your branch and added a commit to implement symbol mirroring exactly the same way as footprint mirroring (point 3 of https://github.com/LibrePCB/LibrePCB/pull/268#issuecomment-392324737) and to (hopefully) fix the wrong text position I mentioned in https://github.com/LibrePCB/LibrePCB/pull/268#issuecomment-395874249. Netlabel mirroring is still not working, but I will open a separate issue for that, so it can be fixed later...

If CI succeeds, I will merge this so it will be included in our first release :)

Thanks for contributing @eqvinox! :smiley:

PR #269, opened 2018-05-27
Add Russian translation

ktrace:
Add translation.


dbrgn:
Very nice :slightly_smiling_face: 

@ubruhin we should set up a community translation tool soon. Something like Transifex.

ubruhin:
Thanks @ktrace!

Unfortunately we don't have a workflow for translations yet, and also the current translation files in the repository are extremely outdated. We started working on this now, see #270.

I will wait with merging this until we know how to proceed exactly. Maybe we stop checking in translation files into this repository, then I'll import your translations directly into Transifex instead...

dbrgn:
@ktrace you're of course welcome to contribute translations on Transifex!

Issue #270, opened 2018-05-27
Set up community translation project


dbrgn:
We should have a simple way for the community to contribute and fix translations.

Transifex is a great tool for this, and free for open source: https://www.transifex.com/

TODO:

- [x] Set up Transifex for a test run
- [ ] Document how to contribute translations

ubruhin:
Yep, good idea :+1: 

I created a project here: https://www.transifex.com/librepcb/librepcb-application/.

dbrgn:
Nice!

We should also create a translation guide. One part of that is "how to translate", the other one should be how to consistently translate technical terms.

We should for example create a table for all languages where we can specify how to translate words like "component" or "device". Otherwise we'll end up with inconsistent and confusing translations.

ubruhin:
@dbrgn I guess translation works pretty good so far, so can we close this issue?

dbrgn:
Yes!

PR #271, opened 2018-05-27
Add neutral translation file

dbrgn:
This can be uploaded to translation tools like Transifex.

Refs #270. Let's discuss this at Coredump.


ubruhin:
I'm not sure if this is the right way because we would need to update that file regularly. I'll try to set up CI to automatically generate that file with every commit, and automatically upload it to Transifex. Then we would probably never have to care about translation files in this repository...

dbrgn:
I agree with the upload-through-CI approach. Let's close this.

PR #272, opened 2018-05-29
Fix catch by value of polymorphic exceptions

rnestler:
Fixes a lot of `warning: catching polymorphic type ‘class
librepcb::Exception’ by value [-Wcatch-value=]` warnings.

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/fsm/bes_drawplane.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/fsm/bes_drawplane.cpp
@@ -275,7 +275,7 @@ bool BES_DrawPlane::start(Board& board, const Point& pos) noexcept
         mLastVertexPos = pos;
         makeSelectedLayerVisible();
         return true;
-    } catch (Exception e) {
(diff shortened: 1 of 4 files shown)


ubruhin:
Looks good :+1:

It seems that this warning was never issued on Travis-CI (otherwise CI would fail because of `-Werror`). @rnestler could you maybe also check if we can enable this warning on CI?

rnestler:
> It seems that this warning was never issued on Travis-CI (otherwise CI would fail because of -Werror). @rnestler could you maybe also check if we can enable this warning on CI?

I suspect it's due to the much older compilers being used for CI. We could just upgrade to a newer compiler though.

ubruhin:
> We could just upgrade to a newer compiler though.

Would you take a look at that? If yes, I'll wait with merging this PR. If not, I'll merge it now.

rnestler:
> Would you take a look at that? If yes, I'll wait with merging this PR. If not, I'll merge it now.

You can merge it now, I can still take a look at it later.

ubruhin:
Thanks!

PR #284, opened 2018-07-04
Simplify build configuration & change build output path

ubruhin:
#### 1) Stop copying "share" directory to build output

The copy action of the "share" directory had several drawbacks:

- Increased build time
- Unnecessary complex build process
- Workaround on macOS needed to fix wrong relative path to executable

Now the application determines at startup whether it's running from inside the build output directory (i.e. it's a developer build) or from outside (i.e. it's a packaged/installed build). If it's a developer build, the "share" directory of the repository root is used. If not, the directory "../share" relative to the executable is used.

#### 2) Improve detection of runtime resources availability

Because the hidden file ".librepcb-resources" is ignored in the deployment, a warning is raised on official builds saying that the runtime resources are missing (which was wrong).

- Replace ".librepcb-resources" with "README.md" to fix deployment
- Provide more debug output in case of missing runtime resources

#### 3) Change build output path to "./output"

Previously the build output path was "./generated/<unix|mac|windows>/", depending on the operating system. Now it is simplified to "./output", (independent of the operating system) because it doesn't make sense to use different output paths for different systems.

In addition, the variable "GENERATED_DIR" in all *.pro files is no longer needed.

## ATTENTION

This changes the location of generated files! So maybe this breaks some packaging scripts. It's best to use `make install` for copying the generated files, because this will always work, independent of the location of the build output.


rnestler:
> Previously the build output path was "./generated/<unix|mac|windows>/", depending on the operating system. Now it is simplified to "./output", (independent of the operating system) because it doesn't make sense to use different output paths for different systems.

Actually if you are cross compiling it *does* matter. But we have dedicated build machines for every OS so we don't care.

ubruhin:
> Actually if you are cross compiling it does matter. But we have dedicated build machines for every OS so we don't care.

You can anyway run `qmake && make` in an arbitrary directory, then the whole build output goes there. So when building for different systems on the same machine, one can just run `qmake && make` in different directories.

The path I changed to `./output` is only used *inside* this build directory.

Issue #275, opened 2018-06-08
Library Editor: Sort table widgets by name


ubruhin:
Sort rows of following table widgets by name:

- `ComponentSignalListEditorWidget`
- `FootprintListEditorWidget`
- `PackagePadListEditorWidget`

See also #210, #211 and https://github.com/LibrePCB/librepcb-rfcs/issues/13#issuecomment-395731545.

Note: It's not as easy as enabling auto-sort of the table widgets. It's quite tricky to implement it properly...

![deepinscreenshot_select-area_20180608125454](https://user-images.githubusercontent.com/5735720/41155083-d60febd0-6b1c-11e8-8797-c79337e5ed94.png)

PR #285, opened 2018-07-04
Library Editor pin name sort added (when possible) #275, #210, #211

lukabosnjakovic:
### Disclaimer first :D
I'm programmer but I'm not professional C++ programmer so I may have some serious errors. This is my first open-source contribute so I don't know how to fully contribute and I don't know best practices. As example I didn't make tests, tests in this way are new to me so I decided not to do them because I really don't know what to type :D 
Also I read CONTRIBUTING.md and I tried to stick to it. You will judge did I do good job or not :D

## Technique
Instead sorting table (QTableWidget) i decided to sort object that will handle data for displaying in table. The result is that you get sorted table as you want :). Reason for this is I didn't succeed to make usable class that is extended from QTableWidgetItem and has functional function for sorting. Problem is data in table. Those are strings and comparison works different than comparison to integers. So if you look you will see that embedded sort is working ok but for strings not integers (when there are they)

## About my changes
I added functionality for sorting pin names in Library Editor wherever I could. I think it is ok, but I didn't succeed to add sort in Component Signal Variant (LibEditor > Components > (open component) > (double click on row in Symbol Variants section)). Reason is that this table is filled with Pin-Signal-Map object that has UUIDs for Symbol Pins and I couldn't sort them because I lost in the code :D maybe one day I do it but for now I think I must take focus on some other problems. I didn't always used my function for sorting, for simplicity I used in few times QTableWidget's function for sorting if it was working ok. Also I removed vertical header form tables and replace it with normal cells/columns. This is because that way it is easier to manipulate with them.

As I'm not pro C++ programmer I spent too much time on understanding how classes are bounded so this is why I'm so slow :D 
 
I think I cover issues #275, #210, #211 with this pull req.

Your comments are very welcome and I expect them :)

### DIFF
--- a/libs/librepcb/common/fileio/serializableobjectlist.h
+++ b/libs/librepcb/common/fileio/serializableobjectlist.h
@@ -284,20 +284,33 @@ class SerializableObjectList : public SerializableObject
         }
 
         // Convenience Methods
-        SerializableObjectList<T, P> sortedByUuid() const noexcept {
+        SerializableObjectList<T, P> sortedByUuid(bool asc) const noexcept {
             SerializableObjectList<T, P> copiedList;
             copiedList.mObjects = mObjects; // copy only the pointers, not the objects!
             qSort(copiedList.mObjects.begin(), copiedList.mObjects.end(),
-                  [](const std::shared_ptr<T>& ptr1, const std::shared_ptr<T>& ptr2)
-                    {return ptr1->getUuid() < ptr2->getUuid();});
+                  [&asc](const std::shared_ptr<T>& ptr1, const std::shared_ptr<T>& ptr2)
+                    {
+                if(asc){
+                    return ptr1->getUuid() < ptr2->getUuid();
+                } else {
+                    return ptr1->getUuid() > ptr2->getUuid();
+                }});
+            return copiedList;
+        }
+        SerializableObjectList<T, P> sortedByName(bool asc) noexcept {
+            //qSort(copiedList.mObjects.begin(), copiedList.mObjects.end(),
+            qSort(mObjects.begin(), mObjects.end(),
+                  [this, &asc](const std::shared_ptr<T>& ptr1, const std::shared_ptr<T>& ptr2)
+                    {
+                        return myComparison(ptr1->getName(), ptr2->getName(), asc);
+            });
+            SerializableObjectList<T, P> copiedList;
+            copiedList.mObjects = mObjects; // copy only the pointers, not the objects!
             return copiedList;
         }
         SerializableObjectList<T, P> sortedByName() const noexcept {
             SerializableObjectList<T, P> copiedList;
             copiedList.mObjects = mObjects; // copy only the pointers, not the objects!
-            qSort(copiedList.mObjects.begin(), copiedList.mObjects.end(),
-                  [](const std::shared_ptr<T>& ptr1, const std::shared_ptr<T>& ptr2)
-                    {return ptr1->getName() < ptr2->getName();});
             return copiedList;
         }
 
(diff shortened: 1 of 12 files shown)


rnestler:
Awesome! Thanks for the contribution. I did only briefly look at the code and can't really judge on the contents, @ubruhin is probably better suited for that :wink: 

I did notice though that there is some [commented](https://github.com/LibrePCB/LibrePCB/pull/285/files#diff-7590900e775f76db5cc90d10a05a69c2R301) out [code](https://github.com/LibrePCB/LibrePCB/pull/285/files#diff-7590900e775f76db5cc90d10a05a69c2R391). This is perfectly OK for a first draft of a pull-request, but should be cleaned later on before merging.

Also you should rebase (https://git-scm.com/book/en/v2/Git-Branching-Rebasing) your merge request onto the latest master. There are merge conflicts so the CI checks can't run.

ubruhin:
Hi @lukabosnjakovic, thank you very much for the contribution!

> This is my first open-source contribute so I don't know how to fully contribute and I don't know best practices. As example I didn't make tests, tests in this way are new to me so I decided not to do them because I really don't know what to type :D

Don't worry, tests are nice to have, but not required to get pull requests merged :)

> Instead sorting table (QTableWidget) i decided to sort object that will handle data for displaying in table. The result is that you get sorted table as you want :)

Unfortunately it leads to other issues when modifying the underlying data (changing their order as done by `sortedByName()`) instead of just sort the representation in the GUI. Modifying data is not allowed without using undo commands (see documentation [here](https://doxygen.librepcb.org/master/d5/d75/doc_project.html#doc_project_undostack)) because this leads to serious errors when using undo/redo. You can verify this for example with following procedure:

1. Create a new package in the library editor (without any pads)
2. Add three new pads in the pad list (named "1", "2" and "3")
3. Remove pad "1" from the list
4. Click on the pad list header to reverse sorting
5. Click "Undo"
6. Click "Redo" --> This makes the application crashing because of a failed assert!

Using undo commands for sorting would fix this issue, but it would introduce other problems (data is considered as modified because of sorting). So the only solution would be to do sorting only in the representation in the GUI. Would you maybe give another try to do it that way?

lukabosnjakovic:
Ok, I didn't think about that I admit :D 

Than I'll start again with GUI approach, I think I'll be faster now when I'm little in the code. Can we cancel this poll request or how to proceed ?

ubruhin:
> Than I'll start again with GUI approach, I think I'll be faster now when I'm little in the code.

That would be great!

> Can we cancel this poll request or how to proceed ?

You have two options:

1. Either close this pull request and open a new one once you have the new implementation ready
2. Or continue working on your branch `LibraryEditorPinNamingSort` by adding new commits, squashing and rebasing. This pull request is then updated automatically every time you push to that branch.

For me it doesn't matter which option you choose. If you want to start the new implementation from scratch anyway, and aren't familiar with Git features like rebase, squash and force push, probably the first option is simpler.

lukabosnjakovic:
I'll try second option, because it is harder :D I already rebase my branch as @rnestler propose to me, so I'll just continue to work. I used git with GUI on Windows so now I learn how to use it from cmd as I am in Linux now and GUI is not good as it is on Windows.

ubruhin:
> I'll try second option, because it is harder :D

Good choice :smiley: 

Just let us know if you need some help with the new implementation or with Git ;)

rnestler:
> I used git with GUI on Windows so now I learn how to use it from cmd as I am in Linux now and GUI is not good as it is on Windows.

There is a nice GUI for git which is also available for Linux: https://www.syntevo.com/smartgit/. It is not open source, but free for non-commercial use.
I prefer to work with the CLI though :)

ubruhin:
It seems that the current implementation of all the table widgets in the library editor has several drawbacks (missing sort functionality is just one of them), thus I started to completely refactor them. Since this PR is pretty old and no longer needed after the refactoring, I'll close it.

Sorry for the inconvenience ;) Thanks anyway for your time @lukabosnjakovic  :slightly_smiling_face:

PR #287, opened 2018-07-10
Add initial setup for automated functional tests

ubruhin:
This adds an initial setup to perform automated functional tests ("GUI tests") using [funq](https://github.com/parkouss/funq) and [pytest](https://docs.pytest.org/en/latest/index.html). I think such tests are very useful to keep the quality and stability of LibrePCB as high as possible. So it would be great to add more and more tests in future.

peek 2018-07-10 22-42

### Other Changes

There were some small changes required in the source code of LibrePCB to make `funq` working properly:

#### Use INI for application settings on all platforms

Tests require to control local application settings (`~/.config/LibrePCB/` on Linux). On Windows, Qt uses the Registry for these settings, but this is a mess for testing. I think the Registry is bullshit anyway, so let's use INI instead. Now settings have the same format on all platforms.

#### Slightly delay opening projects passed as arguments

This is required to make `libFunq` initializing as early as possible (it injects a TCP server into the application, which then communicates with the tests written in Python).

#### Wrap `QFileDialog` to allow disabling native file dialogs

Add new common class `FileDialog` as a wrapper for `QFileDialog`. It allows to disable native file dialogs with the environment variable `LIBREPCB_DISABLE_NATIVE_DIALOGS`. This is used for automated functional tests because native dialogs cannot be tested.

#### Move unit tests from ./tests to ./tests/unittests

To clearly distinguish between unit tests and functional tests.

#### Rename tests executable from `tests` to `librepcb-unittests`

Just to be more expressive.

### DIFF
--- a/apps/librepcb/controlpanel/controlpanel.cpp
+++ b/apps/librepcb/controlpanel/controlpanel.cpp
@@ -22,7 +22,6 @@
  ****************************************************************************************/
 #include <QtCore>
 #include <QtWidgets>
-#include <QFileDialog>
 #include "controlpanel.h"
 #include "ui_controlpanel.h"
 #include <librepcb/workspace/workspace.h>
@@ -40,6 +39,7 @@
 #include <librepcb/common/application.h>
 #include <librepcb/common/dialogs/aboutdialog.h>
 #include <librepcb/common/fileio/fileutils.h>
+#include <librepcb/common/dialogs/filedialog.h>
 #include "../markdown/markdownconverter.h"
 #include "projectlibraryupdater/projectlibraryupdater.h"
 
@@ -436,8 +436,8 @@ void ControlPanel::on_actionOpen_Project_triggered()
     QString lastOpenedFile = settings.value("controlpanel/last_open_project",
                              mWorkspace.getPath().toStr()).toString();
 
-    FilePath filepath(QFileDialog::getOpenFileName(this, tr("Open Project"), lastOpenedFile,
-                                    tr("LibrePCB project files (%1)").arg("*.lpp")));
+    FilePath filepath(FileDialog::getOpenFileName(this, tr("Open Project"), lastOpenedFile,
+                      tr("LibrePCB project files (%1)").arg("*.lpp")));
 
     if (!filepath.isValid())
         return;
--- a/apps/librepcb/firstrunwizard/firstrunwizardpage_workspacepath.cpp
+++ b/apps/librepcb/firstrunwizard/firstrunwizardpage_workspacepath.cpp
@@ -20,10 +20,10 @@
 /*****************************************************************************************
  *  Includes
  ****************************************************************************************/
-#include <QFileDialog>
 #include "firstrunwizardpage_workspacepath.h"
 #include "ui_firstrunwizardpage_workspacepath.h"
 #include <librepcb/common/fileio/filepath.h>
+#include <librepcb/common/dialogs/filedialog.h>
 #include <librepcb/workspace/workspace.h>
 
 /*****************************************************************************************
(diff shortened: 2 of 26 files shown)


apps/librepcb/controlpanel/controlpanel.cpp
```diff
@@ -122,13 +122,12 @@ ControlPanel::ControlPanel(Workspace& workspace) :
...
-        FilePath filepath(arg);
-        if ((filepath.isExistingFile()) && (filepath.getSuffix() == "lpp"))
-            openProject(filepath);
-    }
+    // slightly delay opening projects to not block processing of other important events
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 4, 0))
+    QTimer::singleShot(10, this, &ControlPanel::openProjectsPassedByCommandLine);
+#else
+    QTimer::singleShot(10, this, SLOT(openProjectsPassedByCommandLine()));
+#endif
```
rnestler: Do we really want to introduce technical dept like that before the first release? Wouldn't it be easier to just require Qt >= 5.4?
People working on ancient, soon to be unsupported distributions can easily install a newer Qt version from other sources.

apps/librepcb/main.cpp
```diff
@@ -82,6 +82,24 @@ int main(int argc, char* argv[])
     // Initialize all 3rd party libraries
     init3rdPartyLibs();
 
+    // Migrate from old setting format to INI
+    // backward compatibility - remove this some time!
```
rnestler: Maybe mark it with a `TODO`. Most editors and IDE's will highlight those.

apps/librepcb/main.cpp
```diff
@@ -82,6 +82,24 @@ int main(int argc, char* argv[])
...
+    if (!FilePath(newSettings.fileName()).isExistingFile()) {
+        qInfo() << "Migrating old settings to" << newSettings.fileName();
+        foreach (const QString& key, oldSettings.allKeys()) {
+            newSettings.setValue(key, oldSettings.value(key));
+        }
+    }
+
+    // Make sure the INI format is used for settings on all platforms because:
+    // - Consistent storage format on all platforms
+    // - Windows Registry is bullshit
```
rnestler: Maybe use less profane language :laughing: Something like: .ini files are easily version controlled compared to registry settings.

0xB767B: +1 for bullshit :-D

apps/librepcb/controlpanel/controlpanel.cpp
ubruhin: Discussing this topic (once again...) requires more time than I needed for the implementation ;) I stick with Qt5.2 support until Ubuntu 14.04 support has ended. It's *really* not a big deal to keep Qt5.2 support!

apps/librepcb/main.cpp
ubruhin: Too late... I already added exactly the same comment at several other locations (unfortunately without TODO), now I use the same comment just for consistency. I will remove them anyway after the first release...

apps/librepcb/main.cpp
ubruhin: I could add a ";)" to make clear that it is meant to amuse the reader :)

apps/librepcb/controlpanel/controlpanel.cpp
rnestler: So I'll keep mentioning it until April 2019 :wink: . And then I'll start complaining about unnecessary backwards compatibility with Qt < 5.9 :stuck_out_tongue_closed_eyes:

ubruhin:
Extracted independent changes of this to separate PRs: #294, #295

I will wait with merging this PR because I'm still evaluating another GUI testing tool.

ubruhin:
OK I decided to stay with [funq](https://github.com/parkouss/funq) as testing framework, so I updated this branch to make it ready for merging.

If someone is looking for a small side project: There are many things at [funq](https://github.com/parkouss/funq) which could be improved ;) I already contributed some improvements to that project to get it initially working for LibrePCB...

PR #296, opened 2018-07-30
Refactor and improve handling of invalid objects

### LINKED ISSUE #157: Decide how to deal with invalid objects

### LINKED ISSUE #288: Improve error messages for invalid library elements
##### LibrePCB VERSION

Commit d7d9096ec

##### OS / ENVIRONMENT

Arch Linux

##### SUMMARY

Please tell me what I did wrong :slightly_smiling_face: 

screenshot

ubruhin:
- Use `optional<T>` from https://github.com/TartanLlama/optional to represent nullable objects:
  - Use `optional<Uuid>` instead of nullable `Uuid`
    - Small change in the file format: Null UUIDs are now stored as `none` instead of `null`
  - Use `optional<Version>` instead of nullable `Version`
- Use `constrained_type<T>` from https://github.com/foonathan/type_safe to create constrained types:
  - Use constrained `Length` types `UnsignedLength` and `PositiveLength`
  - Use constrained `Ratio` type `UnsignedRatio`
  - Use constrained `QString` type `AttributeKey`
  - Use constrained `QString` type `GraphicsLayerName`
  - Use constrained `QString` type `ElementName`
  - Use constrained `QString` type `ComponentPrefix`
  - Use constrained `QString` type `ComponentSymbolVariantItemSuffix`
  - Use constrained `QString` type `CircuitIdentifier`
- `BoardDesignRules`: Disallow invalid design rules
- Generally relax some requirements on elements validity

The optional and constrained types are very useful to avoid invalid objects in many situations. Many invalid values are now rejected right in the user input forms and thus cannot propagate into the lower level classes.

Many elements have now pretty strong restrictions regarding the allowed characters. Especially symbol pins, package pads, component signals, component names and net signals must now follow the regex `^[-a-zA-Z0-9_+/!?@#$]{1,32}$`, otherwise they are rejected. Maybe we can relax these restrictions some time, but better start with strong restrictions and relax them later, the other way around would be much more cumbersome.

Fixes #157, Fixes #288


dbrgn:
> Use optional<T> from https://github.com/TartanLlama/optional

Woah, that looks awesome :) Especially the map / and_then / etc stuff.

ubruhin:
> Woah, that looks awesome :) Especially the map / and_then / etc stuff.

Yep ;)

I expect nobody wants to review this PR? :grin:

libs/librepcb/common/application.cpp
```diff
@@ -37,35 +37,32 @@ namespace librepcb {
...
-               "Please correct this in the file 'librepcbcommon.pro'.");
+    // check "git describe" version
+    if (mGitVersion.isEmpty()) {
+        qWarning() << "Git revision not compiled into the executable!";
+    }
+
+    // check file format version
+    if (!mFileFormatVersion.isPrefixOf(mAppVersion)) {
+        qFatal("The file format version is not a prefix of the application version!");
+    }
```
dbrgn: Why do you pipe a string into qWarning, but call qFatal with a parameter?

ubruhin: Because qFatal does not provide the << operator.

PR #298, opened 2018-08-01
Library Editor: Make "Close Library Editor" menu item working

### LINKED ISSUE #297: Library Editor: "Close Library Editor" Menu Item has no effect
##### LibrePCB VERSION

7acdba9af0cdf3bf03d957a04dffda7966bfb33d

##### OS / ENVIRONMENT

Arch Linux

##### SUMMARY

Clicking the "Close Library Editor" menu item in Library Editor has no effect.

##### STEPS TO REPRODUCE

Click menu item.

##### EXPECTED RESULTS

Library Editor should close.

##### ACTUAL RESULTS

Library Editor does not close.

0xB767B:
Fixes #297

### DIFF
--- a/libs/librepcb/libraryeditor/libraryeditor.cpp
+++ b/libs/librepcb/libraryeditor/libraryeditor.cpp
@@ -59,6 +59,8 @@ LibraryEditor::LibraryEditor(workspace::Workspace& ws, QSharedPointer<Library> l
     mCurrentEditorWidget(nullptr), mLock(lib->getFilePath())
 {
     mUi->setupUi(this);
+    connect(mUi->actionClose, &QAction::triggered,
+            this, &LibraryEditor::close);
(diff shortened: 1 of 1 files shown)


ubruhin:
Thx :)

The `closeTriggered()` function is not needed, you can connect the signal directly to the slot `LibraryEditor::close` (which is inherited from `QWidget`).

And could you please rebase the branch to master?

0xB767B:
I thought it would make sense to call ```closeAndDestroy``` so that saving is handled?

Wasn't the request already rebased?

ubruhin:
> I thought it would make sense to call closeAndDestroy so that saving is handled?

There's an event handler for the close event which should call closeAndDestroy anyway, so no need to do it manually (clicking on the "X" in the title bar also just calls `close()`).

ubruhin:
Thanks! :+1:

PR #299, opened 2018-08-04
Add command-line tool "librepcb-cli"

ubruhin:
This adds a first version of a `librepcb-cli` application which is intended to automate some frequently used tasks. It's especially useful for Continuous Integration of LibrePCB projects (e.g. with Travis-CI).

Currently it supports to load a project and...
- Report ERC messages (report failure if there are non-approved messages)
- Export schematics as PDF
- Export boards as Gerber/Excellon

For usage instructions see documentation here: https://docs.librepcb.org/_branches/add-cli/#cli

### DIFF
--- a/apps/librepcb-cli/commandlineinterface.cpp
+++ b/apps/librepcb-cli/commandlineinterface.cpp
@@ -1,0 +1,338 @@
+/*
+ * LibrePCB - Professional EDA for everyone!
+ * Copyright (C) 2013 LibrePCB Developers, see AUTHORS.md for contributors.
+ * http://librepcb.org/
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*****************************************************************************************
+ *  Includes
+ ****************************************************************************************/
+#include <QtCore>
+#include "commandlineinterface.h"
+#include <librepcb/common/application.h>
+#include <librepcb/common/debug.h>
+#include <librepcb/common/attributes/attributesubstitutor.h>
+#include <librepcb/project/project.h>
+#include <librepcb/project/boards/board.h>
+#include <librepcb/project/boards/boardgerberexport.h>
+#include <librepcb/project/erc/ercmsg.h>
+#include <librepcb/project/erc/ercmsglist.h>
+
+/*****************************************************************************************
+ *  Namespace
+ ****************************************************************************************/
+namespace librepcb {
+namespace cli {
+
+using namespace librepcb::project;
+
+/*****************************************************************************************
+ *  Constructors / Destructor
+ ****************************************************************************************/
+
+CommandLineInterface::CommandLineInterface(const Application& app) noexcept :
+    mApp(app)
+{
+}
+
+/*****************************************************************************************
+ *  General Methods
+ ****************************************************************************************/
+
+int CommandLineInterface::execute() noexcept
+{
+    QMap<QString, QPair<QString, QString>> commands = {
+        {
+            "open-project", {
+                tr("Open a project to execute project-related tasks."),
+                tr("open-project [command_options]")
+            }
+        },
+    };
+
+    // Add global options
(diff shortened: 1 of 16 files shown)


apps/librepcb-cli/commandlineinterface.cpp
```diff
@@ -0,0 +1,334 @@
...
+    parser.addPositionalArgument("command", tr("The command to execute."));
+
+    // Define options for "open-project"
+    QCommandLineOption checkErcOption(
+        "check-erc",
+        tr("Check ERC messages and add message count to exit code (to report failure if "
+           "there are non-approved messages)."));
+    QCommandLineOption exportSchematicsPdfOption(
+        "export-schematics-pdf",
+        tr("Export schematics as PDF to given file. Existing file will be overwritten."),
```
dbrgn: -> existing file**s**

apps/librepcb-cli/commandlineinterface.cpp
```diff
@@ -0,0 +1,334 @@
...
+    s << str;
+    for (int i = 0; i < newlines; ++i) {
+        s << endl;
+    }
+}
+
+void CommandLineInterface::addToExitCode(qint16& exitCode, int count) noexcept
+{
+    exitCode = qMin(static_cast<int>(exitCode) + count, INT16_MAX);
+}
```
dbrgn: I'm not sure if using the exit code to convey data is a good idea. What if we would use the exit code like a bitmask? The LSB bit could indicate ERC errors, for example. Another bit could indicate files-overwritten. The error count could be printed to stderr in a well-defined format instead. That would give us more flexibility regarding the meaning of the exit code.

apps/librepcb-cli/commandlineinterface.h
```diff
@@ -0,0 +1,95 @@
...
+ * @brief The CommandLineInterface class
+ */
+class CommandLineInterface final
+{
+        Q_DECLARE_TR_FUNCTIONS(CommandLineInterface);
+
+    public:
+
+        enum ExitCode {
+            FATAL_ERROR             = -1000,
```
dbrgn: TIL that the exit code is a signed integer, not a signed byte.

dbrgn:
In theory the "create project" dialog could ask to generate CI config files for Travis/CircleCI/Gitlab CI :slightly_smiling_face: However, that would require some kind of opt-in (since you don't want to pollute projects). Same thing would apply to the .gitignore file, which some people might not want. What do you think?

apps/librepcb-cli/commandlineinterface.cpp
ubruhin: Hmm yeah now I also read some advices to not use exit codes other than 0 (SUCCESS) and 1 (FAILURE). Or at least only 0-127 for our own error codes. See for example https://stackoverflow.com/questions/18248511/program-exit-status-conventions

I think I'll change it to only return 0 or 1, the error reason is printed to stderr anyway...

ubruhin:
> In theory the "create project" dialog could ask to generate CI config files for Travis/CircleCI/Gitlab CI slightly_smiling_face However, that would require some kind of opt-in (since you don't want to pollute projects). 

Yes maybe a dropdown to choose the used CI service would be nice. But I would just wait a bit with implementing this until the CLI is more mature...

In addition to changing the exit code, I think I'll also change these two things:
- Rename `--export-schematics-pdf=<file>` to `--export-schematics=<file>`. The file type is then automatically determined by the file ending. This allows to support many different formats (PDF, PNG, SVG, ...) with a single command line option (which can be given multiple times). Maybe it would some time make sense to add (optional) parameters for the export, for example something like `--export-schematics="schematics_{{PAGE}}.png,width=1024px,height=768px,background=transparent"`.
- Rename `--check-erc` to `--erc` (the "c" in "erc" stands for "check" anyway).

apps/librepcb-cli/commandlineinterface.cpp
dbrgn: I like being able to determine the type of error from the status code. But I'd try to stick within the range -128 to 127.

ubruhin: What would be the use-case of that? If you want to know if the ERC failed or the Gerber export failed, just perform these tasks as two separate commands instead of one combined command:

    librepcb-cli open-project --check-erc MyProject.lpp
    librepcb-cli open-project --export-gerber MyProject.lpp

dbrgn: I don't have a concrete use case, just in general I think it's nice if the exit code says something about the nature of the failure :)

apps/librepcb-cli/commandlineinterface.cpp
ubruhin: Done

ubruhin:
Updated the branch as described above. In addition, I renamed `--export-gerber` to `--export-pcb-fabrication-data` (Gerber was not completely correct, it also generates Excellon files).

Also updated the documentation accordingly: https://docs.librepcb.org/_branches/add-cli/#cli

PR #300, opened 2018-08-14
Mark some Eagle importer strings as non-translatable

dbrgn:


dbrgn:
Actually I found some more strings. I'll rename the branch and will create a new PR.

PR #301, opened 2018-08-14
Mark many strings as non-translatable

dbrgn:
When creating a new dialog, please make sure to remove the "translatable" checkbox when necessary.

### DIFF
--- a/apps/librepcb/firstrunwizard/firstrunwizardpage_workspacepath.ui
+++ b/apps/librepcb/firstrunwizard/firstrunwizardpage_workspacepath.ui
@@ -14,7 +14,7 @@
    <string>WizardPage</string>
   </property>
   <property name="title">
-   <string>Select Workspace Path</string>
+   <string>Select Wor&amp;kspace Path</string>
   </property>
   <property name="subTitle">
    <string>Please select a directory to open or create a LibrePCB workspace.</string>
@@ -73,7 +73,7 @@ Workspaces are platform independent, so they can be used across different operat
      <item>
       <widget class="QToolButton" name="btnCreateWsBrowse">
        <property name="text">
-        <string>...</string>
+        <string notr="true">...</string>
        </property>
       </widget>
      </item>
@@ -89,7 +89,7 @@ Workspaces are platform independent, so they can be used across different operat
    <item>
     <widget class="QRadioButton" name="rbtnOpenWs">
      <property name="text">
-      <string>Open an existing workspace</string>
+      <string>Open an e&amp;xisting workspace</string>
      </property>
     </widget>
    </item>
@@ -121,7 +121,7 @@ Workspaces are platform independent, so they can be used across different operat
         <bool>false</bool>
        </property>
        <property name="text">
-        <string>...</string>
+        <string notr="true">...</string>
        </property>
       </widget>
      </item>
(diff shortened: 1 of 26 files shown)


dbrgn:
Any idea why the tests fail?

ubruhin:
```
In file included from ../../../../libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.cpp:24:0:
/home/travis/build/LibrePCB/LibrePCB/build/libs/librepcb/libraryeditor/ui_newelementwizardpage_entermetadata.h: In member function ‘void librepcb::library::editor::Ui_NewElementWizardPage_EnterMetadata::retranslateUi(QWizardPage*)’:
/home/travis/build/LibrePCB/LibrePCB/build/libs/librepcb/libraryeditor/ui_newelementwizardpage_entermetadata.h:156:25: error: ‘class QPlainTextEdit’ has no member named ‘setPlaceholderText’
         edtDescription->setPlaceholderText(QString());
                         ^
```

Probably this was introduced with [this change](https://github.com/LibrePCB/LibrePCB/pull/301/files#diff-f052b3407e7162f97311460e2d480090R77)(?).

dbrgn:
> Are these all strings which shouldn't be translated, or just a "random" selection of them?

All that I found in the dialogs. I went through all dialogs I think (unless I missed some).

> Probably this was introduced with this change(?).

Oh, whoops, what does that change do? Must have been something done by the designer. I'll revert that line.

apps/librepcb/firstrunwizard/firstrunwizardpage_workspacepath.ui
```diff
@@ -14,7 +14,7 @@
    <string>WizardPage</string>
   </property>
   <property name="title">
-   <string>Select Workspace Path</string>
+   <string>Select Wor&amp;kspace Path</string>
```
dbrgn: I just noticed these changes now. Do we want them?

libs/librepcb/libraryeditor/dev/deviceeditorwidget.ui
```diff
@@ -426,6 +426,7 @@
  </customwidgets>
  <resources>
   <include location="../../../../img/images.qrc"/>
+  <include location="../../../../img/images.qrc"/>
```
dbrgn: This one is strange as well.

apps/librepcb/firstrunwizard/firstrunwizardpage_workspacepath.ui
ubruhin: Yes you can keep them, these are the automatically created keyboard shortcuts (ALT+k in this case). Although I wonder why this is created for the title of a wizard page... ;)

libs/librepcb/projecteditor/newprojectwizard/newprojectwizardpage_metadata.ui
```diff
@@ -213,6 +213,7 @@
  </widget>
  <resources>
   <include location="../../../../img/images.qrc"/>
+  <include location="../../../../img/images.qrc"/>
```
ubruhin: same here...

ubruhin:
Thanks @dbrgn :+1:

PR #309, opened 2018-08-20
Add instructions about updating the repository

rnestler:
A lot of times when pulling, compilation fails afterwards because the
submodules aren't up to date.


ubruhin:
Thanks :smiley:

PR #319, opened 2018-08-27
Fix GUI issues on Windows

### LINKED ISSUE #304: Library Editor: Unreadable text on component signals list
##### LibrePCB VERSION
 0.1.0 (first_pcb-655-g6b08041-dirty)

##### OS / ENVIRONMENT
win7 x64

##### SUMMARY
when creating a new component via the dialog the leftmost text in the component signals table is unreadable because the font-size is too large / the row height is too small

##### STEPS TO REPRODUCE
create a new component via the new component dialog in the library editor

##### EXPECTED RESULTS
being able to read the text

##### ACTUAL RESULTS
cant read the text

image

### LINKED ISSUE #307: Schematic Editor: Add component dialog minimum height sometimes gliches
##### LibrePCB VERSION
 0.1.0 (first_pcb-655-g6b08041-dirty)

##### OS / ENVIRONMENT
win7x64
##### SUMMARY
add component dialog minimum height sometimes gliches while moving the dialog when a component is already selected

##### STEPS TO REPRODUCE
- open the 'add component' dialog of the schematic editor 
- select a component
- move the dialog by draging it on the title bar  
- ???????
- profit ^^ 

this **does not** happen if only a symbol is selected it has to be a whole component
##### EXPECTED RESULTS
the size of the dialog shouldn't change

##### ACTUAL RESULTS
the size increases 
**SPECULATION**: it looks like, since this doesn't occur if only a symbol is selected, or in fact if nothing is selected at all, at some point during the move the size of the symbol display is set back to 100% instead of the 50% it should be, since it shares its space with the package display

ubruhin:
- Component Editor: Fix unreadable text in signals list
  - Increasing the minimum row width to 17 pixels seems to fix the unreadable text on Windows.
  - Fixes #304 
- AddComponentDialog: Fix size glitches on Windows
  - Reorganize layout: Replace some HBox/VBoxLayouts by GridLayout.
  - Always show (empty) device/package, even if no device selected.
  - Fixes #307

### DIFF
--- a/libs/librepcb/libraryeditor/cmp/componentsignallisteditorwidget.cpp
+++ b/libs/librepcb/libraryeditor/cmp/componentsignallisteditorwidget.cpp
@@ -337,7 +337,7 @@ void ComponentSignalListEditorWidget::setTableRowContent(int row, const tl::opti
     mTable->setItem(row, COLUMN_FORCEDNETNAME, new QTableWidgetItem(forcedNetName));
 
     // button
-    int btnSize = requiredCheckBox->sizeHint().height();
+    int btnSize = qMax(requiredCheckBox->sizeHint().height(), 17);
     QToolButton* btnAddRemove = new QToolButton(this);
     btnAddRemove->setProperty("row", row);
     btnAddRemove->setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Fixed);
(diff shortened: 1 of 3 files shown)


SaculRennorb:
very nice, gj guys - the size glich was insanely annoying for actually working with the software

PR #320, opened 2018-09-01
Allow dragging symbols by circular grab areas

### LINKED ISSUE #311: Schematic Editor: Circle grab area of symbols not working
##### LibrePCB VERSION
0.1.0 (first_pcb-655-g6b08041-dirty)

##### OS / ENVIRONMENT
win7x64

##### SUMMARY
circle grab area isn't working on (custom) components , rectangle grab area works tho

##### STEPS TO REPRODUCE
create a custom component that uses a circle in its symbol and set the checkbox for grab area, then put it into your schema  - you will only be able to move the symbol by draging it on its center point, not the whole circle
##### EXPECTED RESULTS
circle with the checkbox set should define a grab area 

##### ACTUAL RESULTS
circles don't have a grab area

dbrgn:
I don't really know what I'm doing, but this seems to do the trick.

Fixes #311.

@ubruhin does the `mBoundingRect` need to be updated too? Do you have some high-level documentation on how all this code works, or what it does?

### DIFF
--- a/libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
+++ b/libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
@@ -23,7 +23,6 @@
 #include <QtCore>
 #include <QtWidgets>
 #include <QPrinter>
-#include <QPainterPathStroker>
 #include "sgi_symbol.h"
 #include "../items/si_symbol.h"
 #include "../schematic.h"
@@ -92,7 +91,7 @@ void SGI_Symbol::updateCacheAndRepaint() noexcept
             stroker.setCapStyle(Qt::RoundCap);
             stroker.setJoinStyle(Qt::RoundJoin);
             stroker.setWidth(2 * w);
-            mShape = stroker.createStroke(polygonPath);
+            mShape = mShape.united(stroker.createStroke(polygonPath));
         }
     }
 
--- a/libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
+++ b/libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
@@ -67,6 +67,7 @@ void SGI_Symbol::updateCacheAndRepaint() noexcept
     prepareGeometryChange();
 
     mBoundingRect = QRectF();
+
     mShape = QPainterPath();
     mShape.setFillRule(Qt::WindingFill);
 
@@ -77,10 +78,34 @@ void SGI_Symbol::updateCacheAndRepaint() noexcept
 
     // polygons
     for (const Polygon& polygon : mLibSymbol.getPolygons()) {
+        // query polygon path and line width
         QPainterPath polygonPath = polygon.getPath().toQPainterPathPx();
         qreal w = polygon.getLineWidth()->toPx() / 2;
+
+        // update bounding rectangle
         mBoundingRect = mBoundingRect.united(polygonPath.boundingRect().adjusted(-w, -w, w, w));
+
+        // update shape
         if (polygon.isGrabArea()) mShape = mShape.united(polygonPath);
+    }
+
+    // circles
+    for (const Circle& circle : mLibSymbol.getCircles()) {
+        // get circle radius, including compensation for the stroke width
+        qreal w = circle.getLineWidth()->toPx() / 2;
+        qreal r = circle.getDiameter()->toPx() / 2 + w;
+
+        // get the bounding rectangle for the circle
+        QPointF center = circle.getCenter().toPxQPointF();
+        QRectF boundingRect = QRectF(QPointF(center.x() - r, center.y() - r), QSizeF(r, r));
+
+        // update bounding rectangle
+        mBoundingRect = mBoundingRect.united(boundingRect);
+
+        // update shape
+        if (circle.isGrabArea()) {
+            mShape.addEllipse(circle.getCenter().toPxQPointF(), r, r);
+        }
     }
 
     // texts
--- a/libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
+++ b/libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
@@ -91,6 +91,9 @@ void SGI_Symbol::updateCacheAndRepaint() noexcept
             stroker.setCapStyle(Qt::RoundCap);
             stroker.setJoinStyle(Qt::RoundJoin);
             stroker.setWidth(2 * w);
+            // add polygon area
+            mShape = mShape.united(polygonPath);
+            // add stroke area
             mShape = mShape.united(stroker.createStroke(polygonPath));
         }
     }
--- a/libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
+++ b/libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
@@ -103,7 +103,7 @@ void SGI_Symbol::updateCacheAndRepaint() noexcept
 
         // get the bounding rectangle for the circle
         QPointF center = circle.getCenter().toPxQPointF();
-        QRectF boundingRect = QRectF(QPointF(center.x() - r, center.y() - r), QSizeF(r, r));
+        QRectF boundingRect = QRectF(QPointF(center.x() - r, center.y() - r), QSizeF(r * 2, r * 2));
 
         // update bounding rectangle
         mBoundingRect = mBoundingRect.united(boundingRect);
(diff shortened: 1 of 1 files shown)


ubruhin:
Nice :)

> does the mBoundingRect need to be updated too?

Yes, it even needs to be updated always (independent of whether the circle is a grab area or not).

> Do you have some high-level documentation on how all this code works, or what it does?

No, but `mBoundingRect` and `mShape` are used for [`QGraphicsItem::boundingRect()`](https://doc.qt.io/qt-5/qgraphicsitem.html#boundingRect) resp. [`QGraphicsItem::shape()`](https://doc.qt.io/qt-5/qgraphicsitem.html#shape) so there is some documentation. Instead of calculating them every time these methods are called, I calculate them only if something was modified (i.e. caching them for performance reasons).

dbrgn:
Thanks, PR updated.

By the way, I noticed that rectangular selection does not work properly for rectangles.

For circles, when going over the stroke width, it gets selected:

circle

(Mouse pointer isn't rendered on screenshot)

But for rectangles, you need to go over half of the stroke width for it to get selected:

rect

I assume it's a wrong mShape, but for the shape the polygon is used, so there would need to be a way to "grow" that polygon by half the stroke thickness somehow...

ubruhin:
> I assume it's a wrong mShape, but for the shape the polygon is used, so there would need to be a way to "grow" that polygon by half the stroke thickness somehow...

Yeah, that could be done with [`QPainterPathStroker`](https://doc.qt.io/qt-5/qpainterpathstroker.html). If you want to use it, please set following properties (because our pen is always round):

```
setCapStyle(Qt::RoundCap);
setJoinStyle(Qt::RoundJoin);
```

libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
```diff
@@ -66,7 +66,13 @@ void SGI_Symbol::updateCacheAndRepaint() noexcept
 {
     prepareGeometryChange();
 
+    // Docs: https://doc.qt.io/qt-5/qgraphicsitem.html#boundingRect
     mBoundingRect = QRectF();
+
+    // Docs: https://doc.qt.io/qt-5/qgraphicsitem.html#shape
+    // The default implementation calls boundingRect() to return a simple rectangular shape,
+    // but subclasses can reimplement this function to return a more accurate shape
+    // for non-rectangular items.
```
ubruhin: Hmm honestly I don't like these comments, because:
- We have over a dozen graphics items, it's not practical to add these comments to each of them.
- I don't see a reason to copy Qt's documentation into our source code. When working on our graphics items, it should be clear anyway to read Qt's documentation (graphics items aren't our concept, it's the concept of Qt). See also their complete graphics view documentation: https://doc.qt.io/qt-5/graphicsview.html

libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
```diff
@@ -77,12 +83,36 @@ void SGI_Symbol::updateCacheAndRepaint() noexcept
 
     // polygons
     for (const Polygon& polygon : mLibSymbol.getPolygons()) {
+        // get the bounding rectangle for the polygon
```
ubruhin: This comment is misleading as the next line does *not* get the bounding rect, it gets the exact outline. Better no comment than a wrong one :wink: (isn't the code self-explaining anyway?)

dbrgn: True, I'll fix that.

> isn't the code self-explaining anyway?

It took me a while to understand what it does. Especially if you're not familiar with the Qt graphics APIs. And even if the code is easy to understand, I think comments help to logically group multiple statements. But opinions on that vary :slightly_smiling_face:

ubruhin:
@dbrgn What's the current status of this PR?

dbrgn:
Ah, sorry, I almost forgot about it.

I'll try to do the requested fixes by tonight or tomorrow.

libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
dbrgn: Well, they would have helped me a lot :) But I removed them.

libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
dbrgn: Fixed the comment.

dbrgn:
> Yeah, that could be done with QPainterPathStroker.

Very nice, works :slightly_smiling_face: PR updated & rebased.

libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
```diff
@@ -77,10 +79,40 @@ void SGI_Symbol::updateCacheAndRepaint() noexcept
...
+
+    // circles
+    for (const Circle& circle : mLibSymbol.getCircles()) {
+        // get circle radius, including compensation for the stroke width
+        qreal w = circle.getLineWidth()->toPx() / 2;
+        qreal r = circle.getDiameter()->toPx() / 2 + w;
+
+        // get the bounding rectangle for the circle
+        QPointF center = circle.getCenter().toPxQPointF();
+        QRectF boundingRect = QRectF(QPointF(center.x() - r, center.y() - r), QSizeF(r, r));
```
ubruhin: I guess the calculation of the bounding rect is wrong. The second argument of `QRectF` is called *size* (i.e. width/height), but you pass the radius instead of the diameter.

libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
```diff
@@ -77,10 +79,40 @@ void SGI_Symbol::updateCacheAndRepaint() noexcept
...
         mBoundingRect = mBoundingRect.united(polygonPath.boundingRect().adjusted(-w, -w, w, w));
-        if (polygon.isGrabArea()) mShape = mShape.united(polygonPath);
+
+        // update shape
+        if (polygon.isGrabArea()) {
+            QPainterPathStroker stroker;
+            stroker.setCapStyle(Qt::RoundCap);
+            stroker.setJoinStyle(Qt::RoundJoin);
+            stroker.setWidth(2 * w);
+            mShape = stroker.createStroke(polygonPath);
```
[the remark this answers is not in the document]
dbrgn: D'oh, I didn't realize that it was in a loop... Sorry :slightly_smiling_face:

libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
dbrgn: Oh, well spotted. I think you're right :slightly_smiling_face: How could I have noticed that while testing? What's the effect of the bounding box on the behavior?

libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
```diff
@@ -23,6 +23,7 @@
 #include <QtCore>
 #include <QtWidgets>
 #include <QPrinter>
+#include <QPainterPathStroker>
```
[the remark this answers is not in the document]
dbrgn: I used that because it said so in the Qt docs, but it seems that it's not necessary indeed :slightly_smiling_face:

libs/librepcb/project/schematics/graphicsitems/sgi_symbol.cpp
ubruhin: It should be visible when enabling the graphics layer `GraphicsLayer::sDebugGraphicsItemsBoundingRects`, but unfortunately currently there is no way to enable it easily, so it needs to be done with an ugly hack in the source code ;)

The effect of a wrong bounding box is that the whole symbol disappears from the screen if the bounding box is outside the visible area. Mostly this happens of you scroll around, so that half of a symbol is outside the visible area, and the other half also disappears even if still in the visible area.

dbrgn: Ah, thanks, good to know!

ubruhin:
Sorry, still not working properly^^ Tested it right now and it turned out that polygons can no longer be grabbed on its body:

grafik

You have to re-add the line `mShape = mShape.united(polygonPath);` because this adds also the poyigons *area* to its shape, in addition to its outline.

dbrgn:
Ah, that's how the QPainterPathStroker works...

dbrgn:
Fixed now :sweat_smile:

ubruhin:
Thanks! :smiley: 

Do you want to rebase&squash by yourself or should I do it?

Issue #188, opened 2017-10-06
Define development process used after considering file format as stable


ubruhin:
Currently the development is done on the `master` branch, where breaking changes of the file format can occur at any time. An upgrade mechanism (see #185) may only be provided to upgrade one stable file format version to the next stable file format version, but not to upgrade intermediate (unstable) file format versions. So libraries and projects which were created with such a development snapshot may get completely useless (can no longer be opened) in future. This is *very* dangerous for end users!

I see two possible ways how to proceed developing after considering the file format as stable:
- Continue developing on the `master` branch, but warn users to *never* productively use LibrePCB compiled from the `master` branch. Not sure if this works (who reads such warnings...) :wink: 
- Develop on another branch (e.g. on `develop` as suggested in [git flow](http://nvie.com/posts/a-successful-git-branching-model/)) and use `master` only for stable releases. So LibrePCB compiled from the `master` branch is not dangerous and can be used productively at any time.

rnestler:
Wouldn't it be easiest to just provide upgrade paths from the released version to the current master?

So one could work on ones project with a released version and for testing open it with the current master version to check if the import works and to test the latest features.

ubruhin:
> Wouldn't it be easiest to just provide upgrade paths from the released version to the current master?

Of course, we always need to provide an upgrade path from the last stable release to the current development snapshot (i.e. `master` or `develop`). But we will *not* provide an upgrade path from the current development snapshot to the next stable release, so this upgrade path basically ends in a dead-end ;) And *this* is what we need to avoid to (accidentally) happen for end users.

rnestler:
@ubruhin I would consider master a development snapshot and only tags to be (semi-)stable releases. I also think git flow is a bit over engineering and not really useful if you don't plan to maintain several stable versions.

dbrgn:
I agree with @rnestler, but it would be good to show a warning in non-release builds.

dbrgn:
By the way, even though `master` gives no guarantees, I'd still expect file format experimentation / development to happen in feature branches, so it's probably rare that a file format change is done and then reverted or immediately changed again in master.

ubruhin:
Yep I agree too, I think we should continue developing on feature branches which are then merged into `master`. So the file format on `master` is always considered as unstable. But because of that, we really should add a scary warning, maybe in two ways:

- A comment in README.md
- Compiled into the application, for example a big scary warning in the Control Panel.

rnestler:
[...]

PR #326, opened 2018-09-09
Doxygen: Add release workflow documentation

ubruhin:
Our first release should be ready soon, so we need a strategy how to create and maintain official releases :) See #188.

This is my idea for a release workflow. Any thoughts about it?

Rendered: https://doxygen.librepcb.org/_branches/add-release-workflow-documentation/da/dbc/doc_release_workflow.html


dbrgn:
Looks good generally.

If I understand this correctly, then after releasing version 2.0.0, the master branch is immediately considered to be version 3.0.0, right? This would mean that if 90% of the features don't affect the file format, 90% of the features would need to be cherry-picked or merged to the release branch, for version 2.0.1, 2.1.0, etc. That seems like a lot of effort (especially considering that after the file format has stabilized, maybe a year or more can pass before the next major version is released).

dev/doxygen/pages/release_workflow.md
```diff
@@ -0,0 +1,155 @@
...
+development. Generally we do *not* maintain older releases than the last stable
+release because this would be too much effort. So for example as soon as the
+version "0.2.0" is released, the version "0.1.x" is no longer maintained.
+
+
+# Git Tags
+
+For every officially released version we create a Git tag on the corresponding
+release branch. The tag name contains all three numbers of the application's
+version number. Examples: "0.1.0-rc1", "0.1.0", "1.2.0", "1.2.1"
```
dbrgn: I'd propose that we PGP-sign all release tags and the version bump commits using a PGP key that we publish on our website and maybe on https://keybase.io/.

The same PGP key would be used for signing releases.

dev/doxygen/pages/release_workflow.md
```diff
@@ -0,0 +1,155 @@
...
+Because (at least) the `master` branch is always unstable, it's very dangerous
+for end users to work productively with this version (it could break projects
+etc.!). So we should warn users in that situation.
+
+This is done by displaying warnings in the application's GUI. Their visibility
+is controlled by the preprocessor define `FILE_FORMAT_STABLE`. It's set to `0`
+(warning visible) on all branches, except on release branches where it's set to
+`1` (warning disabled). As release branches always have a stable file format, it
+should be safe to use *any* state of a release branch (even if not officially
+released).
```
dbrgn: It's important that the warning is shown (and maybe acknowledged) *before* the user can open any project.

dev/doxygen/pages/release_workflow.md
```diff
@@ -0,0 +1,155 @@
...
+
+
+# Translations {#doc_release_workflow_translations}
+
+We use [Transifex] to translate LibrePCB into other languages. As Transifex
+doesn't have support for branches, we use a separate [resource] for each major
+application version. The translations for the `master` branch are contained in
+the resource `librepcb.ts`, while translations for releases are contained in
+resources like `librepcb-0.1.ts`.
```
dbrgn: Does Transifex sync translations between the different resources? So if someone translates strings in librepcb-0.1.ts, does that also affect `librepcb.ts`?

dev/doxygen/pages/release_workflow.md
```diff
@@ -0,0 +1,155 @@
...
+
+On the `master` branch, no translation files are checked into our source
+repository because they change very often (avoid commit spam). But on release
+branches, the translations are checked into the repository so they are always
+available when building official releases (important for packagers). Of course
+these translations are based on the Transifex resource for the corresponding
+major version, *not* on the latest `librepcb.ts` resource.
```
dbrgn: Would you integrate translations for released versions manually?

dev/doxygen/pages/release_workflow.md
```diff
@@ -0,0 +1,155 @@
...
+
+
+# Changelog {#doc_release_workflow_changelog}
+
+The changelog is not checked into the repository because it would probably lead
+to confusion (add it to the `master` branch, release branches or even both?).
+Instead, we use the [GitHub Releases] page to list the changes of every release
+(resp. tag).
```
dbrgn: Maybe it would be good to link those from the website (or even integrate the changelog into the website).

dev/doxygen/pages/release_workflow.md
```diff
@@ -0,0 +1,155 @@
...
+
+Releases built by ourselves (Windows builds, installers, AppImage etc.) are
+published at https://download.librepcb.org. The website http://librepcb.org
+contains direct download links to the corresponding files.
+
+Packages of package managers (e.g. APT, Flatpak, homebrew, ...) are updated by
+the corresponding package maintainers. Typically they clone the tag we released
+and then build the application from sources. Thanks to the translations checked
+into the release branches, package maintainers don't have to care about
+downloading translations from [Transifex].
```
dbrgn: Signing workflow is missing here :)

dev/doxygen/pages/release_workflow.md
ubruhin: Yeah for tags and releases it would be nice. But why the version bump commits?

dev/doxygen/pages/release_workflow.md
ubruhin: I guess the [Translation Memory](https://docs.transifex.com/setup/translation-memory/) should automatically fill in translated strings from any other resource with the same context.

Generally I would propose that translators always translate `librepcb.ts`, not resources of older releases. Translations of older releases should be updated only with the translation memory. But if the translation memory works well, maybe it even doesn't care on what resource translators are working on :)

dev/doxygen/pages/release_workflow.md
dbrgn: Mainly because - in contrast to tags - the commit signature is rendered/displayed on GitHub :slightly_smiling_face: Also, because commits are harder to manipulate than commits.

ubruhin: But how is a version bump commit different than any other commit?

> Also, because commits are harder to manipulate than commits.

:grin:

ubruhin: Ah, probably you mean the version bump commits on the release branches, not on master, and these commits are shown on GitHub because tags/branches point to them :)

ubruhin:
> If I understand this correctly, then after releasing version 2.0.0, the master branch is immediately considered to be version 3.0.0, right? This would mean that if 90% of the features don't affect the file format, 90% of the features would need to be cherry-picked or merged to the release branch, for version 2.0.1, 2.1.0, etc. That seems like a lot of effort (especially considering that after the file format has stabilized, maybe a year or more can pass before the next major version is released).

Correct. To avoid that, we could prepare releases directly on master (all the commits which I've drawn on the release branch would be on master then), and tags are created directly on master. Then the release branch and increment to the next major version could be done just before a file format change is needed. Such changes could even be hold back to stay on master as long as possible.

Overall this may require less effort because cherry-picking could be replaced by fast-forward merges, but I think it requires much more discipline on the master branch (on every commit on master you *must* be sure that it doesn't affect the file format) and also leads to a confusing Git history (sometimes release branches are equal to master, sometimes not; sometimes the file format on master is stable, sometimes not).

dev/doxygen/pages/release_workflow.md
```diff
@@ -0,0 +1,155 @@
...
+
+For the application we use version numbers according [semantic versioning]
+\(MAJOR.MINOR.PATCH, e.g. "1.2.3"). So every increment of the major number means
+a breaking change, which is a file format change in our case (we consider *any*
+file format change as breaking because we don't provide [forward compatibility]
+at all).
+
+As the file format also needs a version number, we just use the application's
+major version as the file format version. For example the application version
+"1.2" has the file format "1", and "13.37.42" has file format "13". Please note
```
rnestler: > 13.37.42

I see what you did there :wink:

ubruhin: Ooooh, that was not intended :rofl:

rnestler:
>  So for example as soon as the version "0.2.0" is released, the version "0.1.x" is no longer maintained.

So lets see if I get this right: We maintain the "0.1" branch as long as we are developing "0.2" on master. And all non file format breaking changes will actually be cherry picked to the 0.1 branch.
As soon as 0.2 is released we will stop maintaining 0.1 and start 0.3 development on master.

So does it even make sense to have the release branches then? The only time I can see value in it is if we have breaking changes in the 0.2/master branch and later on non-breaking changes which *do not depend* on the previous breaking changes which can be cherry-picked to 0.1.
Otherwise I could see a strategy where we develop 0.1.0 then 0.1.1 on master and as soon as we have a breaking change start naming master 0.2 and then just cherry pick small bugfixes into the 0.1 branch.

rnestler:
> Overall this may require less effort because cherry-picking could be replaced by fast-forward merges, but I think it requires much more discipline on the master branch (on every commit on master you must be sure that it doesn't affect the file format) and also leads to a confusing Git history (sometimes release branches are equal to master, sometimes not; sometimes the file format on master is stable, sometimes not).

Yeah I think this is exactly the trade off: The linear development on master is simpler and less maintenance burden, but requires some developer discipline in regard to file format breaking.

ubruhin:
> So lets see if I get this right

Yes, you got it right :)

> And all non file format breaking changes will actually be cherry picked to the 0.1 branch.

I wouldn't say *all*, but only the most useful ones, and only if the next release is not yet in sight. So generally I think we should always be focused on the next major release instead of maintaining already released versions. Who cares if between 1.0.0, 2.0.0, 3.0.0 etc. are no (or very few) other releases :)

> The linear development on master is simpler and less maintenance burden, but requires some developer discipline in regard to file format breaking.

In addition to discipline, it also needs to stop accepting new (even non-breaking) features while we are in release phase. For example after releasing v1.0.0-rc1 (directly on master), we must not accept anything else than bugfixes until v1.0.0 is released. Who knows how many release candidates need to be created and how long it takes until the final release. Release branches allow to stabilize releases (only cherry-picking bugfixes) while development continues on master.

dev/doxygen/pages/release_workflow.md
dbrgn: Signatures are shown for any signed commits. I only mentioned bump commits because I think it's too tedious to sign every commit, and the version bump commit indirectly signs the entire history chain that leads up to it :slightly_smiling_face:

dbrgn:
> Who cares if between 1.0.0, 2.0.0, 3.0.0 etc. are no (or very few) other releases :)

It should be as easy as possible to create new releases, because it will mean new features and fixes will reach users faster :slightly_smiling_face:

ubruhin:
> It should be as easy as possible to create new releases, because it will mean new features and fixes will reach users faster

For me it's more important to make releases as *stable* as possible, not publishing them as *early* as possible. And stabilizing releases is much easier on separate release branches than on master.

> But pull requests can already be filed and discussed.

And after weeks or months I have to ask contributors to rebase their branches (even if they were ready much sooner), which is annoying for contributors. I prefer that the pain of release management is on my side, not on contributors side ;)

> Alternatively, you could also create a next branch with only breaking changes that will eventually be merged into master before the next release. Basically an inverted cherry-picking workflow, where most features go onto master, and breaking changes go onto next.

And what would we have won? Then the pain is to merge `next` into `master` because this would be done only once a year or so. So the pain is just moved to somewhere else ;) In addition, one would have to choose the target branch of a PR depending on whether the PR changes the file format or not. That's annoying for contributors, and also GitHub doesn't allow to change the target branch afterwards so probably we would have to close and re-open PRs many times...

I think release management is a pain *anyway*, we can only choose where the pain is:
- Either the continuous development is a pain (hold back PRs in release phase, `next` branch, ...)
- Or maintaining older releases is a pain (cherry-picking)

In my opinion the latter option is much better for contributors (because they don't have to worry about release management at all) and leads to faster and easier development.

dbrgn:
> GitHub doesn't allow to change the target branch afterwards

That's not true anymore, since 1-2 years :slightly_smiling_face: 

But I think I agree about the release branches.

dev/doxygen/pages/release_workflow.md
ubruhin: > I only mentioned bump commits because I think it's too tedious to sign every commit, and the version bump commit indirectly signs the entire history chain that leads up to it

Hmm but the signed tag does exactly the same?!

dev/doxygen/pages/release_workflow.md
ubruhin: Yes, my plan is to add translations as a submodule, then I only have to update the submodule on release branches (pointing to a commit which contains the translations for the corresponding release).

ubruhin:
> That's not true anymore, since 1-2 years slightly_smiling_face

Ah, good to know :)

dev/doxygen/pages/release_workflow.md
dbrgn: Uhm, didn't you ask this question already?

> But why the version bump commits?

> Mainly because - in contrast to tags - the commit signature is rendered/displayed on GitHub :slightly_smiling_face:  Also, because commits are harder to manipulate than commits.

But it was just a suggestion, do whatever you want as long as something's signed.

dev/doxygen/pages/release_workflow.md
ubruhin: *How* it's implemented is not part of this documentation, so you may add your comment to the PR which actually implements this warning (which does not yet exist) :)

dev/doxygen/pages/release_workflow.md
ubruhin: Updated documentation accordingly.

dev/doxygen/pages/release_workflow.md
ubruhin: Done.

PR #327, opened 2018-09-15
Use new logo in all images

### LINKED ISSUE #6: Design an Icon for LibrePCB
The current icon is [downloaded from the internet](https://www.iconfinder.com/icons/294397/business_chip_digital_electronics_hardware_high_tech_inside_technology_icon) (it's license allows to use it for personal and commercial purpose). Even if the icon looks pretty nice, it would be better to design our own icon/logo, to be sure it is really unique. 

In addition, [iconfinder.com writes](http://support.iconfinder.com/694-legal/2527-license-overview):

> Using icons for a logo is never a good idea. By definition, icons are not unique and what you are looking for is something original that is not being used somewhere else. Icons are protected by license agreements that don't allow you to apply copyright on them, as you would do for a logo.

ubruhin:
Our new logo:

grafik

Fixes #6


eduardosm:
Maybe it's a good time to run a png optimizer on the new images?

ubruhin:
> Maybe it's a good time to run a png optimizer on the new images?

Good point, now it's really the best time to do it :+1:

I've optimized them with following command:

    pngcrush -rem allb -brute -reduce -ow librepcb.png

I first tried `optipng`, but it removed transparency effects, so it was *not* lossless :(

PR #329, opened 2018-09-16
Add translations as submodule

ubruhin:
Currently on CI we fetch translation files directly from Transifex to integrate them into the nightly builds. But this has several drawbacks:

- It's not possible to download translations without authentication, so you needs a Transifex account to download them. This is especially annoying for packagers, so LibrePCB packages would most probably be built without translations.
- It's hard to track the changes in translation files because they are not versioned.
- We have no backup - if something goes wrong with the translations, it will be a mess to restore them.

So I created the repository https://github.com/LibrePCB/librepcb-i18n and a script on our server which periodically downloads the latest translations from Transifex and commits them. This PR now includes that repository as a submodule, so translations are available for everyone without authentication. On release branches, the submodule will then even point to the translations for the corresponding release. See #326 for details.

This fixes all points listed above, while the most important advantage of this concept is that packagers don't have to worry about translations at all. When cloning tags, the correct translations for the corresponding release will automatically be cloned too, so translations should work out of the box.


dbrgn:
I think that's a great solution. Especially because it also allows us to see diffs of translations.

dbrgn:
Oh and maybe you should disable pull requests in the translations repo, to prevent people from changing the translations directly instead of going through Transifex?

ubruhin:
> Oh and maybe you should disable pull requests in the translations repo, to prevent people from changing the translations directly instead of going through Transifex?

Good point, but unfortunately (AFAIK) on GitHub it's not possible to disable PRs :sob: 

But I could add a note to the readme of the i18n repository that translations must always be done on Transifex and PRs are rejected.

dbrgn:
Oh, I wasn't aware of that.

Yep, I think a README note would suffice :) Maybe as big and fat as possible.

ubruhin:
Done -> https://github.com/LibrePCB/librepcb-i18n

PR #330, opened 2018-09-16
Implement versioning system according release workflow

ubruhin:
Following changes are needed for our new release workflow (see #326):

- Append "-unstable" to application version (to indicate that the version on the master branch is unstable). With that, the current application version is `0.1.0-unstable`.
- Use Git commit SHA as revision instead of "git describe" (the commit SHA is simple and reproducible, which "git describe" is not, it could even be confusing).
- Show warning in builds with unstable file format (not enabled yet because no stable release is available).

Once we enable the warning for unstable builds (after the first release), following dialog appears when starting LibrePCB built from an unstable branch:

grafik

I hope this is fine to warn users if they use an unstable build (maybe even by accident). For developers it's very annoying :wink: As developers should know what they do anyway, I added the possibility to disable the warning by creating the environment variable `LIBREPCB_DISABLE_UNSTABLE_WARNING=1`. The variable could for example be set in the run configuration of QtCreator, so it's only there when starting LibrePCB from within QtCreator.

Maybe there are better solutions for this warning?


dbrgn:
+1 for the warning and the env variable.

I'd mention backups in the warning text.

ubruhin:
> I'd mention backups in the warning text.

Done:

unstable file format _001

Also updated the readme and developers documentation.

PR #331, opened 2018-09-17
Doxygen: Add translation guidelines

dbrgn:
We need some kind of translation guidelines, especially for translating key terms.

I added such a page (still very basic), with a suggestion of some terms in German. I haven't found a word for all of them, and left some in English (since that seems to be how the words are being used).

PS: There are still a few links in CONTRIBUTING.md that need their URL updated to the new `developers` subdomain.

### DIFF
--- a/dev/doxygen/pages/translations.md
+++ b/dev/doxygen/pages/translations.md
@@ -1,9 +1,9 @@
 Translation Guidelines {#doc_translations}
-========================================
+==========================================
 
 [TOC]
 
-# Key Terms {#doc_versioning_file}
+# Key Terms {#doc_translations_terms}
 
 For consistent translations, certain LibrePCB specific terms (like "package" or
 "device") should always be translated in the same way.
@@ -15,8 +15,8 @@ For consistent translations, certain LibrePCB specific terms (like "package" or
 - Symbol Variant: Symbol-Variante
 - Signal: Signal
 - Component: Komponente
-- Device: ??
-- Part: ??
+- Device: Bauteil
+- Part: Typ
 - Package: Gehäuse
 - Pad: Pad
 - Footprint: Footprint
--- a/dev/doxygen/pages/translations.md
+++ b/dev/doxygen/pages/translations.md
@@ -1,0 +1,22 @@
+Translation Guidelines {#doc_translations}
+========================================
+
+[TOC]
+
+# Key Terms {#doc_versioning_file}
+
+For consistent translations, certain LibrePCB specific terms (like "package" or
+"device") should always be translated in the same way.
+
+## DE
+
+- Pin: Pin
+- Symbol: Symbol
+- Symbol Variant: Symbol-Variante
+- Signal: Signal
+- Component: Komponente
+- Device: ??
+- Part: ??
+- Package: Gehäuse
+- Pad: Pad
+- Footprint: Footprint
--- a/dev/doxygen/pages/translations.md
+++ b/dev/doxygen/pages/translations.md
@@ -8,15 +8,14 @@ Translation Guidelines {#doc_translations}
 For consistent translations, certain LibrePCB specific terms (like "package" or
 "device") should always be translated in the same way.
 
-## DE
-
-- Pin: Pin
-- Symbol: Symbol
-- Symbol Variant: Symbol-Variante
-- Signal: Signal
-- Component: Komponente
-- Device: Bauteil
-- Part: Typ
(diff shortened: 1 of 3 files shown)


dev/doxygen/pages/translations.md
```diff
@@ -0,0 +1,22 @@
...
+
+## DE
+
+- Pin: Pin
+- Symbol: Symbol
+- Symbol Variant: Symbol-Variante
+- Signal: Signal
+- Component: Komponente
+- Device: ??
+- Part: ??
```
dbrgn: Bauteil?

ubruhin: Maybe Device->Bauteil and Part->Typ

dev/doxygen/pages/translations.md
```diff
@@ -0,0 +1,22 @@
+Translation Guidelines {#doc_translations}
+========================================
```
ubruhin: line too short :)

dev/doxygen/pages/translations.md
```diff
@@ -0,0 +1,22 @@
...
+
+[TOC]
+
+# Key Terms {#doc_versioning_file}
```
ubruhin: `doc_versioning_file` -> `doc_translations_terms`

dev/doxygen/pages/translations.md
dbrgn: This is exactly why underlining sucks in Markdown :) The hash prefix variant is much nicer in that regard.

dbrgn:
It's now a table.

screenshot

Let's stick with this until it doesn't scale anymore, and then we can still break up the layout or create multiple tables.

ubruhin:
> Let's stick with this until it doesn't scale anymore, and then we can still break up the layout or create multiple tables.

:+1:

I forgot something: Could you also add the new page to mainpage.md? All other pages are also listed there...

And if you want to add some more terms: Workspace, Project, Library, Schematic, Board, Attribute

dbrgn:
Done.

dev/doxygen/pages/translations.md
```diff
@@ -0,0 +1,27 @@
...
+| **Pin** | Pin |
+| **Symbol** | Symbol |
+| **Symbol Variant** | Symbol-Variante |
+| **Component** | Komponente |
+| **Device** | Bauteil |
+| **Part** | Typ |
+| **Package** | Gehäuse |
+| **Pad** | Pad |
+| **Footprint** | Footprint |
+| **Workspace** | Workspace |
```
dbrgn: Couldn't find a translation for "workspace". I think English would be OK.

dev/doxygen/pages/translations.md
```diff
@@ -0,0 +1,27 @@
...
+| **Project** | Projekt |
+| **Library** | Library |
```
dbrgn: I left the translation at "Library" since "Bibliothek" sounds quite stupid. Do you agree?

dev/doxygen/pages/translations.md
```diff
@@ -0,0 +1,27 @@
...
+| **Schematic** | Schema |
+| **Board** | Board |
```
dbrgn: We could translate this as "Platine", but I like "Board" more.

dev/doxygen/pages/translations.md
ubruhin: :+1:

dev/doxygen/pages/translations.md
ubruhin: That's a tricky one... IMHO "Bibliothek" is not too bad, but also not very nice. I think we can go with "Library" for the moment - if it turns out another term would be better, we could still change it afterwards.

dev/doxygen/pages/translations.md
ubruhin: Yeah I think I would also go with "Board".

ubruhin:
@dbrgn Wouldn't it make sense to also add a short description *how* to actually translate? Just 2 or 3 sentences with links to our Transifex project, and to Transifex's translation guide.

ubruhin:
Aren't Transifex [Glossaries](https://docs.transifex.com/setup/glossary/) made exactly for this use case? :wink:

dbrgn:
Oh, I'll have to take a closer look at that!

dbrgn:
They are :)

ubruhin:
:+1: :grin:

PR #334, opened 2018-09-19
AppVeyor: Use ccache to speed up builds

ubruhin:
Unfortunately there are no official `ccache` binaries available, and building from source fails on Windows. Anyone interested in providing trustworthy binaries? If not, I'll use some random, unofficial binaries ;)

`ccache` reduces our AppVeyor build time from ~30min to ~10min!

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,14 +7,19 @@ deploy: off
 
 environment:
   global:
+    CC: "ccache gcc"
+    CXX: "ccache g++"
     MINGW: C:/Qt/Tools/mingw530_32
     QTDIR: C:/Qt/5.9/mingw53_32
     QTIFWDIR: C:/Qt/QtIFW-3.0.1
     VCRT_DIR: 'C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/Redist/x86/Microsoft.VC120.CRT'
     DEPLOY_INSTALLER: true
 
+cache:
+  - C:/msys64/home/appveyor/.ccache
+
 init:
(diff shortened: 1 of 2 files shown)


rnestler:
Can't you just install them with the MSYS2 package manager?

ubruhin:
No, the MSYS2 `ccache` package does not work when using a compiler outside of MSYS2 (I tried it, very strange errors...). And no, I can't use GCC from MSYS2, this would lead to much more annoying issues ;)

rnestler:
So you use which gcc exactly?

ubruhin:
MinGW which is provided by Qt because Qt's DLLs were compiled with that compiler.

ubruhin:
> pacman -S mingw-w64-x86_64-ccache

Oh wow, with that package it seems to work! :smiley:  Previously I installed the `ccache` package which has thrown very strange errors (some files seemed to be even compiled successfully, but some not)...

Updated this PR accordingly.

@rnestler How did you know that the `mingw-w64-x86_64-ccache` packages is needed, instead of just `ccache`?

rnestler:
> @rnestler How did you know that the mingw-w64-x86_64-ccache package is needed, instead of just ccache?

Well I noticed that you seem to use the mingw-w64 toolchain to build so I thought a matching ccache wouldn't hurt :wink:

ubruhin:
> Well I noticed that you seem to use the mingw-w64 toolchain to build so I thought a matching ccache wouldn't hurt wink

Hmm okay, I always wondered why there are different packages for the same thing, but after reading [this](https://github.com/msys2/msys2/wiki/MSYS2-introduction#subsystems), it sounds reasonable :grin: 

Thanks for the tip!

PR #337, opened 2018-09-30
Hide UUIDs in many windows and dialogs

ubruhin:
UUIDs are confusing for users, and were originally added only for debugging purposes anyway, so I removed them now from many windows and dialogs. In the library editor I added a menu item "File -> Show element in file manager" instead, so it's still easy to find the corresponding files in the file system if needed.


ubruhin:
With fce256203ef0cfc7b2f99bb83cb3997d71ae6d3e, a warning is now shown in the component editor if the component is not added to a category - I think now users should immediately see what they have to do :)

PR #346, opened 2018-10-06
AddComponentDialog: place a component by double-clicking list item

### LINKED ISSUE #344: Schematic Editor: Allow placing symbol with double click
##### SUMMARY

When listing symbols, allow placing a certain symbol by double clicking it. Right now, the user has to click "OK".

screenshot

In the screenshot above, double clicking "LED" in the middle column should place the LED symbol.

robbawebba:
To place a symbol from the 'Add Component' dialog, the user currently
has to select the desired symbol from the list and click 'OK'. This
change allows users to place a symbol by double-clicking the symbol in
the list without clicking the 'OK' button.

Fixes #344

### DIFF
--- a/libs/librepcb/projecteditor/dialogs/addcomponentdialog.ui
+++ b/libs/librepcb/projecteditor/dialogs/addcomponentdialog.ui
@@ -204,5 +204,21 @@
     </hint>
    </hints>
   </connection>
+  <connection>
+   <sender>treeComponents</sender>
+   <signal>itemDoubleClicked(QTreeWidgetItem*,int)</signal>
+   <receiver>librepcb::project::editor::AddComponentDialog</receiver>
+   <slot>accept()</slot>
+   <hints>
+    <hint type="sourcelabel">
+     <x>419</x>
+     <y>301</y>
+    </hint>
+    <hint type="destinationlabel">
+     <x>492</x>
+     <y>463</y>
+    </hint>
+   </hints>
+  </connection>
  </connections>
 </ui>
--- a/libs/librepcb/projecteditor/dialogs/addcomponentdialog.ui
+++ b/libs/librepcb/projecteditor/dialogs/addcomponentdialog.ui
@@ -52,6 +52,9 @@
      <property name="headerHidden">
       <bool>true</bool>
      </property>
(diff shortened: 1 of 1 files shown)


ubruhin:
Nice! :+1:

Only one thing: The tree view also reacts on double clicks to expand/collapse items which have children. So when double clicking on a parent item, the dialog is accepted, but the next time opening it, the clicked item is expanded or collapsed. I guess this could be solved by changing the `expandsOnDoubleClick` property to `false`, could you try that?

robbawebba:
@ubruhin Nice catch! I started playing around with the `ExpandsOnDoubleClick` property, but it doesn't seem to solve the problem. When that property is disabled, I can still double-click the parent item which still accepts the dialog without properly selecting a symbol to place.

I want to confirm/clarify the expected behavior of the items in tree view.
- If a tree item has children, then a double-click should expand the item to reveal it's children
- If a tree item has no children (is a leaf), then a double-click should accept the dialog with this item selected

Does this sound about right? If so, I'm not entirely sure how to handle this behavior. Is it possibly to write some c++ code to handle this custom behavior?

ubruhin:
Ah, no I think it's okay if double-clicking on an item with children also accepts the dialog. Collapsing/expanding must then be done with a single click on the arrow at the left of the item, which is fine in my opinion. So if that's the behavior with `expandsOnDoubleClick=false`, everything is fine :)

For clarification: The dialog allows to place either a component (no package defined) or a device (package defined). The tree view shows components, and only if a component has devices, they are shown as children.

robbawebba:
@ubruhin Thanks for the clarification! In that case, my most recent commit sets `expandsOnDoubleClick` to false.

ubruhin:
Perfect, thanks for your contribution @robbawebba :+1: :smiley:

robbawebba:
@ubruhin Happy to help!

I've recently transitioned to more of an embedded engineering role at my current company and have been reading a lot of schematics and board layouts lately while doing board bring-up. I've been thinking about possibly tinkering with my own small designs, just for practice with no intention actually build anything, and this seems like a pretty approachable program! I'll definitely be back :)

PR #358, opened 2018-10-26
Mark many strings as nontranslatable

dbrgn:
Please remove the "translatable" flag for units and unit-suffixes when creating new dialogs - saves time for the translators :slightly_smiling_face:

### DIFF
--- a/libs/librepcb/common/attributes/attrtypecapacitance.cpp
+++ b/libs/librepcb/common/attributes/attrtypecapacitance.cpp
@@ -37,13 +37,13 @@ namespace librepcb {
 
 AttrTypeCapacitance::AttrTypeCapacitance() noexcept
   : AttributeType(Type_t::Capacitance, "capacitance", tr("Capacitance")) {
-  mDefaultUnit = new AttributeUnit("microfarad", tr("μF"));
+  mDefaultUnit = new AttributeUnit("microfarad", "μF");
 
-  mAvailableUnits.append(new AttributeUnit("picofarad", tr("pF")));
-  mAvailableUnits.append(new AttributeUnit("nanofarad", tr("nF")));
+  mAvailableUnits.append(new AttributeUnit("picofarad", "pF"));
+  mAvailableUnits.append(new AttributeUnit("nanofarad", "nF"));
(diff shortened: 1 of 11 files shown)


ubruhin:
Are you really sure that these units are exactly the same in all existing languages? Even in Chinese etc.? I'm not sure if that's the case, so I decided to mark them as translatable :grin: 

The "%1%2" string is marked as translatable because I was also not sure if the order and spacing of physical values is exactly the same in all languages. For currency that's not the case (some languages have the number first, some have the unit first. Some have a space between number and unit, some not).

dbrgn:
> Are you really sure that these units are exactly the same in all existing languages? Even in Chinese etc.?

Yes, because these units are used like symbols.

From https://www.us-metric.org/faq-frequently-asked-questions-about-the-metric-system/#spelling-names: "SI unit symbols are the same in all languages, as you can see on this nutrition label from Japanese snack cakes."

> The "%1%2" string is marked as translatable because I was also not sure if the order and spacing of physical values is exactly the same in all languages.

Fair enough, but in that case you need a "context" explanation string for translators, otherwise everybody will simply use "%1%2" because there's no information on what %1 and %2 are.

By the way, at least in western languages, there should always be a space between the number and the unit symbol: https://www.us-metric.org/faq-frequently-asked-questions-about-the-metric-system/#spacing

ubruhin:
> Yes, because these units are used like symbols.

OK, good to know :)

> Fair enough, but in that case you need a "context" explanation string for translators

Sure, that would help :)

> By the way, at least in western languages, there should always be a space between the number and the unit symbol

I heard about that just a few months ago, and I have to admit that I really don't like that space :( IMHO schematics look much nicer if there is no space between value and unit...

So, do we now need translations for "%1%2" or not? If not, following lines:

```
return QString("%1%2").arg(QLocale().toString(v), unit->getSymbolTr());
```

could be simplified to:
```
return QLocale().toString(v) % unit->getSymbolTr();
```

dbrgn:
I applied the simplification now. If someone complains, we can always re-add it, with a proper translation context.

Branch rebased.

ubruhin:
Thanks :smiley:

PR #361, opened 2018-10-27
Allow deselecting items while keeping other items selected

### LINKED ISSUE #350: Allow deselecting items while keeping other items selected
##### SUMMARY

In all editors (symbol/package/schematic/board) it's currently possible to add single items to the selection by pressing the CTRL key while clicking on an item. But it's not possible to remove single items from the selection. Only the whole selection can be cleared at once.

It should be possible to deselect single items by clicking on a selected item while pressing the CTRL key.

Related RFC: https://github.com/LibrePCB/librepcb-rfcs/issues/29

##### IMPLEMENTATION

The selection logic is in the corresponding editor state machines, so this feature must be implemented there:
- `SymbolEditorState_Select`
- `PackageEditorState_Select`
- `SES_Select`
- `BES_Select`

I'm not completely sure, but I guess this should be an easy issue :smiley:

jameshclrk:
Fixes #350

### DIFF
--- a/libs/librepcb/projecteditor/schematiceditor/fsm/ses_select.cpp
+++ b/libs/librepcb/projecteditor/schematiceditor/fsm/ses_select.cpp
@@ -208,10 +208,16 @@ SES_Base::ProcRetVal SES_Select::proccessIdleSceneLeftClick(
     schematic.clearSelection();
     return ForceStayInState;
   }
-  if (!items.first()->isSelected()) {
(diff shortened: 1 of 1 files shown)


ubruhin:
Nice! :smiley: 

I quickly tried it. Toggle item selection works, but unfortunately now it's no longer possible to move all selected items together (selection gets cleared when start dragging). I hope you can fix that :)

jameshclrk:
I've rebased with a new approach, I think this addresses the moving of selected items :)

PR #363, opened 2018-10-29
Length: Implement mmStringToNm() to use only integer operations

eduardosm:
If tests should be extended or there are some style issues, or anything else, please, say so.

### DIFF
--- a/libs/librepcb/common/units/length.cpp
+++ b/libs/librepcb/common/units/length.cpp
@@ -25,6 +25,7 @@
 #include <QtCore>
 
 #include <limits>
+#include <type_traits>
 
 /*******************************************************************************
  *  Namespace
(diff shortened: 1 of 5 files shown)


ubruhin:
Awesome! :smiley: 

That's a lot of code xD Did you write it completely by yourself or is it copied from somewhere?

I couldn't yet review it in detail (will do that soon), just two things I noticed already:
- Test class `UnitsLengthTest` should be named only `LengthTest`
- Generally it would be better to write only one (or a few) `EXPECT_*()` statements per test function, and supply the function with different test data. See for example `PointTest::testRotate()`. It's not tragic, I would also merge it the way it is now :)

ubruhin:
> I wrote all myself grinning but its not the first time I wrote similar code to parse numbers.

OK, I see :)

> Done, I also moved the test to the parent (common) directory, because pointtest.cpp was there.

:+1:

> I took a look at it, but I'm not sure how to pass different test data to EXPECT_EQ and EXPECT_THROW, unless I create two different test classes.

I've found a test which does exactly the same thing: 

https://github.com/LibrePCB/LibrePCB/blob/35af6dc8d323821b1f9c7d17551e701c45434eee/tests/unittests/common/uuidtest.cpp#L161-L168

eduardosm:
> I've found a test which does exactly the same thing:
> 
> [LibrePCB/tests/unittests/common/uuidtest.cpp](https://github.com/LibrePCB/LibrePCB/blob/35af6dc8d323821b1f9c7d17551e701c45434eee/tests/unittests/common/uuidtest.cpp#L161-L168)

I have modified the tests accordingly

ubruhin:
Thank you! :)

PR #365, opened 2018-11-01
Length: Reimplement toMmString() to avoid using floating point

eduardosm:
I also removed a copy&paste leftover from the previous pull request :sweat_smile:


ubruhin:
It seems that you implemented it to use exponential representation (e.g. "1e-6") for small numbers. The old implementation didn't do that because I don't like exponential representation in files ;)

In addition, exact millimeters are now serialized without ".0" suffix. This is also bad for files because you can no longer distinguish between integer and floating point numbers in files. Writing e.g. "5.0" instead of "5" makes clear that this is a floating point number.

Could you adjust the code to get the old behavior back?

eduardosm:
I did it that way to make the result compact, but no problem, I'll change it

eduardosm:
Done :)

libs/librepcb/common/units/length.cpp
```diff
@@ -37,10 +37,51 @@ namespace librepcb {
...
 QString Length::toMmString() const noexcept {
-  QString str = QLocale::c().toString(toMm(), 'f', 6);
-  for (int i = 0; (i < 5) && str.endsWith(QLatin1Char('0')); ++i) {
-    str.chop(1);
+  using LengthBaseU_t = std::make_unsigned<LengthBase_t>::type;
+
+  LengthBase_t nm = toNm();
+  LengthBaseU_t nm_abs;
+  if (nm < 0) {
+    nm_abs = -static_cast<LengthBaseU_t>(nm);
```
ubruhin: Hmm this looks strange. The typecast `static_cast<LengthBaseU_t>(nm)` evaluates to a (large) unsigned number. The minus sign then converts it back to a signed number. And the assignment to `nm_abs` leads in an implicit cast back to an unsigned integer. Or am I misunderstanding something?

Shouldn't it just be that way?

```suggestion
    nm_abs = static_cast<LengthBaseU_t>(-nm);
```

eduardosm: AKAIC, signed overflow has undefined behavior, whereas unsigned overflow is well defined. The negation operator on a signed integer can overflow when the operand is the minimum value. The negation operator on an unsigned integer (bigger than `short`) will compute the two's complement negation and return it as an unsigned integer.

That's why I did the negation on the unsigned number, to handle the situation where `nm` has the minimum value, even though it's a corner case.

eduardosm:
> > I did it that way to make the result compact
> 
> Generally, compact is good. But our goal is also to have a canonical file format (i.e. well defined order and representation of every node in the file). Allowing exponential numbers would either break this goal, or requires us to document _exactly_ when to use exponential representation and when not. That in turn would make it much harder for 3rd-party Software to write valid, canonical LibrePCB files.
> 
> The same for the decimal point. But the decimal point is also useful for parsers of LibrePCB files to represent values in their correct type.

That makes sense, thanks for the explanation :+1:

libs/librepcb/common/units/length.cpp
ubruhin: Ah, good point! :smiley:

ubruhin:
Thanks @eduardosm :)

PR #368, opened 2018-11-04
Length: Reimplement mapNmToGrid() without using floating point

eduardosm:
Maybe its worth defining `LengthBaseU_t` outside the functions?
Regarding tests, I don't know how to pass different test data to different test methods from the same class.

### DIFF
--- a/libs/librepcb/common/units/length.cpp
+++ b/libs/librepcb/common/units/length.cpp
@@ -197,7 +197,7 @@ LengthBase_t Length::mapNmToGrid(LengthBase_t  nanometers,
   }
 
   LengthBaseU_t remainder = nm_abs % grid_interval;
-  if (remainder >= (grid_interval / 2)) {
+  if (remainder > (grid_interval / 2)) {
     // snap away from zero, but it might overflow
     LengthBaseU_t tmp_snapped = nm_abs + (grid_interval - remainder);
     if ((tmp_snapped < nm_abs) || (tmp_snapped > max)) {
--- a/libs/librepcb/common/units/length.cpp
+++ b/libs/librepcb/common/units/length.cpp
@@ -96,7 +96,11 @@ Length Length::abs() const noexcept {
 }
 
 Length& Length::makeAbs() noexcept {
-  mNanometers = qAbs(mNanometers);
+  if (mNanometers == std::numeric_limits<LengthBase_t>::min()) {
+    mNanometers = std::numeric_limits<LengthBase_t>::max();
+  } else {
+    mNanometers = qAbs(mNanometers);
+  }
   return *this;
 }
 
@@ -177,8 +181,7 @@ LengthBase_t Length::mapNmToGrid(LengthBase_t  nanometers,
                                  const Length& gridInterval) noexcept {
   using LengthBaseU_t = std::make_unsigned<LengthBase_t>::type;
 
-  // assume gridInterval >= 0
-  LengthBaseU_t grid_interval = static_cast<LengthBaseU_t>(gridInterval.mNanometers);
+  LengthBaseU_t grid_interval = static_cast<LengthBaseU_t>(gridInterval.abs().mNanometers);
   if (grid_interval == 0)
     return nanometers;
 
--- a/libs/librepcb/common/units/length.cpp
+++ b/libs/librepcb/common/units/length.cpp
@@ -200,7 +200,7 @@ LengthBase_t Length::mapNmToGrid(LengthBase_t  nanometers,
   if (remainder >= (grid_interval / 2)) {
     // snap away from zero, but it might overflow
     LengthBaseU_t tmp_snapped = nm_abs + (grid_interval - remainder);
-    if (tmp_snapped < nm_abs || tmp_snapped > max) {
+    if ((tmp_snapped < nm_abs) || (tmp_snapped > max)) {
       // overflow, snap towards zero
       nm_abs -= remainder;
     } else {
--- a/libs/librepcb/common/units/length.cpp
+++ b/libs/librepcb/common/units/length.cpp
@@ -175,11 +175,43 @@ void Length::setLengthFromFloat(qreal nanometers) {
 
 LengthBase_t Length::mapNmToGrid(LengthBase_t  nanometers,
                                  const Length& gridInterval) noexcept {
-  if (gridInterval.mNanometers != 0)
-    return qRound((qreal)nanometers / gridInterval.mNanometers) *
-           gridInterval.mNanometers;
-  else
+  using LengthBaseU_t = std::make_unsigned<LengthBase_t>::type;
+
+  // assume gridInterval >= 0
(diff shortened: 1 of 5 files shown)


libs/librepcb/common/units/length.cpp
```diff
@@ -175,11 +175,43 @@ void Length::setLengthFromFloat(qreal nanometers) {
 
 LengthBase_t Length::mapNmToGrid(LengthBase_t  nanometers,
                                  const Length& gridInterval) noexcept {
-  if (gridInterval.mNanometers != 0)
-    return qRound((qreal)nanometers / gridInterval.mNanometers) *
-           gridInterval.mNanometers;
-  else
+  using LengthBaseU_t = std::make_unsigned<LengthBase_t>::type;
+
+  // assume gridInterval >= 0
```
ubruhin: Could you make `gridInterval` absolute instead of assuming it (the old implementation also worked with negative grid interval)?

You could use `Length::toAbs()` to get the absolute value. But I have to admit that this method doesn't handle underflow properly :grin: You could either fix the handling of underflow (just map the lowest possible number to the largest possible number, although this would give an error of 1 nanometer), or just ignore that underflow (it will probably never happen anyway).

The value returned by `Length::toAbs()` could then be directly `static_cast<>`ed to an unsigned integer without worrying about underflow.

rnestler:
>  Regarding tests, I don't know how to pass different test data to different test methods from the same class.

We are using googletest as a unit test framework. See these stack overflow questions about parameterized tests:
* https://stackoverflow.com/questions/3152326/google-test-parameterized-tests-which-use-an-existing-test-fixture-class
* https://stackoverflow.com/questions/6256758/can-googletest-value-parameterized-with-multiple-different-types-of-parameters

You may also take a look at the gtest documentation: https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#value-parameterized-tests

libs/librepcb/common/units/length.cpp
```diff
@@ -175,11 +175,43 @@ void Length::setLengthFromFloat(qreal nanometers) {
...
+  } else {
+    nm_abs = -static_cast<LengthBaseU_t>(nanometers);
+    max = static_cast<LengthBaseU_t>(std::numeric_limits<LengthBase_t>::min());
+  }
+
+  LengthBaseU_t remainder = nm_abs % grid_interval;
+  if (remainder >= (grid_interval / 2)) {
+    // snap away from zero, but it might overflow
+    LengthBaseU_t tmp_snapped = nm_abs + (grid_interval - remainder);
+    if (tmp_snapped < nm_abs || tmp_snapped > max) {
```
ubruhin: (Probably) not wrong, but I like to have explicit parentheses to make operator precedence clear:

```suggestion
    if ((tmp_snapped < nm_abs) || (tmp_snapped > max)) {
```

tests/unittests/common/lengthtest.cpp
```diff
@@ -68,6 +68,17 @@ TEST_P(LengthTest, testToMmString) {
...
 
+TEST(LengthTest, testSnapToGrid) {
+  EXPECT_EQ(Length(8).mappedToGrid(10), Length(10));
+  EXPECT_EQ(Length(2).mappedToGrid(10), Length(0));
+  EXPECT_EQ(Length(-8).mappedToGrid(10), Length(-10));
+  EXPECT_EQ(Length(-2).mappedToGrid(10), Length(0));
+  EXPECT_EQ(Length(18).mappedToGrid(10), Length(20));
+  EXPECT_EQ(Length(12).mappedToGrid(10), Length(10));
+  EXPECT_EQ(Length(-18).mappedToGrid(10), Length(-20));
+  EXPECT_EQ(Length(-12).mappedToGrid(10), Length(-10));
```
ubruhin: Could you maybe add some tests for following cases?
- Input Length == 0
- Grid Interval == 0
- Grid Interval == 1
- Input Length == Grid Interval
- Input Length == Multiple of Grid Interval

ubruhin:
> Regarding tests, I don't know how to pass different test data to different test methods from the same class.

One option is to just add the two members `gridInterval` and `mappedToGrid` to `LengthTestData`. Of course all methods will run with all test data entries, but I think that's OK (other tests just ignore these new members). I did it that way in `FilePathTest`.

libs/librepcb/common/units/length.cpp
eduardosm: Done. I fixed the handling of overflow in `makeAbs`.

libs/librepcb/common/units/length.cpp
eduardosm: Done

tests/unittests/common/lengthtest.cpp
eduardosm: Done, it looks like it didn't work well with Grid Interval == 1 :smile:

eduardosm:
I have applied most of your suggestions 

> One option is to just add the two members `gridInterval` and `mappedToGrid` to `LengthTestData`. Of course all methods will run with all test data entries, but I think that's OK (other tests just ignore these new members). I did it that way in `FilePathTest`.

I will look into it later, but it doesn't look very scalable if we want to add more tests in the future

eduardosm:
Almost done, but it looks like that `testing::Values` has a limit on how many values it can take

ubruhin:
> I have applied most of your suggestions

Perfect :+1: :smiley: 

> I will look into it later, but it doesn't look very scalable if we want to add more tests in the future

That's true...

> Almost done, but it looks like that testing::Values has a limit on how many values it can take

Oh indeed, I didn't know that. So then probably the way to go is what @rnestler wrote (creating a separate test class for that single method). Would you give it a try?

Btw, nothing important, but according our coding style the members of `LengthTestData` should be lowerCamelCase instead of snake_case ;)

The implementation looks good for me now, so I'm ready to merge this.

eduardosm:
> Oh indeed, I didn't know that. So then probably the way to go is what @rnestler wrote (creating a separate test class for that single method). Would you give it a try?

I added a new test class and named it `LengthSnapTestData`, tell me if you think there is a better name

> Btw, nothing important, but according our coding style the members of `LengthTestData` should be lowerCamelCase instead of snake_case ;)

Oooops, fixed :grin:

ubruhin:
Awesome, thank you @eduardosm :)

Just out of curiosity, are you trying to run LibrePCB on a single-precision FPU or are you just going through the TODOs in the source code? :grin:

eduardosm:
I'm going through TODOs, because I wanted to contributed to this project, but I don't have time for anything more complex. I started with `Length` because I like numbers, besides programming and electronics :)

ubruhin:
OK I see, that's cool :D :+1:
Btw if you don't know it already, we also have [labeled some issues as "easy"](https://github.com/LibrePCB/LibrePCB/issues?q=is%3Aissue+is%3Aopen+label%3Aeasy) to help contributors finding issues to start :)

Issue #200, opened 2017-11-29
Library editor menu items are counter-intuitive


243750496:
![can not remove anything](https://user-images.githubusercontent.com/6311930/33375807-187b9cde-d546-11e7-9f26-cb84e29088f5.png)
hope you can fix

btw: if you choose another two to have a try , you wlll see all three was blue and cannot delete any of them

ubruhin:
Oh that's a misleading GUI. These menu items should be disabled if no library element is open. To remove whole library elements, currently the only way is to open it and then choose File -> Remove Element.

243750496:
find still no able to delete anythins with File -> Remove Element. and update library
 **maybe it's a bug**
![no use of delete elements](https://user-images.githubusercontent.com/6311930/33400796-21c7b11e-d592-11e7-9361-0cb1beffb89f.png)

243750496:
![right way to do delete element](https://user-images.githubusercontent.com/6311930/33420695-82b372d0-d5ea-11e7-9a42-960a9f80403f.png)
this is the rightway to do delete , in main page after opened the library manager , we can not delete it 

only can delete it with opened the item , please correct it ,  it's a worg guide to show up in main page without function

ubruhin:
@0xB767B Maybe you could improve the GUI to avoid confusion? Most important would be to disable/enable some menu items depending on the current state of the library editor.

0xB767B:
@ubruhin Are there any general concepts in LibrePCB how to enable/disable menu-items based on some state of the program?

ubruhin:
Not really. I guess the library editor is currently the only window which enables/disables some `QAction`s depending on which kind of library element is currently open. And the implementation of that is quite pragmatic ;)

Btw, now I noticed that some `QAction`s in the toolbars are disabled properly, but not their corresponding menu items (most `QAction` objects are added to both, a toolbar and a menu).

lukabosnjakovic:
I had similar problem few minutes a go. It is not intuitive that you must first open element and than go to Edit > Remove. My proposal is that maybe you add context menu (right click on element) with remove option. This is more intuitive to me.

dbrgn:
I'd nominate this for the 0.1 final :slightly_smiling_face:

ubruhin:
I'm ready to accept pull requests ;)

ubruhin:
Context menu added in #380.

[...]

PR #380, opened 2018-11-18
WIP: Add library element context menu

dbrgn:
Add context menu entry for removing the currently selected element.

Refs #200.

@ubruhin this is a first draft. Is the way I approached the issue correct? If yes, I can add support for the other library elements.

This could also be extended later on, e.g. when right-clicking on the empty space of the symbol list widget, we could add an entry called "Create symbol". When right-clicking on an existing symbol, we could add an entry called "Create symbol from this symbol" or something like that.

I think once this is in place, we can actually remove the delete menu entries, I think they're bad UX because it's not clear what they're referring to.


libs/librepcb/libraryeditor/lib/libraryoverviewwidget.cpp
```diff
@@ -103,6 +103,14 @@ LibraryOverviewWidget::LibraryOverviewWidget(const Context&          context,
   connect(mDependenciesEditorWidget.data(), &LibraryListEditorWidget::edited,
           this, &LibraryOverviewWidget::commitMetadata);
 
+  // Set up context menu triggers
+  mUi->lstSym->setContextMenuPolicy(Qt::CustomContextMenu);
```
ubruhin: This property can be set in the designer.

libs/librepcb/libraryeditor/lib/libraryoverviewwidget.cpp
```diff
@@ -135,6 +143,20 @@ bool LibraryOverviewWidget::save() noexcept {
   }
 }
 
+void LibraryOverviewWidget::symShowContextMenu(const QPoint &pos) noexcept {
+    QPoint globalPos = mUi->lstSym->mapToGlobal(pos);
+    QMenu menu;
+    menu.addAction(QIcon(":/img/actions/delete.png"), tr("Remove"), this, &LibraryOverviewWidget::symRemove);
```
ubruhin: I wouldn't use a separate function and signal/slot connection for `QAction` handlers. I think it's simpler to get the return value of `QMenu::exec()` to decide what action was triggered. Then you can also use [`QListWidget::itemAt()`](http://doc.qt.io/qt-5/qlistwidget.html#itemAt) to get the item which was actually right-clicked. Later we need it anyway to decide which context menu entries should be shown and which not...

See for example `SymbolEditorState_Select::openContextMenuAtPos()`.

ubruhin:
> @ubruhin this is a first draft. Is the way I approached the issue correct? If yes, I can add support for the other library elements.

Looks not bad so far :+1: Just don't forget that a library element may be opened at the time of its deletion, in that case the corresponding tab needs to be closed first.

> This could also be extended later on, e.g. when right-clicking on the empty space of the symbol list widget, we could add an entry called "Create symbol". When right-clicking on an existing symbol, we could add an entry called "Create symbol from this symbol" or something like that.

:+1:

> I think once this is in place, we can actually remove the delete menu entries, I think they're bad UX because it's not clear what they're referring to.

Would be nice if you could do it in this PR :)

dbrgn:
@ubruhin do you know how to prevent the following bug?

https://tmp.dbrgn.ch/librepcb-2018-11-18_23.06.34.webm

For some reason, the selection is not updated sometimes when right-clicking on another item while the context menu for the first item is still open.

dbrgn:
Regarding closing the tab if it's open:

- This is something I'd have to do in the libraryeditor.cpp, right? So I'd
  register a slot there, and emit an event after deleting the item in
  libraryoverviewwidget.cpp?
- How do I find out whether the element was opened in a tab?

I removed the remove action in the file menu, it's now replaced with the context menu. However, regarding the disabling of the delete/rotate items in the edit menu and toolbar, I think that's complex enough that it should be done in a separate PR. It doesn't have anything to do with removing library elements.

dbrgn:
PR is now refactored and uses `itemAt` to get the currently selected item. Furthermore, instead of the sender,  he selected item text as well as the file path are passed to the `removeSelectedItem` method.

libs/librepcb/libraryeditor/lib/libraryoverviewwidget.cpp
```diff
@@ -254,6 +268,64 @@ void LibraryOverviewWidget::updateElementList(QListWidget& listWidget,
...
+  // Execute the context menu
+  QAction* action = menu.exec(QCursor::pos());
+
+  // Get selected item
+  QListWidget* list = dynamic_cast<QListWidget*>(sender());
+  Q_ASSERT(list);
+  QListWidgetItem* selectedItem = list->itemAt(pos);
+
+  // Get item text
+  const QString itemName = selectedItem->text();
```
ubruhin: This line causes a crash if no item is selected.

libs/librepcb/libraryeditor/lib/libraryoverviewwidget.cpp
```diff
@@ -254,6 +268,64 @@ void LibraryOverviewWidget::updateElementList(QListWidget& listWidget,
...
+                 "other elements which depend on this one! They should be just "
+                 "marked as "
+                 "deprecated instead.\n\nAre you still sure to delete the "
+                 "whole library element "
+                 "\"%1\"?\n\nThis cannot be undone!"))
+          .arg(itemName),
+      QMessageBox::Yes, QMessageBox::Cancel);
+  if (ret == QMessageBox::Yes) {
+    try {
+      FileUtils::removeDirRecursively(itemPath);
```
ubruhin: If the selected element is currently opened in the library editor (as a tab), it must be closed before removing its directory.

libs/librepcb/libraryeditor/libraryeditor.h
```diff
@@ -139,6 +138,7 @@ class LibraryEditor final : public QMainWindow,
   void zoomOutTriggered() noexcept;
   void zoomAllTriggered() noexcept;
   void editGridPropertiesTriggered() noexcept;
+
```
ubruhin: Newline added by accident?

libs/librepcb/libraryeditor/libraryeditor.h
```diff
@@ -148,6 +148,11 @@ class LibraryEditor final : public QMainWindow,
   template <typename ElementType, typename EditWidgetType>
   void editLibraryElementTriggered(const FilePath& fp,
                                    bool            isNewElement) noexcept;
+
+  void removeSymbolTriggered(const FilePath& fp) noexcept;
+  void removeComponentTriggered(const FilePath& fp) noexcept;
+  void removeLibraryElementTriggered(const QString& elementType, const FilePath& fp) noexcept;
```
ubruhin: These methods do not exist in the *.cpp.

dbrgn:
Ok, only issue left is now the closing of the open tab.

dbrgn:
Ok, all requested issues fixed. If you want you can already review, but I also want to add a functional test :slightly_smiling_face:

libs/librepcb/libraryeditor/libraryeditor.h
```diff
@@ -126,10 +126,12 @@ class LibraryEditor final : public QMainWindow,
   // Operator Overloadings
   LibraryEditor& operator=(const LibraryEditor& rhs) = delete;
 
+public slots:
+  void closeTabIfOpen(const FilePath& fp) noexcept;
```
ubruhin: This method does not need to be public, it's never called (directly) from outside this class. In addition, it does not need to be marked as a slot, since C++11 also "normal" methods can be used as a slot when using the overload of `QObject::connect()` which takes function pointers. So I would suggest to move this method below `editDeviceTriggered()`, then the order corresponds to the `connect()` calls in the constructor :)

ubruhin:
Ah, I forgot: If you have clang-format 6.x installed, it would be nice if you could run `./dev/format_code.sh` to format the code ;)

dbrgn:
Ok, all requested issues fixed. If you want you can already review, but I also want to add a functional test :slightly_smiling_face:

dbrgn:
Functional tests are waiting for https://github.com/parkouss/funq/pull/47, so I think this can be merged.

PR #381, opened 2018-11-18
Add rule check for library elements

ubruhin:
Like a PCB DRC, but for library elements :smiley: 

librepcb-library-element-check


dbrgn:
Woah, this is amazing :smiley: It's much better to lint than just to document guidelines.

PR #386, opened 2018-12-02
Allow running clang-format in Docker container

dbrgn:
Are you interested in this?

It's a bit of a hack but seems to work nicely. Simply run the formatting script with `--docker`.

### DIFF
--- a/dev/clang-format/Dockerfile
+++ b/dev/clang-format/Dockerfile
@@ -1,10 +1,4 @@
-# Usage:
-#
-#   $ docker build . -t librepcb/clang-format:6
-#
-# Then from the main directory:
-#
-#   $ docker run --rm -t -i -v $(pwd):/code librepcb/clang-format:6 /bin/bash -c "cd /code && dev/format_code.sh"
+# Container with clang-format 6. Used in the `dev/format_code.sh` script.
 
 FROM ubuntu:bionic
 
--- a/dev/clang-format/Dockerfile
+++ b/dev/clang-format/Dockerfile
@@ -1,15 +1,17 @@
 # Usage:
 #
-# $ docker build . -t librepcb/clang-format:6
+#   $ docker build . -t librepcb/clang-format:6
 #
 # Then from the main directory:
 #
-# $ docker build . -t librepcb/clang-format:6
-# $ docker run --rm -t -i -v $(pwd):/code librepcb/clang-format:6 /bin/bash -c "cd /code && dev/format_code.sh"
+#   $ docker run --rm -t -i -v $(pwd):/code librepcb/clang-format:6 /bin/bash -c "cd /code && dev/format_code.sh"
+
 FROM ubuntu:bionic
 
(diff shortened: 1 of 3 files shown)


dev/clang-format/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# Usage:
+#
+# $ docker build . -t librepcb/clang-format:6
+#
+# Then from the main directory:
+#
+# $ docker build . -t librepcb/clang-format:6
```
ubruhin: Duplicate line (is already there a few lines above).

dev/clang-format/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+# $ docker run --rm -t -i -v $(pwd):/code librepcb/clang-format:6 /bin/bash -c "cd /code && dev/format_code.sh"
+FROM ubuntu:bionic
+
+RUN apt-get -q update \
+ && apt-get -qy install \
+  clang-format \
```
ubruhin: This does not guarantee that version 6 gets installed, it would be better to install `clang-format-6.0` and set it as default with `update-alternatives`.

dev/format_code.sh
```diff
@@ -1,14 +1,35 @@
...
 #   - Run the command "./dev/format_code.sh" in the root of the repository.
+#   - To run clang-format in a docker-container, use the "--docker" parameter.
 
 echo "Formatting modified files with clang-format..."
 
+DOCKER_IMAGE=librepcb/clang-format:6
+if [ "$1" == "--docker" ]; then
+  if [ "$(docker images -q $DOCKER_IMAGE | wc -l)" == "0" ]; then
+    echo "Building clang-format container..."
+    cd clang-format
```
ubruhin: This fails when running the script from the repository's root directory (according usage instructions).

dev/format_code.sh
```diff
@@ -1,14 +1,35 @@
...
+    docker build . -t librepcb/clang-format:6
+    cd ..
+  fi
+
+  echo "[Re-running format_code.sh inside Docker container]"
+  docker run --rm -t -i \
```
ubruhin: Maybe better add `--user $(id -u):$(id -g)` to improve security and avoid possible permission issues with the mounted volume?

dev/clang-format/Dockerfile
dbrgn: It does, as long as the distribution is ubuntu bionic. But I can also use update-alternatives if you want.

dev/format_code.sh
dbrgn: Oops, I used repo root detection when running, but not when building :)

dev/format_code.sh
dbrgn: Yeah, good point.

ubruhin:
Perfect, thanks :smiley: 

I will merge it as soon as CI succeeds...

PR #388, opened 2018-12-08
Make tabs closable with middle mouse button

dbrgn:
This has bugged me for a while :slightly_smiling_face: 

Code is loosely based on https://github.com/mbunkus/mkvtoolnix/commit/4c05f4263372153539535bd6bd401045c8f2ed61.

### DIFF
--- a/libs/librepcb/common/widgets/tabwidget.cpp
+++ b/libs/librepcb/common/widgets/tabwidget.cpp
@@ -1,0 +1,68 @@
+/*
+ * LibrePCB - Professional EDA for everyone!
+ * Copyright (C) 2013 LibrePCB Developers, see AUTHORS.md for contributors.
+ * https://librepcb.org/
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*******************************************************************************
+ *  Includes
+ ******************************************************************************/
+#include "tabwidget.h"
+
+#include <QtCore>
+#include <QtWidgets>
+
+/*******************************************************************************
+ *  Namespace
+ ******************************************************************************/
+namespace librepcb {
+
+/*******************************************************************************
+ *  Constructors / Destructor
+ ******************************************************************************/
+
+TabWidget::TabWidget(QWidget* parent) noexcept : QTabWidget(parent) {
+  tabBar()->installEventFilter(this);
+}
+
+/*******************************************************************************
+ *  Private Methods
+ ******************************************************************************/
+
+bool TabWidget::eventFilter(QObject* o, QEvent* e) {
+  // Handle middle mouse click on closable tabs
+  bool isTabBar           = o == tabBar();
+  bool isMouseButtonPress = e->type() == QEvent::MouseButtonPress;
+  bool tabsAreClosable    = tabBar()->tabsClosable();
+  if (isTabBar && isMouseButtonPress && tabsAreClosable) {
+    auto me = static_cast<QMouseEvent*>(e);
+    if (me->buttons() == Qt::MiddleButton) {
+      auto tabIdx = tabBar()->tabAt(me->pos());
+      if (tabIdx >= 0) {
+        emit tabCloseRequested(tabIdx);
+        return true;
+      }
+    }
+  }
+
+  return QTabWidget::eventFilter(o, e);
+}
+
+/*******************************************************************************
+ *  End of File
+ ******************************************************************************/
+
+}  // namespace librepcb
--- a/libs/librepcb/common/widgets/tabwidget.h
+++ b/libs/librepcb/common/widgets/tabwidget.h
@@ -1,0 +1,67 @@
+/*
+ * LibrePCB - Professional EDA for everyone!
+ * Copyright (C) 2013 LibrePCB Developers, see AUTHORS.md for contributors.
+ * https://librepcb.org/
+ *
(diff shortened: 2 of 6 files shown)


libs/librepcb/common/widgets/tabwidget.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include <QtCore>
+#include <QtWidgets>
+
+/*******************************************************************************
+ *  Namespace
+ ******************************************************************************/
+namespace librepcb {
+
+TabWidget::TabWidget(QWidget *parent) : QTabWidget{parent} {
```
ubruhin: For consistency:

```suggestion
TabWidget::TabWidget(QWidget *parent) : QTabWidget(parent) {
```

libs/librepcb/common/widgets/tabwidget.cpp
```diff
@@ -0,0 +1,61 @@
...
+  tabBar()->installEventFilter(this);
+}
+
```
ubruhin: For consistency, please insert this here (see other files):

```cpp
/*******************************************************************************
 *  Private Methods
```

dbrgn:
Rebased and reformatted.

libs/librepcb/common/widgets/tabwidget.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "tabwidget.h"
+
+#include <QtCore>
+#include <QtWidgets>
+
+/*******************************************************************************
+ *  Namespace
+ ******************************************************************************/
+namespace librepcb {
+
```
ubruhin: For consistency, please insert this here (see other files):

```cpp
/*******************************************************************************
 *  Constructors / Destructor
```

libs/librepcb/common/widgets/tabwidget.cpp
```diff
@@ -0,0 +1,61 @@
...
+TabWidget::TabWidget(QWidget* parent) : QTabWidget{parent} {
```
ubruhin: For consistency:

```suggestion
TabWidget::TabWidget(QWidget* parent) : QTabWidget(parent) {
```

libs/librepcb/common/widgets/tabwidget.cpp
```diff
@@ -0,0 +1,61 @@
...
+  // Handle middle mouse click on closable tabs
+  bool isTabBar           = o == tabBar();
+  bool isMouseButtonPress = e->type() == QEvent::MouseButtonPress;
+  bool tabsAreClosable    = tabBar()->tabsClosable();
+  if (isTabBar && isMouseButtonPress && tabsAreClosable) {
+    auto me = static_cast<QMouseEvent*>(e);
+    if (me->buttons() == Qt::MiddleButton) {
+      auto tabIdx = tabBar()->tabAt(me->pos());
+      if (tabIdx >= 0 &&
+          tabBar()->tabButton(tabIdx, QTabBar::RightSide) != nullptr) {
```
ubruhin: I would remove this check because the right side button might not always be the close button. See [Qt docs](http://doc.qt.io/qt-5/qtabbar.html#tabsClosable-prop):

> When tabsClosable is set to true a close button will appear on the tab on either the left or right hand side depending upon the style. When the button is clicked the tab the signal tabCloseRequested will be emitted.

In addition, there might be other ways to close a tab, for example a context menu.

But that's fine, it's up to the receiver of the `tabCloseRequested()` signal anyway to actually close the tab. So if you emit that signal on a non-closable tab the receiver can just ignore it. In our case this check should be moved to `LibraryEditor::tabCloseRequested()` instead.

libs/librepcb/common/widgets/tabwidget.cpp
```diff
@@ -0,0 +1,61 @@
...
+        emit tabCloseRequested(tabIdx);
+      }
+      return true;
```
ubruhin: Maybe better move this line into the block above, right after `emit tabCloseRequested(tabIdx);`? We're not interested in middle clicks outside of tabs, so we shouldn't consume these events.

libs/librepcb/common/widgets/tabwidget.h
```diff
@@ -0,0 +1,65 @@
...
+ * @author dbrgn
+ * @date 2018-12-07
+ */
+class TabWidget final : public QTabWidget {
+  Q_OBJECT
+
+public:
+  TabWidget(QWidget* parent);
+
+protected:
```
ubruhin: This could be `private` because the class is `final` anyway:

```suggestion
private:
```

libs/librepcb/common/widgets/tabwidget.h
```diff
@@ -0,0 +1,65 @@
...
+ * null, it emits the tabCloseRequested event.
+ *
+ * @author dbrgn
+ * @date 2018-12-07
+ */
+class TabWidget final : public QTabWidget {
+  Q_OBJECT
+
+public:
+  TabWidget(QWidget* parent);
```
ubruhin: Add `explicit` (to avoid surprises), `noexcept` (according our coding guidelines) and `= nullptr` (for consistency and to support default-constructing):
```suggestion
  explicit TabWidget(QWidget* parent = nullptr) noexcept;
```

In addition, please explicitly disable the copy constructor and assignment operator (see for example the `StatusBar` class).

libs/librepcb/common/widgets/tabwidget.cpp
dbrgn: Ok, I have no idea what the difference is :slightly_smiling_face:

ubruhin:
Hmm my review seems to be based on an older revision, although I started a new review after you pushed :( Maybe you need to run clang-format again after applying the suggestions ;)

libs/librepcb/common/widgets/tabwidget.cpp
dbrgn: Good point.

dbrgn:
Implemented requested changes.

dbrgn:
@ubruhin you have to merge the PR yourself, I don't have the permission to do so :slightly_smiling_face: (Or maybe I misunderstood your accept-without-merge?)

ubruhin:
> @ubruhin you have to merge the PR yourself, I don't have the permission to do so slightly_smiling_face (Or maybe I misunderstood your accept-without-merge?)

I know, I just dismissed my previous review and then had to wait until CI succeeds :)

Thanks for the improvement!

dbrgn:
Ah :slightly_smiling_face:

PR #398, opened 2018-12-25
Library Editor: Fix possibly outdated information in component signal editor

### LINKED ISSUE #392: Forbidden characters in signal names don't result in error message

```
LibrePCB Version: 0.1.0-rc2
Git Revision:     2800328030285b1c5094975a384a1759c9a4807f
Build Date:       2018-11-09T20:08:42
Qt Version:       5.12.0 (built against 5.11.2)
CPU Architecture: x86_64
Operating System: Arch Linux
Platform Plugin:  xcb
```

##### SUMMARY

See [this discussion](https://librepcb.discourse.group/t/allowed-characters-in-pin-names/53).

When I create a new component using the wizard and enter 3.3V for a signal, it is silently changed to 33V. After I added the signal, I can change it back to 3.3V. But it appears that it’s not saved (without an error message), since if I add another signal with the name 33V, LibrePCB complains that the signal with that name already exists. Furthermore, if I rename another signal, then the 3.3V signal name is again turned into 33V.

##### EXPECTED RESULTS

If a symbol is disallowed, entering it should result in an error message. It should not be removed silently.

picknick-:
componentSignalListedEditorWidget: Fix wrong behaviour when name of
signal contains forbidden characters.

Updating signal table when name and name without forbidden characters are
the same.

Fixes #392

### DIFF
--- a/libs/librepcb/libraryeditor/cmp/componentsignallisteditorwidget.cpp
+++ b/libs/librepcb/libraryeditor/cmp/componentsignallisteditorwidget.cpp
@@ -426,8 +426,8 @@ bool ComponentSignalListEditorWidget::setName(const Uuid&    uuid,
   try {
     ComponentSignal* signal = mSignalList->get(uuid).get();  // can throw
     if (signal->getName() == name) {
-        updateTable();
-        return true;
+      updateTable();
+      return true;
     }
     CircuitIdentifier constrainedName = validateNameOrThrow(name);  // can throw
     if (mUndoStack) {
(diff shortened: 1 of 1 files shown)


ubruhin:
Thanks @picknick-!

Looks good so far. But there is still a case where the table is not updated properly: If you enter a signal name which already exists, a message box appears which says that the signal name is already used. After closing the message box, the table still contains the entered name instead of reverting to the old name. Would you give it a try to fix this too? :)

picknick-:
Sure, fixed in latest commit.

ubruhin:
Works great now :)

I just noticed that you used 4 spaces as indentation, but our coding style is to us only 2 spaces (you could use clang-format to automatically format the code, see [details here](https://developers.librepcb.org/df/d24/doc_code_style_guide.html#doc_code_style_guide_format) - but in that simple case it's easy to do it by hand). If you fixed it, I think this is ready to merge.

ubruhin:
Thanks for updating! :smiley:

PR #405, opened 2019-01-09
Improve format code script

rnestler:
Fixes the issues reported by the shellcheck utility.

### DIFF
--- a/dev/format_code.sh
+++ b/dev/format_code.sh
@@ -36,9 +36,7 @@ if [ "$DOCKER" == "--docker" ]; then
 
   if [ "$(docker images -q $DOCKER_IMAGE | wc -l)" == "0" ]; then
     echo "Building clang-format container..."
-    cd "$REPO_ROOT/dev/clang-format"
-    docker build . -t librepcb/clang-format:6
-    cd -
+    docker build "$REPO_ROOT/dev/clang-format" -t librepcb/clang-format:6
   fi
 
   echo "[Re-running format_code.sh inside Docker container]"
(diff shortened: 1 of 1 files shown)


dev/format_code.sh
```diff
@@ -78,8 +81,7 @@ do
   done
 done
 
-echo "Finished: $COUNTER files modified."
-
+echo "Finished: $COUNTER files modified."
```
ubruhin: Hmm what happened here? ;)

dev/format_code.sh
```diff
@@ -34,13 +36,14 @@ if [ "$DOCKER" == "--docker" ]; then
 
   if [ "$(docker images -q $DOCKER_IMAGE | wc -l)" == "0" ]; then
     echo "Building clang-format container..."
-    cd "$REPO_ROOT/dev/clang-format"
-    docker build . -t librepcb/clang-format:6
-    cd -
+    (
+      cd "$REPO_ROOT/dev/clang-format"
+      docker build . -t librepcb/clang-format:6
+    )
```
ubruhin: An alternative might be to remove the `cd` and pass the path to `docker build` instead...

rnestler: That would indeed be simpler and cleaner.

dev/format_code.sh
rnestler: Huh no idea. I normally look at the git diff before commiting... Fixed.

dev/format_code.sh
rnestler: Fixed.

rnestler:
> Would you also be interested in fixing a bug? :grin: The script fails if files were removed (compared to master) because git diff also returns removed files, and those can't be found by clang-format...

Probably wouldn't be too hard. But not in this PR :wink:

ubruhin:
OK :)

Thanks for the fixup, looks good :+1:

PR #411, opened 2019-01-24
Add tolerance to placement clearance check

### LINKED ISSUE #403: Library messages: Rounding error when checking clearance
When checking for correct clearance (at least 150 µm) between silkscreen and copper, a rounding error occurs. See https://github.com/LibrePCB-Libraries/LibrePCB_Base.lplib/pull/17.

Example case: If the silkscreen ends at X 2.15 and the pad starts at X 2.3, the difference (2.3 - 2.15) is 0.14999999999, resulting in a warning even though without floating point rounding issues there would be no problem.

Example warnings:

img

dbrgn:
This gets rid of error messages due to rounding errors that occur if there is exactly 150 µm of clearance.

Seems to resolve #403.

### DIFF
--- a/libs/librepcb/library/pkg/packagecheck.cpp
+++ b/libs/librepcb/library/pkg/packagecheck.cpp
@@ -154,8 +154,9 @@ void PackageCheck::checkPadsOverlapWithPlacement(MsgList& msgs) const {
       std::shared_ptr<const FootprintPad> pad = it.ptr();
       std::shared_ptr<const PackagePad>   pkgPad =
           mPackage.getPads().find(pad->getUuid());
-      Length clearance(150000);  // 150um
-      Path   stopMaskPath = pad->getOutline(clearance);
+      Length clearance(150000);  // 150 µm
+      Length tolerance(000100); // 0.1 µm, to avoid rounding issues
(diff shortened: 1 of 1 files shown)


libs/librepcb/library/pkg/packagecheck.cpp
```diff
@@ -154,8 +154,9 @@ void PackageCheck::checkPadsOverlapWithPlacement(MsgList& msgs) const {
       std::shared_ptr<const FootprintPad> pad = it.ptr();
       std::shared_ptr<const PackagePad>   pkgPad =
           mPackage.getPads().find(pad->getUuid());
-      Length clearance(150000);  // 150um
-      Path   stopMaskPath = pad->getOutline(clearance);
+      Length clearance(150000);  // 150 µm
+      Length tolerance(000100); // 0.1 µm, to avoid rounding issues
```
ubruhin: I guess this is not 0.1µm because literals starting with `0` are interpreted as octal numbers ;) And shouldn't 0.01µm be enough?

dbrgn: D'oh :smile: Will fix.

dbrgn: Done.

PR #413, opened 2019-01-27
Package editor: Change default line width to 0.2mm

dbrgn:
As per https://github.com/LibrePCB/librepcb-doc/pull/35


ubruhin:
Thx! :+1:

Issue #390, opened 2018-12-10
The component with the UUID does not exist in the project's library!


shakenov-chinga:
Hi guys!
I was working on my new project, and decided to try LibrePCB for that purpose. You made quite a progress since last year and I want to thank you for your hard work :+1: .
So, I created several libraries and used it in the project. Then I made little changes in one of its component, updated libraries from within schematics window and got error window with message 
`The component with the UUID "aeb89f16-126b-40b7-84e1-242b0ec7eafd" does not exist in the project's library!`
It does exist, and it is from a different library, where I didn't touch anything.

I started LibrePCB using terminal, here is an output:
~/LibrePCB/stable/bin$ ./librepcb
Load stroke font: "newstroke.bene"
Start loading font "/home/chinga/LibrePCB/stable/share/librepcb/fontobene/newstroke.bene"
[  INFO   ] "LibrePCB 0.1.0 (d7458d3b3e126499902e1a66a0ef889f516a7c97)" (:0)
[  INFO   ] "Qt version: 5.9.1 (compiled against 5.9.1)" (:0)
[  INFO   ] "Resources directory: /home/chinga/LibrePCB/stable/share/librepcb" (:0)
[  INFO   ] "Application settings: /home/chinga/.config/LibrePCB/LibrePCB.ini" (:0)
[DEBUG-MSG] Started network access manager thread. (:0)
[ WARNING ] There was a stale lock on the workspace: "FilePath(/home/chinga/Dropbox/LibrePCB)" (:0)
[DEBUG-MSG] Load workspace settings... (:0)
[  INFO   ] Workspace settings file not found, default settings will be used. (:0)
[DEBUG-MSG] Workspace settings successfully loaded! (:0)
[DEBUG-MSG] Load local workspace library: "connectors.lplib" (:0)
[EXCEPTION] Not a valid URL. (../../../../libs/librepcb/common/fileio/../fileio/sexpression.h:282)
[EXCEPTION] File parse error: Not a valid URL.

/home/chinga/Dropbox/LibrePCB/v0.1/libraries/local/connectors.lplib/library.lp
Line,Column: -1,-1
Invalid Content: "" (../../../../libs/librepcb/common/fileio/../fileio/sexpression.h:98)
[...]

Issue #172, opened 2017-08-01
Create a file system abstraction layer


ubruhin:
A LibrePCB project consist of many single files which need to be read and written all together. Adding library elements to a project is actually a recursive copy operation of a directory. And the auto-save & restore-after-crash feature creates a copy of every single file.

So LibrePCB **heavily** uses file system operations. But unfortunately this is not very well implemented until now, many classes are very complicated because they must implement for example the restore-after-crash feature by themselves. I'm pretty sure there are some cases where bad things can happen (e.g. destroying project files) because of bugs.

I think it would be best to introduce some kind of file system abstraction. It should be responsible for:
- [x] Allowing to open whole projects and library elements in read-only mode
- [x] Creating file locks on opened projects or library elements
- [x] Periodically creating backup files + allowing to restore the last backup after a crash
- [x] Writing all files of a project or library element (pseudo-)atomically to the disk
- [ ] Allow to import/export LibrePCB projects from/to ZIP files
- [x] Load and save project library together with other files of the project (currently each library element handles loading/saving from/to files by itself, decoupled from the project)
- [x] Postpone all file operations until the project is saved the next time:
  - When adding a library element, its directory mustn't be copied to the project directory until the user saves the project. But the new library element must still be available! -> Save it to `/tmp`?
  - When adding a new sheet to the schematic, the new file mustn't be written to disk until it is saved
  - When removing a sheet from the schematic, the file mustn't be removed from disk until it is saved
- [ ] Maybe only write files which have changed (improve performance, reduce write operations)
- [ ] Watching for file modifications in the background (e.g. switching the git branch while a project is open)

[The doxygen documentation](http://librepcb.org/LibrePCB-Doxygen/master/d5/d75/doc_project.html#doc_project_save) describes (parts of) how this is implemented currently.

ubruhin:
The most important tasks are implemented in #422, thus closing this issue.

PR #422, opened 2019-02-28
Refactor file access with a transactional file system

ubruhin:
## Description

This is a complete refactoring of the file system access of library elements and projects. See #172 for the original motivation and goal. Although this PR does not implement all tasks defined in #172, it's still a huge improvement compared to the current concept.

Maybe this fixes the issues reported in #390 and [this forum thread](https://librepcb.discourse.group/t/cant-open-project/70), but I'm not completely sure because I couldn't reproduce these issues.

## Implementation

A new class `TransactionalFileSystem` allows to perform file operations (list/read/write/add/remove) without writing them to the harddrive immediately. Instead, all operations are cached in memory and can be written all at once to the harddrive in an extremely safe way. The changes are (pseudo-)atomically written to disk to guarantee consistent project files in every case. See [Atomic Saving Procedure](https://developers.librepcb.org/_branches/172-transactional-file-system/d5/d75/doc_project.html#doc_project_save) for details.

In addition, it allows to create and restore periodic backups (autosave). See [Automatic Periodic Saving](https://developers.librepcb.org/_branches/172-transactional-file-system/d5/d75/doc_project.html#doc_project_autosave) for details.

Basically it does the same as we already have, but with several improvements:
- Much more clean software architecture. Project related classes no longer need to know anything about backup/restore/autosave features, these are now all handled by the `TransactionalFileSystem` class. So the complexity of project related classes is reduced.
- The safety should be higher now, i.e. LibrePCB should behave much more robust in case of application crashes or file access issues. For example the autosave backups should be very reliable now.
- Much more future-proof because we can now perform file format upgrades in-memory, i.e. without writing an upgrade immediately to the harddrive. This is useful for users as they will be able to open projects with an older file format without the need of a one-way upgrade.

Overall, it was much harder than I initially thought, thus it cost me a lot of time to implement it :disappointed: That's the reason why there wasn't a lot of progress the last weeks... ;)


ubruhin:
I think this is ready now. I just hope there are no critical bugs left :wink:

Issue #418, opened 2019-02-11
Add support for slotted pad holes


ubruhin:
##### SUMMARY

Some THT footprints require pads with slotted holes, i.e. obrounds. So we should add support for them.

##### SOLUTION

Maybe just add a second dimension to the size information of the pad's drill, i.e. X/Y size instead of a diameter. In the Excellon output, use the G85 command if X and Y sizes aren't equal.

dbrgn:
Example: This footprint with oval / obround pads:

![image](https://user-images.githubusercontent.com/105168/63650770-af10c880-c74e-11e9-9d94-5d412f0d2b6f.png)

PR #423, opened 2019-03-03
Add support for slotted pad holes

pinoaffe:
Initial implementation of support for obround / slotted pad holes aka plated slots

Some sanity checks are still necessary (trying to make a drill bigger than the surrounding pad should probably throw an error), as are some UX changes, and better handling of file format versions

related issue: https://github.com/LibrePCB/LibrePCB/issues/418


libs/librepcb/common/cam/excellongenerator.cpp
```diff
@@ -47,7 +48,38 @@ ExcellonGenerator::~ExcellonGenerator() noexcept {
...
+    start.rotate(rotation, pos);
+    end.rotate(rotation, pos);
+
+    struct slotData temp = {
+      .start = start,
+      .end = end,
+      .width = width
+    };
+    mSlotList.append(temp);
+  }
```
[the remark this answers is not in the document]
pinoaffe: Yeah I wasn't quite sure where to put the conversion, but I figured that since the diameter+startpos+endpos representation is Excellon-specific, it would make sense to keep it local to Excellon-specific code, to keep the code as modular as possible, but putting the conversion logic in BoardGerberExport is also an option.

AFAIK Excellon slots with identical start- and endpoint are defined to be equivalent to drills, so it shouldn't matter, but it might indeed be cleaner to just create a slot anyhow

libs/librepcb/common/cam/excellongenerator.cpp
```diff
@@ -25,6 +25,7 @@
 #include "../fileio/fileutils.h"
 
 #include <QtCore>
+#include <QSet>
```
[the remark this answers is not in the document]
pinoaffe: okidoki

libs/librepcb/common/cam/excellongenerator.h
```diff
@@ -72,9 +73,19 @@ class ExcellonGenerator final {
...
+    PositiveLength width;
+  };
+
+
   // Excellon Data
-  QString                  mOutput;
-  QMultiMap<Length, Point> mDrillList;
+  QString                          mOutput;
+  QMultiMap<PositiveLength, Point> mDrillList;
+  QList<struct slotData>           mSlotList;
```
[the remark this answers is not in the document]
pinoaffe: I think a `QMultiMap<PositiveLength, QPair<Point, Point>>` would be a good idea, but I like having a seperate toollist better than repeatedly calling uniqueKeys and merging the resulting lists - your call tho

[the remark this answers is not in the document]
ubruhin: Ok, that's fine for me :)

libs/librepcb/common/cam/excellongenerator.cpp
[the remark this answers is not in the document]
pinoaffe: I moved the slot representation conversion logic into the BoardGerberExport class

ubruhin:
And the more I think about it, the more I would love to see following solution:

- Create class `DrillSize` with two `PositiveLength` members and with following functionality:
  - Serialization/deserialization from/to `SExpression` (instead of "abusing" `Point` for this task :wink:)
  - Convenience method `isCircular()` (to check if width==height, which is needed at several locations)
  - (note that this class can some time also be used by the `Hole` class as it makes sense to also allow non-plated oblongs)
- In `FootprintPad`, use `tl::optional<DrillSize>` to specify drill size.
- Don't take the pad's board side (top/bottom/THT) into account to decide whether there is a drill or not. If the `optional` contains a drill size, the drill should be drawn (and exported to Excellon), even if it's a SMT pad. Yes somehow it's strange, but for that case we could show a warning the package editor. IMHO it's even more strange to have a drill size specified, but not drawing the drill because the pad is SMT.
- Implement a clean file format upgrade path. When loading from file format 0.1 and board side is not THT, drill size must be set to `nullopt`. If it is THT, drill size must be initialized with width==height==diameter.

@pinoaffe What do you think about this?

Btw, could you run `clang-format` (see [docs](https://developers.librepcb.org/df/d24/doc_code_style_guide.html))? Or I could do it by myself by adding a commit to your branch if you like, just tell me.

libs/librepcb/library/pkg/footprintpad.cpp
```diff
@@ -134,9 +138,8 @@ QPainterPath FootprintPad::toQPainterPathPx(const Length& expansion) const
     noexcept {
   QPainterPath p = getOutline(expansion).toQPainterPathPx();
   if (mBoardSide == BoardSide::THT) {
-    p.setFillRule(Qt::OddEvenFill);  // important to subtract the hole!
-    p.addEllipse(QPointF(0, 0), mDrillDiameter->toPx() / 2,
-                 mDrillDiameter->toPx() / 2);
+    return p.subtracted(Path::obround(mDrillWidth, mDrillHeight).toQPainterPathPx());
+    // TODO: check whether this is centered correctly
```
ubruhin: Did you check this TODO? :grin:

libs/librepcb/library/pkg/msg/msgmalformeddrill.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+MsgMalformedDrill::MsgMalformedDrill(const FootprintPad& pad, errorType error) noexcept
+  : LibraryElementCheckMessage(
+        Severity::Error,
+        QString(tr("Malformed pad drill: '%1'")).arg(pad.getUuid().toStr()),
+        tr(QString(
+	    "The size of a drill may not exceed the pad size."
+	    "When it does, behaviour is undefined and it may not be plated."
+	    "In this case, the %1 of the drill exceeds the %1 of the pad."
+	).arg((error == WIDER) ? "width" : "height").toStdString().c_str())) {
```
ubruhin: 1. The order `tr(QString("..."))` is wrong, it should be `QString(tr("..."))` instead.
2. "width" and "height" need a `tr()` too, as they need to be translated
3. Writing `.toStdString().c_str()` in a Qt program is almost never the way to go (it's even dangerous, I guess this actually returns a `const char*` pointer to a temporary string which was deleted before you access it). I think this can be easily resolved once you fix point 1. above :smiley:

libs/librepcb/project/boards/boardgerberexport.cpp
```diff
@@ -331,8 +331,33 @@ int BoardGerberExport::drawPthDrills(ExcellonGenerator& gen) const {
...
       const library::FootprintPad& libPad = pad->getLibPad();
       if (libPad.getBoardSide() == library::FootprintPad::BoardSide::THT) {
-        gen.drill(pad->getPosition(),
-                  PositiveLength(*libPad.getDrillDiameter()));  // can throw
+        if (*libPad.getDrillWidth() == *libPad.getDrillHeight()) {
+          gen.drill(pad->getPosition(), libPad.getDrillWidth());
+        }
+        else {
+          if (libPad.getWidth() > libPad.getDrillWidth()
+              and libPad.getHeight() > libPad.getDrillHeight()) {
```
ubruhin: I would remove this check. If the drill is larger than the pad size, the generated Excellon file is wrong anyway, no matter if the drills are exported or not. IMHO it's a matter of the ERC to generate an error in that case, but the Excellon generator should just export the actual underlying data instead of applying some "magic" which doesn't fix the issue either.

libs/librepcb/library/pkg/footprintpad.cpp
pinoaffe: oh yes, will remove it

libs/librepcb/project/boards/boardgerberexport.cpp
pinoaffe: In such a case, where the input is broken, no output can be correct.
However, I'd rather generate sane output, which could have been correct given other input, than generate nonsensical output for which it is unclear what the board house will do

pinoaffe:
I implemented a basic DrillSize class, and put it to use.
I could only recall one place where a DrillSize object would be checked for being circular, so I did not implement an isCircular method.
If/When DrillSize is also used for holes or other unplated obrounds, it might be handy to add this and other such methods, but for now it seemed like an unnecessary abstraction.
I think it might be handy to add a generic file format versioning layer, for example with versioned sub-classes or container classes of SExpression.
Without this or something alike, small classes will need to be passed versioning info separately, or will need access to parent- or sibling nodes in order to be able to de-serialize correctly.

pinoaffe:
ah, there already seems to be another solution to versioning, described in https://developers.librepcb.org/d1/d69/doc_versioning.html , never mind then

ubruhin:
> I implemented a basic DrillSize class, and put it to use.

Thanks! :+1:

> I think it might be handy to add a generic file format versioning layer

Yep, you're absolutely right. I spent some time on this, and I think at the moment (to avoid requiring a large refactoring) it would be most reasonable to also pass the file format version to the constructors taking an `SExpression`...

I have now tested the slots feature on my machine. Unfortunately I saw several issues:

1. When saving a package which contains pads without drill, it cannot be loaded afterwards: `File parse error: Child not found: drill`
2. Obround pads are now rendered with many edges: 
grafik
3. In the package editor, THT pads cannot be converted to SMT because the drill doesn't get removed: 
grafik
3. Similarly, SMT pads cannot be modified at all, when clicking "Apply" (even without modifications), following error occurs: 
grafik
4. In the generated Excellon file I get a tool with a diameter of 1nm, with many drills drawn with that size. I don't know where this is coming from, but I suspect it has to do with the `UnsignedLength(1)` objects.
5. Either Gerbview is wrong or the slots are sometimes wrongly exported to Excellon. See the different rotation of slots in LibrePCB (top) vs. Gerbview (bottom):
grafik
6. We must not rely on the exceptions thrown by `tl::optional<>::operator->()` (is there even an exception thrown? where is this documented?) because we never catch exceptions which are not derived from our own exception class, so this would crash the application.
7. Some code is not exception-safe, in particular, methods which can throw exceptions are called from `noexcept` methods without a try/catch block -> potential application crashes.
8. The `DrillSize` class is located in `libs/librepcb/library` but is not in the `librepcb::library` namespace. Actually I would move it to `libs/librepcb/common` and remove all logic which is `FootprintPad`-specific.
9. There is still a `Makefile` commited ;)

I'm really sorry, but due to these issues I can't merge this PR in the current state. To be honest, I think it would consume a lot of time for both of us until everything is fine. To save time, I would offer to fix these issues by myself to get this merged soon, would this be OK for you? :)

libs/librepcb/project/boards/boardgerberexport.cpp
ubruhin: I see your point. But when sending nonsensical output to the board house, there is a chance to get feedback from the board house that your file is nonsensical ;) When silently removing the drills, the output looks OK and the board house will probably just manufacture it, then you get wrong boards (and loose money) for almost sure.

Of course I would not generate *invalid* output, i.e. invalid Excellon files. But in case of too large drills, the file is totally fine, it just doesn't make lot of sense.

pinoaffe:
I'd be willing to work on this some more, but feel free to take over.
In the latter case, I don't really have anything to add other than that I would not know where my code relies on exceptions thrown by `tl::optional<>::operator->()`

ubruhin:
> I'd be willing to work on this some more

Cool, thanks for your patience :smiley: 

So maybe you could fix some of the issues I explained above, and I could take over afterwards if there are still some issues left ;)

libs/librepcb/library/pkg/footprintpad.cpp
```diff
@@ -176,12 +176,43 @@ void FootprintPad::setHeight(const PositiveLength& height) noexcept {
...
+void FootprintPad::setDrillSize(const PositiveLength& width,
+                                const PositiveLength& height) noexcept {
+  mDrillSize = DrillSize(width, height);
+  if (mRegisteredGraphicsItem)
+    mRegisteredGraphicsItem->setShape(toQPainterPathPx());
+}
+
+void FootprintPad::setDrillWidth(const PositiveLength& width) {
+  // raises an exception if mDrillSize == nullopt
+  mDrillSize->setWidth(width);
```
ubruhin: > In the latter case, I don't really have anything to add other than that I would not know where my code relies on exceptions thrown by tl::optional<>::operator->()

For example here :)

In this case, I would just remove all get/setDrillWidth/Height from `FootprintPad`, I would only provide `getDrillSize()` and `setDrillSize()`...

pinoaffe:
>     1. When saving a package which contains pads without drill, it cannot be loaded afterwards: `File parse error: Child not found: drill`
This should be fixed now
I also made the `Version(const QVector<uint>& numbers)` in order to easily generate `Version` objects to compare to.
>     5. In the generated Excellon file I get a tool with a diameter of 1nm, with many drills drawn with that size. I don't know where this is coming from, but I suspect it has to do with the UnsignedLength(1) objects.
I was unable to replicate this, could you send me your test files?
>     6. Either Gerbview is wrong or the slots are sometimes wrongly exported to Excellon. See the different rotation of slots in LibrePCB (top) vs. Gerbview (bottom):
This is an issue in my code, and I have partially fixed this.
 I had wrongly assumed that as with the pad- and footprint positions, the rotation of the footprint would already have been added to the rotation of the FootprintPads. That should be fixed now
I was however unable to replicate the issue of "tearing", where the drills go halfway across the board.
>     9. The `DrillSize` class is located in `libs/librepcb/library` but is not in the `librepcb::library` namespace. Actually I would move it to `libs/librepcb/common` and remove all logic which is `FootprintPad`-specific.
I moved the `DrillSize` class and moved all `FootprintPad`-specific code into the `FootprintPad` class.
>     10. There is still a `Makefile` commited ;)
Should be fixed too


I'll take a look at exception safety and the editor next

ubruhin:
> I also made the Version(const QVector<uint>& numbers) in order to easily generate Version objects to compare to.

Unfortunately that's not legal because it allows to create invalid `Version` objects ;) See class documentation for details. I'd recommend to use `Version::fromString("0.1")` instead to create the objects for comparison.

> I was unable to replicate this, could you send me your test files?

Seems to be fixed now. I can still reproduce it with an older revision of this branch, but not with the current revision.

> This is an issue in my code, and I have partially fixed this.
I had wrongly assumed that as with the pad- and footprint positions, the rotation of the footprint would already have been added to the rotation of the FootprintPads. That should be fixed now
I was however unable to replicate the issue of "tearing", where the drills go halfway across the board.

Ok, I will check it again soon (unfortunatelly I have deleted my test project :sob:).

> I moved the DrillSize class and moved all FootprintPad-specific code into the FootprintPad class.

> I'll take a look at exception safety and the editor next

Perfect :+1:

libs/librepcb/common/cam/excellongenerator.cpp
```diff
@@ -47,7 +47,14 @@ ExcellonGenerator::~ExcellonGenerator() noexcept {
...
+  if (!mToolList.contains(dia)) mToolList.append(dia);
+}
+
+void ExcellonGenerator::slot(const Point& start, const Point& end,
+                             const PositiveLength& width) noexcept {
+  mSlotList.insert(width, qMakePair(start, end));
+  if (!mToolList.contains(width)) mToolList.append(width);
 }
 
 void ExcellonGenerator::generate() {
```
ubruhin, line 60: Since the tool list is no longer sorted (like `QMultiMap` which was used previously), the tool list now need to be sorted manually:

```suggestion
void ExcellonGenerator::generate() {
  std::sort(mToolList.begin(), mToolList.end());
```

I suspect this fixes the failed unit test (see CI output).

libs/librepcb/library/pkg/packagepad.h
```diff
@@ -28,6 +28,7 @@
 #include <librepcb/common/fileio/cmd/cmdlistelementremove.h>
 #include <librepcb/common/fileio/cmd/cmdlistelementsswap.h>
 #include <librepcb/common/fileio/serializableobjectlist.h>
+#include <optional/tl/optional.hpp>
```
ubruhin, line 31: Actually `optional` is not used in that file(?), thus let's revert this:

```suggestion
```

libs/librepcb/library/pkg/footprintpad.h
```diff
@@ -78,7 +80,13 @@ class FootprintPad final : public SerializableObject {
                Shape shape, const PositiveLength& width,
                const PositiveLength& height,
                const UnsignedLength& drillDiameter, BoardSide side) noexcept;
+  FootprintPad(const Uuid& padUuid, const Point& pos, const Angle& rot,
+               Shape shape, const PositiveLength& width,
+               const PositiveLength&          height,
+               const tl::optional<DrillSize>& drillSize,
+               BoardSide                      side) noexcept;
```
ubruhin, line 87: Same here about "convenience" constructors: Please don't add an additional constructor, but just adjust the existing constructor. The old constructor does not provide any advantage.

libs/librepcb/library/pkg/footprintpad.h
```diff
@@ -108,7 +116,8 @@ class FootprintPad final : public SerializableObject {
   bool setShape(Shape shape) noexcept;
   bool setWidth(const PositiveLength& width) noexcept;
   bool setHeight(const PositiveLength& height) noexcept;
-  bool setDrillDiameter(const UnsignedLength& diameter) noexcept;
+  bool setDrillSize(const DrillSize& drillSize) noexcept;
+  bool setDrillSize(const tl::optional<DrillSize>& drillSize) noexcept;
```
ubruhin, line 120: Same here: Please only adjust the old method instead of adding a new one. One method to set the drill size is enough.

libs/librepcb/library/pkg/msg/msgmissingdrill.cpp
```diff
@@ -0,0 +1,57 @@
...
+MsgMissingDrill::MsgMissingDrill(const FootprintPad& pad) noexcept
+  : LibraryElementCheckMessage(
+        Severity::Error,
+        QString(tr("Missing Drill in pad: '%1'")).arg(pad.getUuid().toStr()),
+        tr("One of the THT pads of one of the footprints of this package "
+           "does not have a drill configured. This means the pad is virtually "
+           "indistinguishable from a pair of SMD pads on opposite sides "
+           "of the board, i.e. they are not connected (despite all this, "
+           "LibrePCB will treat them as if they were connected.)")) {
+}  // namespace library
```
ubruhin, line 47: Wrong comment:

```suggestion
}
```

libs/librepcb/library/pkg/msg/msgdrillinsmdpad.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+MsgDrillInSmdPad::MsgDrillInSmdPad(const FootprintPad& pad) noexcept
+  : LibraryElementCheckMessage(
+        Severity::Error,
+        QString(tr("Drill in SMD pad: '%1'")).arg(pad.getUuid().toStr()),
+        tr("One of the SMD pads of one of the footprints of this package "
+           "has a drill set. This drill will not be present in the output "
+           "generated by LibrePCB. We recommend replacing the drill with "
+           "either a via or a non-plated hole in the board layout.")) {
+}  // namespace library
```
ubruhin, line 46: Wrong comment:

```suggestion
}
```

libs/librepcb/common/drillsize.h
```diff
@@ -0,0 +1,168 @@
...
+  DrillSize(const DrillSize& drillSize) noexcept
+    : mWidth(drillSize.mWidth), mHeight(drillSize.mHeight) {}
+
+  /**
+   * @brief Return an "optional" DrillSize object
+   *
+   * @param drillDiamater the intended diameter to be drilled
+   */
+  tl::optional<DrillSize> optionalDrillSize(
+      const UnsignedLength& drillDiameter);
```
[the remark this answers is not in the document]
pinoaffe: It is implemented and used in order to "parse" the drill sizes in footprint pads, should this be done differently?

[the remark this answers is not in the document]
pinoaffe: ah, never mind, I just saw that I accidentally duplicated the function definition as a member function of the class even though I ended up opting to not make it a member function, so I deleted the superfluous function prototype

libs/librepcb/library/pkg/footprintpad.h
pinoaffe: I removed everything I'd consider a "convenience constructor" and migrated other code to use the remaining ones

libs/librepcb/library/pkg/footprintpad.h
pinoaffe: I removed everything I'd consider a "convenience constructor" and migrated other code to use the remaining ones

libs/librepcb/library/pkg/footprintpad.cpp
```diff
@@ -218,12 +218,26 @@ bool FootprintPad::setHeight(const PositiveLength& height) noexcept {
...
+      }
+    } else {
+      mDrillSize = DrillSize(*drillSize);
+    }
+  } else {
+    if (mDrillSize)
+      mDrillSize = tl::nullopt;
+    else
+      return false;
   }
```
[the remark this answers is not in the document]
pinoaffe: This behavior is what the code was doing before my most recent change, I simply inlined the setDrillSize I deleted.
I didn't know there was a sensible == for tl::optional, but since there is, this code is indeed unnecessarily complicated.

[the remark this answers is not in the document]
pinoaffe: Okay so if I'm reading https://github.com/TartanLlama/optional/blob/master/include/tl/optional.hpp#L1980 correctly, simply doing `mDrillSize = drillSize;` would make mDrillSize refer to the DrillSize in drillSize (if there is one).
Because drillSize is and mDrillSize is not const, this does not seem desirable - am I seeing this correctly?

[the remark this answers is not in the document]
ubruhin: No, the link you posted is not the assignment operator used here. I think this is the correct one: https://github.com/TartanLlama/optional/blob/master/include/tl/optional.hpp#L1149

So it simply copies the contained value is there is one. There are no pointers or references used at all.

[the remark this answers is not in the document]
pinoaffe: Ah, sweet, that makes it a *lot* simpler

libs/librepcb/project/boards/drc/boarddesignrulecheck.cpp
```diff
@@ -405,18 +405,31 @@ void BoardDesignRuleCheck::checkMinimumPthRestring(int progressStart,
...
+                .arg(pad->getDisplayText().simplified(),
+                     formatLength(qMin(widthRestring, heightRestring)));
+        PositiveLength locWidth = PositiveLength(
+            drillWidth + *mOptions.minPthRestring + *mOptions.minPthRestring);
+        PositiveLength locHeight = PositiveLength(
+            drillHeight + *mOptions.minPthRestring + *mOptions.minPthRestring);
+        Path location =
+            Path::obround(locWidth, locHeight)
+                .translated(pad->getPosition())
+                .rotated(pad->getRotation() + device->getRotation());
```
ubruhin, line 432: In some cases, a wrong position is shown for this kind of DRC message:

grafik

I suspect you have to do the rotation first, and then the translation(?). And probably mirroring also needs to be considered.

Or as an alternative, just revert it to the original implementation which only shows a circle instead of an oblong :wink:

ubruhin:
And btw, after a quick test of the generated Excellon files, it looks like slots of mirrored devices are exported with wrong rotation:

grafik
grafik

libs/librepcb/project/boards/drc/boarddesignrulecheck.cpp
pinoaffe: oh yes, I'm guessing .rotated rotates around the origin, makes sense :), will fix

jakobwuhrer:
>     * The refactoring of the serialization/deserialization which I mentioned in my last comment unfortunately requires much more effort than expected :( So let's keep your implementation (which checks the number of SExpression childs instead of the file format version) for now. But could you please still serialize a null `DrillSize` as `none` instead of `0`?
Aight, will do

>     * Could you update this branch with the latest changes from `master`? This should fix the broken CI...
Done

>     * In the library editor, somehow I don't like to always enter both, the drill width and the drill height since in 99% of the cases one creates circular drills. Currently I see two possible variants to improve it:
Yeah, that could get annoying
>       * When changing the value in the "drill width" input field, automatically change the value in the "drill height" input field to the same value too. So in 99% of the cases one only has to specify the "drill width" to get circular drills. If a non-circular drill is needed, one has to change the "drill height" value accordingly. Not very nice solution IMHO, but probably simple to implement and much less annoying for users.
To me this seems like a very bad solution, potentially causing more grieve than it takes away. Automatically changing things the user enters is bad UI, and doing so without clearly communicating this fact through a common, well-known design seems like very bad UI
>       * Or add a checkbox "circular" or some checkable toolbutton to switch between circular and non-circular mode. In circular mode, both input fields (width/height) are always synchronized. See for example these widgets I found in GIMP (but don't ask me how to implement this with Qt see_no_evil ):
Yeah, that seems like the best solution, I'll try to look into how this could be achieved

pinoaffe:
> I reviewed now everything except the generated Excellon output (will do it soon). Overall it looks quite nice +1 See my comments for the smaller issues I found. In addition:
> 
>     * The refactoring of the serialization/deserialization which I mentioned in my last comment unfortunately requires much more effort than expected :( So let's keep your implementation (which checks the number of SExpression childs instead of the file format version) for now. But could you please still serialize a null `DrillSIze` as `none` instead of `0`?
Will do
>     * Could you update this branch with the latest changes from `master`? This should fix the broken CI...
Done
>     * In the library editor, somehow I don't like to always enter both, the drill width and the drill height since in 99% of the cases one creates circular drills. Currently I see two possible variants to improve it:
Seems annoying, yeah
>       * When changing the value in the "drill width" input field, automatically change the value in the "drill height" input field to the same value too. So in 99% of the cases one only has to specify the "drill width" to get circular drills. If a non-circular drill is needed, one has to change the "drill height" value accordingly. Not very nice solution IMHO, but probably simple to implement and much less annoying for users.
I don't think that's a good idea, automagically changing things the user enters seems like bad UI
>       * Or add a checkbox "circular" or some checkable toolbutton to switch between circular and non-circular mode. In circular mode, both input fields (width/height) are always synchronized. See for example these widgets I found in GIMP (but don't ask me how to implement this with Qt see_no_evil )
That seems like a good solution, will take a look

dbrgn:
@ubruhin now that #1071 has been merged, what are your plans for this feature / this PR? Could the code used in #1071 be re-used for slotted pad holes as well?

ubruhin:
> @ubruhin now that #1071 has been merged, what are your plans for this feature / this PR? Could the code used in #1071 be re-used for slotted pad holes as well?

Yeah I'm working on plated slots and plan to open a new PR soon. I'll close this one once the new PR is ready, but for now keeping it open for context.

ubruhin:
Closing this in favor of #1076 which is now merged. Sorry for the slow & difficult progress on this topic and thanks for your effort @pinoaffe! I think it was worth to replace it by a new approach which is now much more powerful :muscle:

PR #425, opened 2019-03-09
Library Editor: Extend functionality of overview lists

### LINKED ISSUE #347: Library Editor: Add context menus to overview lists
##### SUMMARY
The lists displaying the elements contained in a library should have a context menu. The menu should contain following entries:

- New Child Category (only on categories; should open the "New Library Element", pre-filled with the selected category as parent)
- Copy (should open the "New Library Element", pre-filled with the data from the selected element)
- Remove (must show the warning just like the menu item "File -> Remove Element")

grafik

EDIT: I forgot the most important menu item :sob:  The top most menu item should be "Edit", which has the same effect as double-clicking a list item. That item should be set as default action with `QMenu::setDefaultAction()`.

It would also be nice to allow multi-selection to allow opening and removing multiple elements at once ("New Child Category" and "Copy" must be disabled in that case).

##### IMPLEMENTATION

Generally the actual logic is already implemented (creating category children, copying existing elements and removing elements). So the new context menus should use this existing implementations to perform their actions.

Related to #200

ubruhin:
- Allow multi selection in lists (to edit or remove multiple elements at once)
- Enable "delete" shortcut to remove elements with the keyboard shortcut
- Add new context menu items to the library editor overview tab:
  - "Edit": Same as double click
  - "Copy": To open the "New Library Element" dialog, prefilled with the values of the selected element (disabled if multiple elements are selected)
- Add some more functional tests

grafik

Fixes #347

### DIFF
--- a/libs/librepcb/libraryeditor/lib/libraryoverviewwidget.cpp
+++ b/libs/librepcb/libraryeditor/lib/libraryoverviewwidget.cpp
@@ -153,6 +153,17 @@ bool LibraryOverviewWidget::save() noexcept {
   }
 }
 
+bool LibraryOverviewWidget::remove() noexcept {
+  if (QListWidget* list = dynamic_cast<QListWidget*>(focusWidget())) {
+    QHash<QListWidgetItem*, FilePath> selectedItemPaths =
+        getElementListItemFilePaths(list->selectedItems());
+    removeItems(selectedItemPaths);
+    return true;
+  } else {
+    return false;
+  }
+}
+
 /*******************************************************************************
  *  Private Methods
  ******************************************************************************/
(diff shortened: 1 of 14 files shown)


dbrgn:
Awesome!

If you want to add another small improvement: Add an entry "Add" to the context menu when clicking on empty space in one of those lists.

image

ubruhin:
> Add an entry "Add" to the context menu when clicking on empty space in one of those lists.

Good point -> done :smiley: 

But the menu item is also shown when clicking on an item, because in a full list it's not possible to click on empty space.

grafik

dbrgn:
> But the menu item is also shown when clicking on an item, because in a full list it's not possible to click on empty space.

Good point :slightly_smiling_face:

Issue #13, opened 2015-07-30
Add clipboard functionality (cut/copy/paste) to schematic/board/library editors


ubruhin:
Provide clipboard functionality in all graphical editors:

- [x] Symbol Editor
- [x] Package Editor
- [ ] Schematic Editor
- [ ] Board Editor

dbrgn:
May I nominate this for 0.1? If you have a lot of the same part, you want to multiply that part somehow, including rotation etc.

dbrgn:
Actually, never mind, 0.2 is probably fine since you can create boards without this :) It's a mere convenience.

ubruhin:
It probably requires really high effort because it's not trivial to implement this. And as it has no effect on the file format, I see no reason to give this high priority (although it would be very useful for users).

alfishe:
I would prefer the whole thing documented as user story, but still...

Schematic editor:
Assuming that copy operation affect all selected items (whole schematic page in worst case)
Q1: All selected items copied with their relative positions so layout kept after pasting. Right?
Q2: When paste is activated - where all that item layout should be placed? I.e. how to calculate top-left (or center point) for the pasted content?
Q3: Dealing with different units on source and destination pages / schematics
Q4: Library references to items used during copy / paste
Q5: What if nets on source / destination are different? Or opposite - the same (but we have different GUIDs for nets with same names)

EdizonTN:
Please I have another improvement - **snap cursor to grid** in case of use clipboard (as cut/copy origin). It's more powerful. 

My proposal for access: 
1. in case of use the clippborad's by icons (buttons) - use delta as 0
2. In case of use the clipboard's by CTRL+C (shorcut keys) - use delta of cursor aligned to grid. Next in paste with this delta, and use rotate/mirror function, also calculate with this delta (rotate around cursor aligned to grid).

doom-fr:
This functionality would be great to tidy up schematics with multiple pages, moving part of schematics from one page to another (now you have to delete and redraw everything.

Regarding that, I do not understand why it is still not developed as it seems to be easy comparing to other functionalities of LibrePCB.
Am I wrong about difficulty ?

ubruhin:
FYI, I'm working on this right now :)

> Am I wrong about difficulty ?

[...]

PR #442, opened 2019-04-16
Add clipboard functionality (cut/copy/paste) to symbol and package editors

ubruhin:
This implements the clipboard functionality for the library editor. It allows to copy symbol elements (pins, polygons etc.) between symbols, and footprint elements (pads, polygons etc.) between packages. This also works across different libraries, i.e. it's possible to copy elements from one library to another.

I even spent some time to copy the elements as an image to the clipboard, so after pressing CTRL+C you have nice images like these in your clipboard:

grafik grafik

Isn't this cool? :grin:

The only problem so far is that footprint pads can't be pasted if there is no corresponding package pad available, or if it is already used. That's a limitation of the current file format (a footprint pad must be assigned to a package pad). This might be fixed in a future file format version...

Fixes the first two tasks of #13.


dbrgn:
> Isn't this cool?

Haha, awesome!

> The only problem so far is that footprint pads can't be pasted if there is no corresponding package pad available, or if it is already used.

But if there are 8 package pads available, and you create a footprint pad, can you copy-paste that 7 times so that it uses the next free package pad every time?

ubruhin:
> But if there are 8 package pads available, and you create a footprint pad, can you copy-paste that 7 times so that it uses the next free package pad every time?

No, currently not. Basically because it could be dangerous as the user don't see to which package pad a footprint pad is connected to. The only way to see it is to open the pad properties dialog. So the pads might not be connected to the package pad you would expect. In addition, it's not clear which the "next free package pad" is - what order of pads should be used?

I'm not sure which variant would be better:
1. Automatically choose a "next free package pad", with the risk that it's not the one you would expect.
2. If no exactly matching pad is available, don't paste the pin, with the drawback that sometimes pasting is not possible, and it feels like a bug - but it's always safe.

I decided to go with variant 2. Would you prefer variant 1 @dbrgn?

dbrgn:
Yeah, maybe 2 is the safe choice for now.

I think it would be good if pads could be unconnected (with a big warning/error in the messages list). One reason for that is that right now you can't swap two package pads because you cannot temporarily disconnect one of them, and you can't have two footprint pads with the same package pad either.

ubruhin:
Yep, that's what I mean with the file format limitation ;) Once this limitation is removed, copy&paste of pads can be improved.

rnestler:
> Yep, that's what I mean with the file format limitation ;) Once this limitation is removed, copy&paste of pads can be improved.

I though a bit about that. Just because it can't be in the file format, doesn't mean it can't be in the editor at least temporarily. I think I'd be fine if it refuses to save to file if not all pads are assigned yet.

ubruhin:
> I though a bit about that. Just because it can't be in the file format, doesn't mean it can't be in the editor at least temporarily.

The internal data structure directly represents the file format, so it has the same restrictions.

> I think I'd be fine if it refuses to save to file if not all pads are assigned yet.

IMHO that would be horrible ;) For example if you're editing a package in the editor and then for some reason have to immediately stop working on this, it would be very annoying if you can't save your changes -> saving should be possible at any time.

PR #443, opened 2019-04-17
Add search field to symbol/package/component chooser dialogs

ubruhin:
In the library editor there are various locations where you need to choose a symbol, package or component with a dialog. This PR adds a search text field to these dialogs to help finding elements, basically just like we already have it in the "Add Component" dialog in the schematic editor.

grafik


dbrgn:
Cool! Does this also fix the search itself?

ubruhin:
With "fix the search itself" you mean the "Add Component" dialog in the schematic editor with the not very powerful search? Then no, this PR has nothing to do with that search. The "Add component" dialog is more complex because it searches for components and devices, while the dialogs in the library editor only search one element type which works perfectly.

PR #447, opened 2019-05-09
Library editor overview: Rename "copy" to "duplicate"

dbrgn:
The "copy" context menu action implies that there's a paste action.
However, there is no such thing. What really happens is a duplication, a
new library element is created from the current element.


dbrgn:
Not sure why the tests fail.

ubruhin:
> Not sure why the tests fail.

It's because of race conditions in Funq, see https://github.com/parkouss/funq/pull/55#issuecomment-450524551 for details. This is really annoying :(

PR #453, opened 2019-05-22
Library Editor: Preselect pin name when opening the pin properties dialog

EmbedME:
Preselecting the pin name improves the workflow when customizing pins: double-click pin, enter new name and finish with Enter. The deletion or manual selection of the previous name is no longer necessary.

### DIFF
--- a/libs/librepcb/libraryeditor/sym/dialogs/symbolpinpropertiesdialog.cpp
+++ b/libs/librepcb/libraryeditor/sym/dialogs/symbolpinpropertiesdialog.cpp
@@ -55,6 +55,9 @@ SymbolPinPropertiesDialog::SymbolPinPropertiesDialog(SymbolPin& pin,
   mUi->spbPosY->setValue(mSymbolPin.getPosition().getY().toMm());
   mUi->spbRotation->setValue(mSymbolPin.getRotation().toDeg());
   mUi->spbLength->setValue(mSymbolPin.getLength()->toMm());
+      
+  // preselect name
+  mUi->edtName->setSelection(0, mUi->edtName->text().size());
 }
 
 SymbolPinPropertiesDialog::~SymbolPinPropertiesDialog() noexcept {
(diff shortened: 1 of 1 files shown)


libs/librepcb/libraryeditor/sym/dialogs/symbolpinpropertiesdialog.cpp
```diff
@@ -55,6 +55,9 @@ SymbolPinPropertiesDialog::SymbolPinPropertiesDialog(SymbolPin& pin,
   mUi->spbPosY->setValue(mSymbolPin.getPosition().getY().toMm());
   mUi->spbRotation->setValue(mSymbolPin.getRotation().toDeg());
   mUi->spbLength->setValue(mSymbolPin.getLength()->toMm());
+
```
ubruhin: It seems that the GitHub webinterface doesn't show them, but my local Git client says there are whitespaces on this line. Would be nice to have them removed ;)

libs/librepcb/libraryeditor/sym/dialogs/symbolpinpropertiesdialog.cpp
```diff
@@ -55,6 +55,9 @@ SymbolPinPropertiesDialog::SymbolPinPropertiesDialog(SymbolPin& pin,
   mUi->spbPosY->setValue(mSymbolPin.getPosition().getY().toMm());
   mUi->spbRotation->setValue(mSymbolPin.getRotation().toDeg());
   mUi->spbLength->setValue(mSymbolPin.getLength()->toMm());
+      
+  // preselect name
+  mUi->edtName->setSelection(0, mUi->edtName->text().size());
```
ubruhin: Is there a reason for not using [`selectAll()`](https://doc.qt.io/qt-5/qlineedit.html#selectAll)? Would be cleaner and simpler IMHO...

libs/librepcb/libraryeditor/sym/dialogs/symbolpinpropertiesdialog.cpp
EmbedME: Done.

libs/librepcb/libraryeditor/sym/dialogs/symbolpinpropertiesdialog.cpp
EmbedME: I don't know the function. Of course it's nicer :-) Thanks for the hint.

ubruhin:
Looks good now, thanks! :smiley:

PR #455, opened 2019-05-24
Change predefined norms to "IEC 60617" + "IEEE 315"

ubruhin:
See discussion here: https://librepcb.discourse.group/t/how-do-i-add-ieee-315-to-my-preferences/88

I'm not completely happy with this solution, but it's for sure better than the current situation. IEC 60617 and IEEE 315 are the only two norms used in our official libraries so far, thus I think it makes sense to have these norms as suggestions in the workspace/project settings dialogs.

I also updated the component symbol variant editor dialog to use a combobox instead of a text field for specifying the norm, this way it should be more intuitive and easier to choose a norm.

### DIFF
--- a/libs/librepcb/common/common.pro
+++ b/libs/librepcb/common/common.pro
@@ -214,6 +214,7 @@ HEADERS += \
     network/networkrequest.h \
     network/networkrequestbase.h \
     network/repository.h \
+    norms.h \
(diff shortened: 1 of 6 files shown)


dbrgn:
Yeah, I think this makes sense. :+1:

PR #458, opened 2019-06-01
CI: Replace Travis-CI with Azure Pipelines

ubruhin:
Unfortunately Travis-CI is no longer able to build LibrePCB on MacOS because we reached the maximum build time of 50 minutes. So we need another CI service for the MacOS builds, and [Azure Pipelines](https://azure.microsoft.com/en-us/services/devops/pipelines/) looks very promising (thanks for the hint @rnestler :wink:). They provide a lot of free Linux, MacOS and Windows resources for open source projects :tada: 

To avoid having three different CI systems, I also migrated Linux builds from Travis-CI to Azure Pipelines, so we no longer need Travis-CI at all. Some time we can probably also replace AppVeyor with Azure Pipelines, but this migration requires much more effort so I keep AppVeyor for the moment.

One advantage of Azure Pipelines is that on Linux we can use custom Docker containers, which speeds up builds. I pushed some initial images to [Docker Hub](https://hub.docker.com/r/librepcb/librepcb-dev) and I will create a new repository with the Dockerfiles soon (it's still work in progress).

Btw, I also tried [Cirrus CI](https://cirrus-ci.org/) - it's also very promising (provides Linux, MacOS, Windows and even FreeBSD for free). It has some advantages compared to Azure Pipelines, but also some disadvantages. Due to more limited resources and some strange issues I had with MacOS builds, I think it's better to go with Azure Pipelines for the moment.


rnestler:
> One advantage of Azure Pipelines is that on Linux we can use custom Docker containers, which speeds up builds. I pushed some initial images to Docker Hub and I will create a new repository with the Dockerfiles soon (it's still work in progress).

The link is broken for me, https://hub.docker.com/r/librepcb/librepcb-dev works.
Also there already was a librepcb image on Docker hub: https://hub.docker.com/r/librepcb/librepcb. Was it not usable?

Also we should probably build and update the Docker images at a regular interval.

ubruhin:
> The link is broken for me, https://hub.docker.com/r/librepcb/librepcb-dev works.

True -> fixed in my comment above.

> Also there already was a librepcb image on Docker hub: https://hub.docker.com/r/librepcb/librepcb. Was it not usable?

That docker image was intended for (potential) users of LibrePCB to allow trying out LibrePCB without compiling it by yourself. This was two years ago, where we didn't have official binary releases. In addition, there was an additional tag to provide a complete environment for development, and one tag which was intended for CI but was never used.

But I'm not really happy with mixing three different use-cases into one docker image, having separate docker images is IMHO much more clean and easier to maintain. I just wonder whether I should name the image for CI `librepcb-dev` or `librepcb-ci` :wink:

In addition, I no longer see the need for providing a Docker image containing LibrePCB, QtCreator etc. GUI applications are a pain with Docker, and now there are much better deployment variants of LibrePCB available (installer, AppImage, Flatpak, AUR, ...). So I wonder it I should just remove the `librepcb/librepcb` docker image (it's not maintained anyway for 2 years).

ubruhin:
Should be ready now. Dockerfiles are here: https://github.com/LibrePCB/docker-librepcb-dev

ci/azure-jobs-doxygen.yml
```diff
@@ -0,0 +1,16 @@
+jobs:
+- job: Doxygen
+  pool:
+    vmImage: 'ubuntu-16.04'
+  container: librepcb/librepcb-dev:ubuntu-14.04-1
```
rnestler: Why Ubuntu 14.04? It is end of life since April 2019.

ci/azure-jobs-linux.yml
```diff
@@ -0,0 +1,54 @@
...
+- job: Linux
+  pool:
+    vmImage: 'ubuntu-16.04'
+  variables:
+    OS: 'linux'
+    ARCH: 'x86_64'
+  strategy:
+    matrix:
+      Ubuntu_1404_GCC:
+        IMAGE: librepcb/librepcb-dev:ubuntu-14.04-1
```
rnestler: Also here: Why Ubuntu 14.04? It is end of life since April 2019.

ci/azure-jobs-doxygen.yml
ubruhin: I tried a more recent Ubuntu version, but for some reason Doxygen failed to run `dot` to create the diagrams. As 14.04 was already used until now, I kept this for the moment - but yes, this should be fixed some day ;)

ci/azure-jobs-linux.yml
ubruhin: Because 14.04 provides Qt 5.2 (and GCC 4.8), which LibrePCB still supports, and currently I see no reason to remove this support :wink:

rnestler:
> So I wonder if I should just remove the librepcb/librepcb docker image (it's not maintained anyway for 2 years).

We should probably at least mark it deprecated.

ci/azure-jobs-doxygen.yml
rnestler: Alright. I created https://github.com/LibrePCB/LibrePCB/issues/461 to resolve it later.

ci/azure-jobs-linux.yml
rnestler: I'm still amazed by your dedication to support legacy systems :slightly_smiling_face: But I'd consider then supporting Debian Jessie which is still in LTS and has gcc 4.9 and Qt 5.3. But there is no point in bloating this PR with this.

ci/azure-jobs-doxygen.yml
ubruhin: Thanks! :+1:

ci/azure-jobs-linux.yml
ubruhin: :+1: Adding more CI jobs makes sense anyway, but yes, this can be done later in a separate PR.

ubruhin:
> We should probably at least mark it deprecated.

Done.

Thanks for reviewing :)

PR #468, opened 2019-06-15
Add a --sudo option to format_code.sh

rnestler:
This allows to run docker with sudo for users which are not in the
docker group on their system.


dbrgn:
Maybe add a usage hint to the comment at the top of the file?

rnestler:
> Maybe add a usage hint to the comment at the top of the file?

Done.

PR #470, opened 2019-06-18
Automatically choose next free numerical pad name

### LINKED ISSUE #469: Package editor usability regression

```
LibrePCB Version: 0.1.1-unstable
Git Revision:     abf328a53eadfbf24ffd9cc9a1acb68104fb6a2d
Build Date:       2019-06-06T13:36:30
Qt Version:       5.9.5 (built against 5.9.5)
CPU Architecture: x86_64
Operating System: Ubuntu 18.04.2 LTS
Platform Plugin:  xcb
```

##### SUMMARY
Adding a new pad without a name gets you an error message

##### STEPS TO REPRODUCE

Try to add a new pad without a name.
new-behaviour

##### EXPECTED RESULTS
The old behavior just choose the next free numeral name:
old-behaviour-2
old-behaviour

##### ACTUAL RESULTS
An error pop up appears.

new-behaviour-2

##### ADDITIONAL INFORMATION

Working version was revision 4b2fe09494395d608464832110b5b656f1a96f35

I suppose #456 brought the regression.

rnestler:
If no pad name is set search for the next free numerical one.

Fixes #469

### DIFF
--- a/libs/librepcb/library/pkg/packagepadlistmodel.cpp
+++ b/libs/librepcb/library/pkg/packagepadlistmodel.cpp
@@ -88,6 +88,10 @@ void PackagePadListModel::addPad(const QVariant& editData) noexcept {
   try {
     QScopedPointer<UndoCommandGroup> cmd(
         new UndoCommandGroup(tr("Add package pad(s)")));
+    // if no name is set we search for the next free numerical pad name
+    if (mNewName.isEmpty()) {
+      mNewName = getNextPadNameProposal();
+    }
     foreach (const QString& name, Toolbox::expandRangesInString(mNewName)) {
       std::shared_ptr<PackagePad> pad = std::make_shared<PackagePad>(
           Uuid::createRandom(), validateNameOrThrow(name));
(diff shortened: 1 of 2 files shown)


libs/librepcb/library/pkg/packagepadlistmodel.cpp
```diff
@@ -88,6 +88,17 @@ void PackagePadListModel::addPad(const QVariant& editData) noexcept {
   try {
     QScopedPointer<UndoCommandGroup> cmd(
         new UndoCommandGroup(tr("Add package pad(s)")));
+    // if no name is set we search for the next free numerical pad name
+    if (mNewName.isEmpty()) {
+      for (unsigned i = 1; i < 32; ++i) {
```
rnestler: This obviously only works for pad names up to 31. Is there a better solution for this problem?

rnestler: Up until 4a03d8491b7421af79935fdd7216b4c2fdd2494a it was done like this:
```cpp
QString PackagePadListEditorWidget::getNextPadNameProposal() const noexcept {
  int i = 1;
  while (mPadList->contains(QString::number(i))) {
    ++i;
  }
  return QString::number(i);
}
```

ubruhin: Seems that this is actually resolved :)

rnestler: I actually was a bit hesistant to solve it that way, because it may be `O(n²)`. But I tested it with 128 pads and it wasn't really noticeable.

rnestler:
@ubruhin Fixed the formatting issue.

ubruhin:
Thanks! :smiley:

PR #471, opened 2019-06-30
Fix possible crash caused by low-level signal/slots

ubruhin:
In some rare cases it was possible to crash the application because slots were detached from signals inside the callback of a slot, probably because following loop was not safe for modifying the `std::set` while iterating over it:

https://github.com/LibrePCB/LibrePCB/blob/7437d2898cb3453d86cdc246e04d9c0b8095da22/libs/librepcb/common/signalslot.h#L117-L119

Thus I changed the loop to this:

https://github.com/LibrePCB/LibrePCB/blob/c26c847961ea1745f7685b26cd56efd268b6c5c2/libs/librepcb/common/signalslot.h#L126-L128

@rnestler Do you know if this way it is safe to modify the container `mSlots` inside the callback? I'm not absolutely sure, but at least now I can't reproduce the crash I was experiencing before this chnge :wink:

### DIFF
--- a/libs/librepcb/common/signalslot.h
+++ b/libs/librepcb/common/signalslot.h
@@ -23,9 +23,9 @@
 /*******************************************************************************
  *  Includes
  ******************************************************************************/
+#include <QtCore>
+
 #include <functional>
-#include <set>
-#include <vector>
 
 /*******************************************************************************
  *  Namespace / Forward Declarations
@@ -84,7 +84,7 @@ public:
    */
   ~Signal() noexcept {
     for (auto slot : mSlots) {
-      slot->mSignals.erase(this);
+      slot->mSignals.remove(this);
     }
   }
 
@@ -104,8 +104,8 @@ public:
    * @param slot  Reference to the slot to detach
    */
   void detach(Slot<Tsender, Args...>& slot) const noexcept {
-    slot.mSignals.erase(this);
-    mSlots.erase(&slot);
+    slot.mSignals.remove(this);
+    mSlots.remove(&slot);
   }
 
   /**
@@ -114,8 +114,18 @@ public:
    * @param args  Arguments passed to the slots
    */
   void notify(Args... args) noexcept {
-    for (auto slot : mSlots) {
-      slot->mCallback(mSender, args...);
+    // Note: A "foreach" loop with a Qt container first creates an implicitly
+    // shared copy of the container (see
+    // https://doc.qt.io/qt-5/containers.html#foreach). This is very important
+    // since the callback might modify the container while iterating over it!
+    // With std containers this would be much more complicated (or less
+    // efficient).
+    foreach (const auto& slot, mSlots) {
+      // Check existence of the slot again because we must not call it if it
+      // was detached (i.e. removed from the container) in the meantime.
+      if (mSlots.contains(slot)) {
+        slot->mCallback(mSender, args...);
+      }
     }
   }
 
@@ -124,7 +134,7 @@ public:
 
 private:
   const Tsender& mSender;  ///< Reference to the sender object
-  mutable std::set<Slot<Tsender, Args...>*> mSlots;  ///< All attached slots
+  mutable QSet<Slot<Tsender, Args...>*> mSlots;  ///< All attached slots
 };
 
 /*******************************************************************************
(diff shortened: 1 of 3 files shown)


rnestler:
> Do you know if this way it is safe to modify the container mSlots inside the callback? 

From https://en.cppreference.com/w/cpp/container/set/erase
> References and iterators to the erased elements are invalidated. Other references and iterators are not affected. 

That means if a callback removes itself it will lead to undefined behavior since the iterator is invalidated. Also with the new loop construct. All the other iterators remain valid. That is also the reason why the following works:
```
    for(auto it = c.begin(); it != c.end(); )
        if(*it % 2 == 1)
            it = c.erase(it); // erase returns the next position
        else
            ++it;
```

The safest bet would probably be to clone the set and then iterate on the clone, but I'm not sure what the implications of that would be (how expensive / complicated).

Another approach if callback can only remove themself would be the following:
```
 for (auto it = mSlots.begin(); it != mSlots.end();) { 
  auto next = it+1;
   (*it)->mCallback(mSender, args...); 
   it = next;
}
```
So we would increment the iterator while it is still valid. But that means that callbacks are not allowed to remove other callbacks!

rnestler:
Another idea would be for the slot to return if they want to get removed:
```cpp

 for (auto it = mSlots.begin(); it != mSlots.end();) { 
   if ((*(it++))->mCallback(mSender, args...) ) {
     it = mSlots.erase(it);
   } else  {
     ++it;
   }
} 
```

ubruhin:
> That means if a callback removes itself it will lead to undefined behavior since the iterator is invalidated. Also with the new loop construct. 

That's why I wrote `(*(it++))->mCallback()` which should first increment the iterator, and then call the callback on the *old* iterator (if it works as intended) :)

But in that case the callback must not remove the *next* item in the container :(

> The safest bet would probably be to clone the set and then iterate on the clone, but I'm not sure what the implications of that would be (how expensive / complicated).

Yeah so probably that's the best option, I think I'll implement it that way (no matter how expensive it is).

Thanks!

rnestler:
> That's why I wrote (*(it++))->mCallback() which should first increment the iterator, and then call the callback on the old iterator (if it works as intended) :)

Ah right of course by using post increment you already do that.

ubruhin:
I had another idea: When using Qt containers instead of std containers, it's much easier :slightly_smiling_face: Since Qt containers are implicitly shared (with copy-on-write), creating a copy of the container is lightweight as long as the container does not get modified in the callback. In addition, Qt's `foreach` loop is safe for modifying the container inside the loop since it gets copied automatically.

libs/librepcb/common/signalslot.h
```diff
@@ -114,8 +121,16 @@ class Signal {
    * @param args  Arguments passed to the slots
    */
   void notify(Args... args) noexcept {
-    for (auto slot : mSlots) {
-      slot->mCallback(mSender, args...);
+    // Note: A "foreach" loop with a Qt container first creates an implicitly
+    // shared copy of the container. This is very important since the callback
+    // might modify the container while iterating over it! With std containers
+    // this would be much more complicated (or less efficient).
```
rnestler: Maybe link to https://doc.qt.io/qt-5/containers.html#foreach for interested people :slightly_smiling_face:

tests/unittests/common/signalslottest.cpp
```diff
@@ -0,0 +1,104 @@
...
+  EXPECT_EQ(101, sender.signal.getSlotCount());
+  foreach (const auto& receiver, receivers) {
+    EXPECT_CALL(*receiver, callback(testing::_, testing::_)).Times(0);
+  }
+}
+
+TEST(SignalSlotTest, testDuringCallbackDetachedSlotsAreNotCalled) {
+  int                                       callbackCounter = 0;
+  Sender                                    sender;
+  QList<std::shared_ptr<Slot<Sender, int>>> receivers;
```
rnestler: I think you are missing an include for `std::shared_ptr`.

libs/librepcb/common/signalslot.h
ubruhin: Done :)

tests/unittests/common/signalslottest.cpp
ubruhin: Yep -> fixed.

Issue #474, opened 2019-07-03
Scrolling with Mac trackpad zooms, but should scroll


jeffwheeler:
##### SUMMARY

Presently, the `GraphicsView` in the schematic and layout editor is difficult to use with a Mac trackpad. When swiping north with two fingers, the widget zooms out, and when swiping south with two fingers, the widget zooms in. Pinching does not work as it normally does on Mac OS.

Expected behavior can be seen in most apps, such as Preview.app.

##### SOLUTION

Standard Mac OS behavior is to scroll in two dimensions using two fingers, and pinch to zoom in/out. Qt supports this with [`QNativeGestureEvent`](https://doc.qt.io/qt-5/qnativegestureevent.html), but after playing with it for a bit I couldn't get it to work quite right.

(Side note, I found [this blog post](https://blog.qt.io/blog/2017/01/19/should-you-be-using-qgraphicsview/) warning that `QGraphicsView` may be deprecated in the future.)

jeffwheeler:
I've posted [this commit](https://github.com/jeffwheeler/LibrePCB/commit/8573842955dc68f7dea7d0b4f47a532ab2d9a64e) which demonstrates using `QNativeGestureEvent`. It works, but I expect it breaks other systems (and Macs with mice) because I disable the scrollwheel handling.

ubruhin:
Thanks for the code example! Looks like it shouldn't be too hard to integrate that code snippet properly. @jeffwheeler would you be interested to create a pull request for that?

PR #477, opened 2019-07-04
GraphicsView: Implement pinch to zoom

jeffwheeler:
This pull request is addresses #474.

The code in the initial pull request should not be merged, but is an initial example. I'd like to figure out how to make it work across platforms and with a mouse.

### DIFF
--- a/libs/librepcb/common/graphics/graphicsview.cpp
+++ b/libs/librepcb/common/graphics/graphicsview.cpp
@@ -212,9 +212,7 @@ bool GraphicsView::event(QEvent* event) {
             return true;
           }
           case Qt::ZoomNativeGesture: {
-            QTransform transform;
-            transform.scale(1+gesture->value(), 1+gesture->value());
-            setTransform(transform, true);
+            scale(1+gesture->value(), 1+gesture->value());
             return true;
           }
           default: {
--- a/libs/librepcb/common/graphics/graphicsview.cpp
+++ b/libs/librepcb/common/graphics/graphicsview.cpp
@@ -29,7 +29,6 @@
 
 #include <QtCore>
 #include <QtWidgets>
-#include <QtDebug>
 
 /*******************************************************************************
  *  Namespace
@@ -278,7 +277,7 @@ bool GraphicsView::eventFilter(QObject* obj, QEvent* event) {
       emit cursorScenePositionChanged(Point::fromPx(e->scenePos()));
       // no break here!
     }
-    [[clang::fallthrough]]; case QEvent::GraphicsSceneMouseDoubleClick:
+    case QEvent::GraphicsSceneMouseDoubleClick:
     case QEvent::GraphicsSceneContextMenu: {
       if (mEventHandlerObject) {
         mEventHandlerObject->graphicsViewEventHandler(event);
--- a/libs/librepcb/common/graphics/graphicsview.cpp
+++ b/libs/librepcb/common/graphics/graphicsview.cpp
@@ -215,6 +215,10 @@ void GraphicsView::wheelEvent(QWheelEvent* event) {
     QGraphicsView::wheelEvent(event);
   }
 }
+#else
+void GraphicsView::wheelEvent(QWheelEvent* event) {
+  QGraphicsView::wheelEvent(event);
+}
 #endif
 
 bool GraphicsView::eventFilter(QObject* obj, QEvent* event) {
--- a/libs/librepcb/common/graphics/graphicsview.cpp
+++ b/libs/librepcb/common/graphics/graphicsview.cpp
@@ -60,6 +60,8 @@ GraphicsView::GraphicsView(QWidget*                     parent,
   mZoomAnimation = new QVariantAnimation();
   connect(mZoomAnimation, &QVariantAnimation::valueChanged, this,
           &GraphicsView::zoomAnimationValueChanged);
+
+  viewport()->grabGesture(Qt::PinchGesture);
 }
 
 GraphicsView::~GraphicsView() noexcept {
--- a/libs/librepcb/common/graphics/graphicsview.cpp
+++ b/libs/librepcb/common/graphics/graphicsview.cpp
@@ -29,6 +29,7 @@
 
 #include <QtCore>
 #include <QtWidgets>
+#include <QtDebug>
 
 /*******************************************************************************
  *  Namespace
@@ -140,6 +141,7 @@ Point GraphicsView::mapGlobalPosToScenePos(const QPoint& globalPosPx,
 
 void GraphicsView::handleMouseWheelEvent(
     QGraphicsSceneWheelEvent* event) noexcept {
+  /*
   if (event->modifiers().testFlag(Qt::ShiftModifier)) {
     // horizontal scrolling
     horizontalScrollBar()->setValue(horizontalScrollBar()->value() -
@@ -160,6 +162,7 @@ void GraphicsView::handleMouseWheelEvent(
     scale(scaleFactor, scaleFactor);
   }
   event->setAccepted(true);
+  */
 }
 
 /*******************************************************************************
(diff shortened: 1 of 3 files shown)


jeffwheeler:
I have updated the pull request with a set of commits that I think should behave nicely for everyone.

I don't believe we need a preference setting to configure this. My intent is to leave mice behavior unchanged (including on Mac), but fix the currently broken trackpad behavior. I don't think anyone would want the existing trackpad behavior.

The commit fails CI on Ubuntu 14.04, I think because of an ancient version of Qt. Is that something you're interested in supporting? Canonical has dropped support for that platform already.

libs/librepcb/common/graphics/graphicsview.cpp
```diff
@@ -139,12 +139,12 @@ Point GraphicsView::mapGlobalPosToScenePos(const QPoint& globalPosPx,
 }
 
 void GraphicsView::handleMouseWheelEvent(
-    QGraphicsSceneWheelEvent* event) noexcept {
-  if (event->modifiers().testFlag(Qt::ShiftModifier)) {
+    QWheelEvent* event) noexcept {
+  if (event->buttons() & Qt::ShiftModifier) {
```
ubruhin: This is not allowed since `event->buttons()` has the return type `Qt::MouseButtons`, while `Qt::ShiftModifier` is of type `Qt::KeyboardModifier`. Same issue a few lines below. These changes break horizontal and vertical scrolling.

And why was the type of the parameter `event` changed from `QGraphicsSceneWheelEvent*` to `QWheelEvent?`?

libs/librepcb/common/graphics/graphicsview.cpp
```diff
@@ -203,6 +203,39 @@ void GraphicsView::zoomAnimationValueChanged(const QVariant& value) noexcept {
  *  Inherited from QGraphicsView
  ******************************************************************************/
 
+bool GraphicsView::event(QEvent* event) {
```
ubruhin: Wouldn't it be possible to handle `QNativeGestureEvent` in the `eventFilter()` method where we already handle every other event? The event handling of `GraphicsView` is special since we don't handle our own events, but those of the underlying `GraphicsScene` object. This way we also get scene events (with scene coordinates) in addition to every other event. So if possible, I would like to handle `QNativeGestureEvent` in the `eventFilter()` method too (to reduce complexity and for consistency).

libs/librepcb/common/graphics/graphicsview.cpp
```diff
@@ -203,6 +203,39 @@ void GraphicsView::zoomAnimationValueChanged(const QVariant& value) noexcept {
...
+        QNativeGestureEvent* gesture = dynamic_cast<QNativeGestureEvent*>(event);
+        switch (gesture->gestureType()) {
+          case Qt::BeginNativeGesture:
+          case Qt::EndNativeGesture: {
+            return true;
+          }
+          case Qt::ZoomNativeGesture: {
+            QTransform transform;
+            transform.scale(1+gesture->value(), 1+gesture->value());
+            setTransform(transform, true);
```
ubruhin: Why not just call `scale()` instead of doing the transformation stuff manually?

libs/librepcb/common/graphics/graphicsview.cpp
```diff
@@ -203,6 +203,39 @@ void GraphicsView::zoomAnimationValueChanged(const QVariant& value) noexcept {
...
+            return false;
+          }
+        }
+    }
+
+    return false;
+}
+
+// This function receives the wheel events for the viewport widget.
+void GraphicsView::wheelEvent(QWheelEvent* event) {
```
ubruhin: Same here: Couldn't `QWheelEvent` be handled in `eventFilter()` instead?

libs/librepcb/common/graphics/graphicsview.cpp
```diff
@@ -203,6 +203,39 @@ void GraphicsView::zoomAnimationValueChanged(const QVariant& value) noexcept {
...
+  if (event->source() == Qt::MouseEventSynthesizedBySystem) {
```
ubruhin: The `source()` method is not available in Qt < 5.5, but this can easily be fixed by adding a #if/endif:
```cpp
#if (QT_VERSION >= QT_VERSION_CHECK(5, 5, 0))
  // Qt >= 5.5
#else
  // Qt < 5.5
#endif
```

jeffwheeler:
Thanks for the feedback, I’ll review in a few days. :)

rnestler:
> For example Debian oldstable still has Qt 5.3.2, and I don't like to drop support for a still supported distribution if not absolutely necessary.

So we could bump the minimal required Qt version from 5.2 to 5.3

libs/librepcb/common/graphics/graphicsview.cpp
jeffwheeler: I've removed the main `event()` function, but I need to keep `wheelEvent()`. The reason is that when the `QWheelEvent` is translated into a `QGraphicsSceneWheelEvent`, it loses the source information, and so I cannot test for whether it's a synthetic event and reject it. Instead, I need to do it here and send it up to the `QAbstractScrollArea` if it's a trackpad event. Otherwise, it'll pass to the `eventFilter()`.

libs/librepcb/common/graphics/graphicsview.cpp
jeffwheeler: Done.

libs/librepcb/common/graphics/graphicsview.cpp
jeffwheeler: I've removed this part of the diff. It's now back to the original design.

libs/librepcb/common/graphics/graphicsview.cpp
jeffwheeler: It is possible. It originally wasn't sending the event because my `grabGesture()` was getting lost after `setUseOpenGl()` reset the viewport. Now that I figured that out, I receive the gesture events in the `eventFilter()`.

libs/librepcb/common/graphics/graphicsview.cpp
jeffwheeler: Done.

jeffwheeler:
Hi @ubruhin, I think your comments should now be addressed. Behavior should be unchanged for old versions of Qt.

ubruhin:
Thanks! Looks like you no longer use `QNativeGestureEvent` :)

Do I understand correctly that with this PR "pinch to zoom" works out-of-the-box on a MacBook?

I think there is still a small issue with the implementation: The method `grabGesture()` is called in `setUseOpenGl()`, but not in the constructor of `GraphicsView`. So by default, gestures are not grabbed and thus (probably) gestures won't work. But actually it still works "by accident" because `setUseOpenGl()` is called by users of `GraphicsView` anyway. If you want to fix this - perfect. If not - don't worry, I'd still merge this since it's not critical :slightly_smiling_face:

jeffwheeler:
> Thanks! Looks like you no longer use `QNativeGestureEvent` :)

Yup. I discovered that `QGestureManager::QGestureManager` has an `#if defined(Q_OS_OSX)`, and if so, registers the native Mac gesture recognizers (such as `QMacPinchGestureRecognizer`).

> Do I understand correctly that with this PR "pinch to zoom" works out-of-the-box on a MacBook?

Yup. :)

> I think there is still a small issue with the implementation: The method `grabGesture()` is called in `setUseOpenGl()`, but not in the constructor of `GraphicsView`. So by default, gestures are not grabbed and thus (probably) gestures won't work. But actually it still works "by accident" because `setUseOpenGl()` is called by users of `GraphicsView` anyway. If you want to fix this - perfect. If not - don't worry, I'd still merge this since it's not critical 🙂

I'll push a fix for this momentarily.

ubruhin:
Perfect, thank you very much for this improvement! :smiley:

PR #478, opened 2019-07-07
Fix deprecation warnings when compiling with Qt 5.13

ubruhin:
CI on MacOS is currently broken because homebrew always installs the latest version of Qt, which is now Qt 5.13 and throws some deprecation warnings. This PR fixes the deprecation warnings to make CI successful again:

- Replace `qSort()` by `std::sort()`
- Replace `QListWidgetItem::setBackgroundColor()` by `QListWidgetItem::setBackground()`

### DIFF
--- a/apps/librepcb-cli/commandlineinterface.cpp
+++ b/apps/librepcb-cli/commandlineinterface.cpp
@@ -37,6 +37,8 @@
 
 #include <QtCore>
 
+#include <algorithm>
+
 /*******************************************************************************
  *  Namespace
  ******************************************************************************/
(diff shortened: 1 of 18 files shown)


rnestler:
> CI on MacOS is currently broken because homebrew always installs the latest version of Qt.

Should we change that to get more predictable CI? On the other hand it is nice to also test with the latest Qt version.

ubruhin:
> Should we change that to get more predictable CI? On the other hand it is nice to also test with the latest Qt version.

I would like to change it but didn't find a way to install a specific version of Qt.

PR #480, opened 2019-07-08
CI: Build on Debian 8 instead of Ubuntu 14.04

rnestler:


ubruhin:
Why removing the 14.04 job, what's the advantage of this? Will you update Doxygen documentation + README.md, remove Qt5.2 compatibility code and write the blog post that/why we drop Qt 5.2 support? Note: Even if you would do these things, I still don't see why we should drop Qt5.2 support without any reason. Maintaining Qt5.2 support requires much less effort than dropping the support. Even communicating with you all the time about this topic needs more time than supporting Qt5.2 for the next 5 years would consume :wink:

dbrgn:
I don't care either way, but over the next months to years slowly things will start breaking. The Ubuntu 14.04 repositories will be turned off, meaning that installing packages in the Docker container will fail. People contributing patches will notice that something doesn't work because it's not compatible with Qt 5.2 (as it has already happened) and will need some guidance on what to change to restore compatibility with a system that nobody is using anymore.

I don't think a blogpost is required as nobody will be relying on Qt 5.2 support with LibrePCB. I fully understand strict backwards compatibility policies and generally like them, but LibrePCB is a desktop application, not a library. If someone wants to use it on a distribution that provides neither bugfixes nor security patches, then those people can use the appimage.

The only advantage I see in not dropping 5.2 right now (especially if @rnestler does all the work) is that we could drop 5.2 and 5.3 at the same time some time in the future. But in the meantime, a few contributors will probably run into backwards compatibility issues and lose some time.

My suggestion would be to let @rnestler go wild and do all the work for you :wink:

ubruhin:
I understand your arguments @dbrgn. But the thing is that supporting Qt5.2 is still *very* easy, easier than actively dropping support (most of the time you don't have to care about 5.2 support, it just works out of the box).

> I don't care either way, but over the next months to years slowly things will start breaking. The Ubuntu 14.04 repositories will be turned off, meaning that installing packages in the Docker container will fail. 

If this happens, of course I wouldn't actively try to still keep support for Qt 5.2. But it didn't happen yet, everything still works fine.

> People contributing patches will notice that something doesn't work because it's not compatible with Qt 5.2 (as it has already happened)

But that would need to drop up to Qt5.4, which is still used in some LTS distributions. So dropping Qt5.2 support wouldn't help in this case. In addition, if someone uses a feature from Qt5.13, of course we don't drop support for Qt<=5.12 just because one #ifdef/endif is missing.

> If someone wants to use it on a distribution that provides neither bugfixes nor security patches, then those people can use the appimage.

But then you can't easily compile LibrePCB on such a system, which is bad for contributors. For example if we drop Qt5.2 support now, all Ubuntu 16.04 users can no longer compile LibrePCB without manually installing a more recent Qt version.

My philosophy of OpenSource is not just open-sourcing the code, but to provide code which works for everyone. Support for Windows is a similar topic - many OpenSource developers don't care about Windows support since they use Linux. I also don't like and don't use Windows (or MacOS), but IMHO that's just selfish.

> The only advantage I see in not dropping 5.2 right now (especially if @rnestler does all the work) is that we could drop 5.2 and 5.3 at the same time some time in the future.

That's one reason why I don't like to drop Qt5.2 while still keeping Qt5.3 - it's just wasted time. Better wait some more time until we release a new major version and then drop support for Qt<=5.4 in one step, that would be much more efficient.

Once again - this discussion already took a multiple of the time needed to keep support for Qt5.2. The effort of @rnestler and me together is for sure >>1h in the last year. Supporting Qt5.2 during the last year was a matter of a few minutes.

ubruhin:
I have an idea @rnestler: If you also adjust doxygen documentation and the README, **and promise to don't give any comment about support of old Qt versions for at least one year** (to save my time and nerves), I'll accept your pull request now :smiley:

rnestler:
> For example if we drop Qt5.2 support now, all Ubuntu 16.04 users can no longer compile LibrePCB without manually installing a more recent Qt version.

You mean Ubuntu 14.04 here.

dbrgn:
> For example if we drop Qt5.2 support now, all Ubuntu 16.04 users can no longer compile LibrePCB without manually installing a more recent Qt version.

If that's true, then I'd also be in favor of keeping Qt 5.2 support. But are you really sure that 14.04 and 16.04 have the same Qt version installed?!

rnestler:
> But are you really sure that 14.04 and 16.04 have the same Qt version installed?!

No. 16.04 has Qt 5.5.

ubruhin:
>> For example if we drop Qt5.2 support now, all Ubuntu 16.04 users can no longer compile LibrePCB without manually installing a more recent Qt version.
>
> You mean Ubuntu 14.04 here. Which is EOL.

Oops, I mean "if we drop Qt5.**5**", sorry :) (thus starting the sentence with "For example..."). But same applies for Qt5.3 on Debian Jessie.

rnestler:
> But same applies for Qt5.3 on Debian Jessie.

Well technically Debian Jessie is also EOL since mid 2018 :wink:

ubruhin:
Can we now start focusing on issues which actually improve LibrePCB and provide advantages for users, instead of endless discussing about totally irrelevant things? For example, is the push&shove router now available as a standalone library @rnestler? :wink:

dbrgn:
@rnestler what's the state on this?

rnestler:
> @rnestler what's the state on this?

Well I lost interest since @ubruhin does seem to care more about supporting Ubuntu 14.04 than Debian 8.

PR #486, opened 2019-07-21
Remove invalid 'modal' property that causes warning

jeffwheeler:
This property in `librarymanager.ui` is invalid. Modality should be set with the `windowModality` property, but it defaults to `Qt::NonModal`.

The removed lines caused this warning on startup:
`[ WARNING ] librepcb::library::manager::LibraryManager::setProperty: Property "modal" invalid, read-only or does not exist (kernel/qobject.cpp:3953)`


ubruhin:
Makes sense, thanks! :+1:

PR #487, opened 2019-07-21
Add toggle in the menu bar for dock widgets

jeffwheeler:
Prior to this commit, it is impossible to restore dock widgets that were accidentally closed, except by deleting the QSettings storage.

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
@@ -120,6 +120,12 @@ BoardEditor::BoardEditor(ProjectEditor& projectEditor, Project& project)
   // setCentralWidget(mGraphicsView);
   mUi->centralwidget->layout()->addWidget(mGraphicsView);
 
+  // Add actions to toggle visibility of dock widgets
+  mUi->menuView->addSeparator();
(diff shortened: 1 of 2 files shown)


jeffwheeler:
I built this because of the bug solved in #488, but it remains relevant after that bug is fixed.

ubruhin:
Thanks for the PR, but actually the dock widgets can already be restored with a right click on a toolbar or menubar :wink:

grafik

I think in most software this is the way to toggle visibility of toolbars and docks. Or isn't this working for you?

jeffwheeler:
The bug described in #488 unfortunately removes all toolbars and menu bars, so there's nowhere to right click.

image

I'm not sure about Linux (Ubuntu?), but on Mac, there's typically menus for this. It makes discovery much easier.

Screen Shot 2019-07-23 at 7 21 00 PM

ubruhin:
Hmm okay, very strange :(

So then I agree to add these menu items. But actually I would expect them in the "View" menu, not in "Tools" (tools is virtually coupled to the corresponding toolbar on the left side, it's not meant to contain other menu items). Could you update the PR accordingly?

jeffwheeler:
Done.

ubruhin:
Perfect, thanks! :smiley:

Issue #49, opened 2016-01-25
Use Ctrl + W to close a Project


dns2utf8:
Use Alt + F4 to close a Window

rnestler:
I wouldn't expect Ctrl+W to close a Project, but to close a tab.

dbrgn:
> Use Alt + F4 to close a Window

This should be handled by your window manager actually.

Is there a specification for this type of shortcuts? Maybe at freedesktop.org?

Issue #48, opened 2016-01-25
Make Ctrl + Q close the whole Application not just the window


dns2utf8:

spookydd:
Ctrl+Q would be Cmd-Q on macos

PR #488, opened 2019-07-21
Fix state saving to occur when window is closed

jeffwheeler:
Presently, there's a bug where I close a schematic editor or board editor, and the state is not saved. Later, the state is saved in the destructor called with `delete` in `~ProjectEditor`.

Because the window is already closed, all the widgets are gone in the saved state. When I reopen the window, I do not have any menus or dock widgets.

**Note** I'd also like to address #48 and #49 in this pull request. I haven't done that yet.

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
@@ -249,11 +249,6 @@ BoardEditor::BoardEditor(ProjectEditor& projectEditor, Project& project)
 }
 
 BoardEditor::~BoardEditor() {
-  // Save Window Geometry
-  QSettings clientSettings;
-  clientSettings.setValue("board_editor/window_geometry", saveGeometry());
-  clientSettings.setValue("board_editor/window_state", saveState());
-
   delete mFsm;
   mFsm = nullptr;
   qDeleteAll(mBoardListActions);
@@ -330,8 +325,14 @@ void BoardEditor::abortAllCommands() noexcept {
 void BoardEditor::closeEvent(QCloseEvent* event) {
   if (!mProjectEditor.windowIsAboutToClose(*this))
     event->ignore();
-  else
+  else {
+    // Save window geometry
+    QSettings clientSettings;
+    clientSettings.setValue("board_editor/window_geometry", saveGeometry());
+    clientSettings.setValue("board_editor/window_state", saveState());
+
     QMainWindow::closeEvent(event);
+  }
 }
 
 /*******************************************************************************
--- a/libs/librepcb/projecteditor/schematiceditor/schematiceditor.cpp
+++ b/libs/librepcb/projecteditor/schematiceditor/schematiceditor.cpp
@@ -216,11 +216,6 @@ SchematicEditor::SchematicEditor(ProjectEditor& projectEditor, Project& project)
 }
 
 SchematicEditor::~SchematicEditor() {
-  // Save Window Geometry
-  QSettings clientSettings;
-  clientSettings.setValue("schematic_editor/window_geometry", saveGeometry());
-  clientSettings.setValue("schematic_editor/window_state", saveState());
-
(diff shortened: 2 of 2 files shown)


ubruhin:
Hmm I can't reproduce the behavior you describe. Which Qt version are you using?

> Because the window is already closed, all the widgets are gone in the saved state.

AFAIK, closing a window basically only means to set its `visible` attribute to `false`. There aren't any widgets "gone", the whole window with all widgets stays there as-is, it's just hidden (you can even re-open it with a tool button in the other editor window). So I don't understand why this should affect saving the window state...

jeffwheeler:
I'm using Qt 5.13.0. This only occurs for me if I close using the red close button on the Mac's menu. If I use Cmd+Q, the state is saved correctly. I wasn't able to figure out exactly why there's a difference, though.

The Qt documentation recommends saving state in the `closeEvent` handler ([documentation link](https://doc.qt.io/qt-5/qmainwindow.html#saveState)).

ubruhin:
OK, I see. But the problem with `closeEvent()` is that it's not called when closing board/schematic editors from within `ProjectEditor`. For example when closing from the menu item "File" -> "Close Project", the state isn't saved at all. I'm not sure what would be a good solution to fix both issues...

ubruhin:
Btw, I was now able to reproduce this bug with Qt 5.12.3. With Qt 5.5 it works properly, so it seems Qt has changed this behavior some time :(

To avoid happening such a bug again in future, I wrote a system test on the branch `client-settings-functional-tests`. Let's see if CI confirms my expectation that the test fails only with recent Qt versions...

ubruhin:
This is a really weird bug, it seems to happen only if the viewport of the graphics view is set to `QGLWidget`, i.e. OpenGL is enabled in workspace settings. @jeffwheeler could you check if the bug is gone after disabling OpenGL?

jeffwheeler:
@ubruhin Yup, you got it exactly! Only the case with OpenGL.

I tried replacing `QGLWidget` ([long deprecated](https://doc.qt.io/qt-5/qglwidget.html)) with `QOpenGLWidget` and it seemed to fix the problem.

ubruhin:
Ah, indeed! Would you update this PR accordingly or should I take care of switching to `QOpenGLWidget`?

jeffwheeler:
@ubruhin Can you do it? Thanks. I'm not sure how to reimplement the format flags (I suspect they're not needed at all). Here's the diff I made to test it, though.

```
diff --git a/libs/librepcb/common/graphics/graphicsview.cpp b/libs/librepcb/common/graphics/graphicsview.cpp
index cb0be5bb6..d5cc947b1 100644
--- a/libs/librepcb/common/graphics/graphicsview.cpp
+++ b/libs/librepcb/common/graphics/graphicsview.cpp
@@ -86,8 +86,7 @@ QRectF GraphicsView::getVisibleSceneRect() const noexcept {
 void GraphicsView::setUseOpenGl(bool useOpenGl) noexcept {
   if (useOpenGl != mUseOpenGl) {
     if (useOpenGl)
-      setViewport(new QGLWidget(QGLFormat(
-          QGL::DoubleBuffer | QGL::AlphaChannel | QGL::SampleBuffers)));
+      setViewport(new QOpenGLWidget());
     else
       setViewport(nullptr);
     mUseOpenGl = useOpenGl;
```

ubruhin:
Ok, I opened #503 for switching to `QOpenGLWidget`.

> I'm not sure how to reimplement the format flags (I suspect they're not needed at all). Here's the diff I made to test it, though.

To be honest, I have no experience with the OpenGL stuff but with the default settings of Qt, schematics and boards look very ugly or even wrong, depending on zoom level etc. Years ago I just played around with the settings to fix these issues (at least on my machine). Since with `QOpenGLWidget` I still have the same issues, so I just set similar options again :wink:

Closing this PR since it is replaced by #503. Thanks a lot for your help @jeffwheeler! :smiley:

PR #489, opened 2019-07-28
Polygon editors: Support copying and moving rows

### LINKED ISSUE #483: Improve Polygon Editor
##### SUMMARY

The polygon editor is currently very limited. The biggest problem ist that rows can't be moved.

For example, here:

img

...it's not easily possible to add more points between row 1 and 2.

##### SOLUTION

- Allow moving rows around
- Allow inserting rows below a selected other row
- Maybe allow copying rows

ubruhin:
Adding buttons for copying and moving rows in the polygon editors:

grafik

In addition, cells now show the unit of the contained value (`mm` and `°`).

Fixes #483


dbrgn:
Cool, thanks!

PR #499, opened 2019-08-13
Add simple BOM export dialog to schematic/board editors

ubruhin:
Although the library does not yet really support part handling (with MPN etc.), it's still possible (and useful) to generate simple BOMs just by iterating over all devices on a board. And if needed, MPNs can be added to the components (in schematics) by using attributes.

grafik

The CSV output then looks like that:

```csv
Quantity;Designators;Value;Device;Package;MPN
4;C1, C2, C3, C4;LGW2E471MELZ40 470μF/250V;Elko 22/10mm;Elko 22/10mm;LGW2E471MELZ40
2;C10, C11;MKS4F044706B00KSSD 4.7μF/250V;MKS4F044706B00KSSD;Capacitor 11x21x31.5 PCM 27.5 Red;MKS4F044706B00KSSD
1;C12;2.2nF;C-0603;C-0603;
3;C13, C14, C22;1μF;C-0603;C-0603;
1;C15;2.2μF 100V;C-1210;C-1210;
1;C16;56pF;C-0603;C-0603;
5;C17, C18, C7, C8, C9;220nF;C-0603;C-0603;
6;C19, C23, C28, C31, C32, C33;100nF;C-0603;C-0603;
1;C20;10μF;C-0805;C-0805;
2;C21, C5;10nF;C-0603;C-0603;
2;C24, C25;47nF 250V;C-1812;C-1812;
1;C26;22μF 25V;C-1210;C-1210;
1;C27;3.3nF;C-0603;C-0603;
1;C29;dnm;C-0805;C-0805;
2;C30, C6;2.2μF 25V;C-0805;C-0805;
1;D1;MBR40250G;MBR40250G;TO220-2;MBR40250G
```


dbrgn:
Cool!

Does this list all attributes or only "MPN"? I often add a "Digikey" key with the order number to the attributes, would be nice if such fields could be included in the BOM too.

ubruhin:
> Does this list all attributes or only "MPN"?

Currently only "MPN".

> I often add a "Digikey" key with the order number to the attributes, would be nice if such fields could be included in the BOM too.

Good point, I'll check if I can make the columns configurable.

EdizonTN:
Hi, 
don't forgot for implement/solve technology parts, which does not need list in BOM.
For example, logo, fiducial marks, ESD mark, Sheet Frames.....

dbrgn:
Maybe something like this in the symbol property? (Ugly Mockup)

image

EdizonTN:
> Maybe something like this in the symbol property? (Ugly Mockup)
**Yeah! This is enough for simple BOM.** 
For future we need little bit more complicated BOM setup and field mapping (MPN, Datasheet, Some user attributes,.. ) depends on PCB variant, or assembly variant (one PCB/schema with more combination of assembled/not assembled devices ..... maybe implemented as new feature ;-) )...

ubruhin:
> Good point, I'll check if I can make the columns configurable.

Done - instead of hardcoding the "MPN" column, arbitrary attributes can now be added as columns:

grafik

In addition, I added commands to the CLI to export BOMs so this can now be automated too.

> don't forgot for implement/solve technology parts, which does not needs listed in BOM.
For example, logo, fiducial marks, ESD mark, Sheet Frames.....

Components which are marked as "Schematic only" in the library aren't included in the BOM since those can't be placed on a board anyway. This should already catch most of the no-BOM components. Everything more than this would introduce a file format change, and I think it makes more sense to wait until we have a clear concept for handling MPNs, datasheets, assembly variants etc.

Of course this is only a very simple BOM export, but since this implementation doesn't affect the file format, it's a good first step until we implement a much more feature-rich BOM handling :wink:

PR #501, opened 2019-08-20
Fix missing plane cut-out for board outlines in footprints

### LINKED ISSUE #500: Package board outlines not subtracted from copper plane

```
LibrePCB Version: 0.1.2-unstable
Git Revision:     b73e8c97606da76d302b869a8de0c4b7231bc7ba
Build Date:       2019-08-11T10:58:10
Qt Version:       5.13.0 (built against 5.13.0)
CPU Architecture: x86_64
Operating System: Arch Linux
Platform Plugin:  xcb
```

##### SUMMARY

I have some packages with closed polygons on the board outline layer.

As you can see, they are not subtracted from the copper plane.

screenshot

I added a small inner polygon on the board itself, and this is properly excluded from the plane, so it's probably due to the fact that the outline is drawn in the external package.

##### STEPS TO REPRODUCE

Clone https://github.com/jbrgn/flashlight/ and check out commit `e64eca6fbabbc6bedec36619efd13dea9c14f048`.

ubruhin:
Fixes #500:

grafik

### DIFF
--- a/libs/librepcb/project/boards/boardplanefragmentsbuilder.cpp
+++ b/libs/librepcb/project/boards/boardplanefragmentsbuilder.cpp
@@ -100,6 +100,20 @@ void BoardPlaneFragmentsBuilder::clipToBoardOutline() {
       boardAreaClipper.AddPath(path, ClipperLib::ptSubject, true);
     }
   }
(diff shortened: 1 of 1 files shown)


dbrgn:
Wohoo! Thanks.

PR #504, opened 2019-08-23
Gerber: Always use multi-quadrant mode

### LINKED ISSUE #247: Don't use single quadrant mode arcs in Gerber generator
gerbv:

screenshot

reference viewer:

screenshot2

oshpark:

screenshot3

gerbview (0 character only):

screenshot4

Here's the output:

```gerber
G04 --- HEADER BEGIN --- *
%TF.GenerationSoftware,LibrePCB,LibrePCB,0.1.0*%
%TF.CreationDate,2018-04-06T01:22:20*%
%TF.ProjectId,gerbertest - default,4fa26089-7cc2-42e7-9238-ecfcf38bbe1d,v1*%
%TF.Part,Single*%
%FSLAX66Y66*%
%MOMM*%
G01*
G74*
G04 --- HEADER END --- *
G04 --- APERTURE LIST BEGIN --- *
%ADD10C,1.35*%
G04 --- APERTURE LIST END --- *
G04 --- BOARD BEGIN --- *
D10*
X48721021Y66920000D02*
G75*
G03*
X46721021Y66920000I-1000000J-750821D01*
G01*
G74*
G03*
X46721021Y58920000I7491334J4000000D01*
G01*
G75*
G03*
X48721021Y58920000I1000000J750821D01*
G01*
G74*
G03*
X48721021Y66920000I7491334J4000000D01*
G01*
G04 --- BOARD END --- *
%TF.MD5,165e92c0a5fc09f49da7b3c7b7935a4f*%
M02*
```

Relevant parts (commented):

- `X48721021Y66920000D02` move to 48721021 / 66920000
- `G75` multi quadrant mode (arcs up to 360°)
- `G03` counterclockwise circular interpolation
- `X46721021Y66920000I-1000000J-750821D01` interpolate to 46.. / 66.. with X offset -1.. and Y offset -0.7..
- `G01` linear interpolation
- `G74` single quadrant mode (arcs up to 90°)
- `G03` counterclockwise circular interpolation
- `X46721021Y58920000I7491334J4000000D01` interpolate to 46.. / 58.. with X offset 0.7.. and Y offset 4..
- `G01` linear interpolation
- `G75` multi quadrant mode (arcs up to 360°)
- `G03` counterclockwise circular interpolation
- `X48721021Y58920000I1000000J750821D01` interpolate to 48.. / 58.. with X offset 1.. and Y offset 0.7..
- `G01` linear interpolation
- `G74` single quadrant mode (arcs up to 90°)
- `G03` counterclockwise circular interpolation
- `X48721021Y66920000I7491334J4000000D01` interpolate to 48.. / 66.. with X offset 0.7.. and Y offset 4..
- `G01` linear interpolation

Honestly I think this looks correct, and both gerbv and gerbview seem to be buggy with regards to the single quadrant mode circular interpolation.

Or could it be some kind of rounding issue?

Here's a rough sketch:

photo5922649731434130894

dbrgn:
@ubruhin here's a first attempt at fixing this issue.

The spec says:

> Another option is not to use very small arcs, e.g. by replacing them with
draws - the error is very small and draws are stable.

Right now I'm replacing arcs shorter than 0.01mm (point-to-point) with straight lines. 

We probably need to handle some special cases, right? A full circle has start==end with angle=360, right? We should probably still draw arcs if start and end are identical.

I guess we can ignore cases where the two points are closer together than 0.01mm with a lage angle, right?

Fixes #247.

### DIFF
--- a/libs/librepcb/common/cam/gerbergenerator.h
+++ b/libs/librepcb/common/cam/gerbergenerator.h
@@ -38,6 +38,7 @@ namespace librepcb {
 class Circle;
 class Path;
 class GerberApertureList;
+class Vertex;
 
 /*******************************************************************************
  *  Class GerberGenerator
@@ -111,6 +112,7 @@ private:
   void    linearInterpolateToPosition(const Point& pos) noexcept;
   void    circularInterpolateToPosition(const Point& start, const Point& center,
                                         const Point& end) noexcept;
+  void    interpolateBetween(const Vertex& from, const Vertex& to) noexcept;
   void    flashAtPosition(const Point& pos) noexcept;
   void    printHeader() noexcept;
   void    printApertureList() noexcept;
--- a/libs/librepcb/common/cam/gerbergenerator.cpp
+++ b/libs/librepcb/common/cam/gerbergenerator.cpp
@@ -108,11 +108,10 @@ void GerberGenerator::drawPathOutline(
       linearInterpolateToPosition(v.getPos());
     } else {
       // arc segment
-      if (v0.getAngle().abs() <= Angle::deg90()) {
-        setMultiQuadrantArcModeOff();
-      } else {
-        setMultiQuadrantArcModeOn();
-      }
+      // note: due to buggy clients when using single quadrant mode,
+      // we always use multi quadrant mode.
+      // see https://github.com/LibrePCB/LibrePCB/issues/247
+      setMultiQuadrantArcModeOn();
       if (v0.getAngle() < 0) {
         switchToCircularCwInterpolationModeG02();
       } else {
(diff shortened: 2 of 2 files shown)


libs/librepcb/common/cam/gerbergenerator.h
```diff
@@ -27,6 +27,7 @@
 #include "../fileio/filepath.h"
 #include "../units/all_length_units.h"
 #include "../uuid.h"
+#include "../geometry/vertex.h"
```
ubruhin: If only references or pointers to `Vertex` are needed, please replace the include by a forward declaration, just like it was done for the `Path` class. This heavily reduces build time if header files are modified.

dbrgn:
> Maybe it would indeed be reasonable to simply ignore the numerical instability. The risk of doing something wrong when handling special cases might be higher than the risk of really getting wrong arcs in the Gerber output

I agree and removed the special case handling. We're already using FSLAX66 format.

Other issues should be handled now.

dbrgn:
@ubruhin ok, rebased & squashed!

PR #509, opened 2019-09-01
Fix wrong rotation of mirrored pads in plane cut-outs

ubruhin:
Pads from mirrored footprints in boards were cut out with wrong rotation from planes:

grafik


dbrgn:
:joy: that looks like modern art!

PR #510, opened 2019-09-01
Slotted NPTH Holes

### LINKED ISSUE #505: Add support for slotted NPTH
##### SUMMARY

Some manufacturers have quite large minimum milling diameters. This means that thin slots cannot be milled.

However, most support the [G85 Excellon slot command](https://github.com/mcous/gerber-parser/issues/3), where a series of holes is drilled to approximate the desired milling path.

##### SOLUTION

The NPTH tool should be extended to support length and rotation. Export should use G85 excellon commands.

dbrgn:
Still WIP, but early feedback is already welcome.

The approach is adding length and rotation fields to NPTH holes.

Fixes #505.

### DIFF
--- a/libs/librepcb/common/common.pro
+++ b/libs/librepcb/common/common.pro
@@ -64,6 +64,7 @@ SOURCES += \
     dialogs/polygonpropertiesdialog.cpp \
     dialogs/stroketextpropertiesdialog.cpp \
     dialogs/textpropertiesdialog.cpp \
+    drillsize.cpp \
     exceptions.cpp \
     fileio/asynccopyoperation.cpp \
     fileio/csvfile.cpp \
@@ -193,6 +194,7 @@ HEADERS += \
     dialogs/polygonpropertiesdialog.h \
     dialogs/stroketextpropertiesdialog.h \
     dialogs/textpropertiesdialog.h \
+    drillsize.h \
     elementname.h \
     exceptions.h \
     fileio/asynccopyoperation.h \
--- a/libs/librepcb/common/dialogs/holepropertiesdialog.cpp
+++ b/libs/librepcb/common/dialogs/holepropertiesdialog.cpp
@@ -50,6 +50,8 @@ HolePropertiesDialog::HolePropertiesDialog(Hole& hole, UndoStack& undoStack,
   mUi->edtDiameter->setValue(mHole.getDiameter());
   mUi->edtPosX->setValue(mHole.getPosition().getX());
   mUi->edtPosY->setValue(mHole.getPosition().getY());
+  mUi->edtLength->setValue(mHole.getLength());
+  mUi->edtRotation->setValue(mHole.getRotation());
 }
 
 HolePropertiesDialog::~HolePropertiesDialog() noexcept {
@@ -84,6 +86,8 @@ bool HolePropertiesDialog::applyChanges() noexcept {
     cmd->setDiameter(mUi->edtDiameter->getValue(), false);
     cmd->setPosition(Point(mUi->edtPosX->getValue(), mUi->edtPosY->getValue()),
                      false);
+    cmd->setLength(mUi->edtLength->getValue(), false);
+    cmd->setRotation(mUi->edtRotation->getValue(), false);
     mUndoStack.execCmd(cmd.take());
     return true;
   } catch (const Exception& e) {
--- a/libs/librepcb/common/dialogs/holepropertiesdialog.ui
+++ b/libs/librepcb/common/dialogs/holepropertiesdialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>304</width>
-    <height>91</height>
+    <width>391</width>
+    <height>189</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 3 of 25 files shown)


dbrgn:
I wonder if this has some duplication with #423... @ubruhin @pinoaffe?

The problem domain is a bit different though since non-plated holes only have effect on drill files and planes, not directly on the gerber files. So the complexity is lower than #423.

ubruhin:
> I wonder if this has some duplication with #423

Yeah, #423 adds the class `DrillSize` which would make sense for both, slotted pads and slotted holes.

> The problem domain is a bit different though since non-plated holes only have effect on drill files

Do you know if slotted holes in Excellon files are fully supported by PCB manufacturers? Or if other tools (e.g. KiCad) also use the slot command in Excellon files?

> Still WIP, but early feedback is already welcome.

So far:
- You don't have to construct a slot painter path manually, just use `Path::obround()` instead.
- For file format upgrades, the idea is to take the file format version into account instead of detecting the existence of S-Expression nodes. All top level file loading classes know the file format version anyway, this should then be passed to the child objects e.g. by extending the constructor parameters from `const SExpression& node` to `const SExpression& node, const Version& fileFormat`.

dbrgn:
> Yeah, #423 adds the class DrillSize which would make sense for both, slotted pads and slotted holes.

Ok, I'll try to integrate that into my PR.

> Do you know if slotted holes in Excellon files are fully supported by PCB manufacturers? Or if other tools (e.g. KiCad) also use the slot command in Excellon files?

I'm not sure about support. It was recommended to me by Aisler.

Regarding other tools: KiCAD offers two modes when exporting:

export

For this board...

board_000

In the default "route" mode, it generates this:

```
T1
G00X127.27Y-93.853
M15
G01X129.27Y-93.853
M16
```

...while in alternate mode, it uses the G85 slot command (as listed in the [Excellon "spec"](https://web.archive.org/web/20071030075236/http://www.excellon.com/manuals/program.htm)):

```
T1
X127.27Y-93.853G85X129.27Y-93.853
```

The route mode instructs the machine to move to the start position, move the drill tool down, then move to the end position and retract it.

The alternate mode instructs the machine to "drill a slot", however it wants to do that.

In the case of Aisler, they differentiate between slots and cutouts: https://aisler.net/help/design-rules-and-specifications/plated-and-non-plated-slots-and-cutouts Their minimum tool size for slots is 0.7mm while the minimum tool size for cutouts is 1.8mm. @PatrickFranken recommended me to use the slot command for a hole that was not fully handled by the mill because it was too small, so I assume they use G85 for slots and normal routing commands for milling / cutouts. (Is this correct, @PatrickFranken?)

I'm not sure what the correct thing to do would be. My feeling is that obround slotted holes should use the slot command by default, while free-form holes should use the routing commands by default. We could make it configurable like KiCad, but I would expect that most fabs should be able to deal with the slot command (it's a manual process in most cases so they'll be able to remap slot commands to regular routing if they cannot handle slotted holes).

dbrgn:
> For file format upgrades, the idea is to take the file format version into account instead of detecting the existence of S-Expression nodes.

@ubruhin how do I need to proceed regarding the file format? We need to bump the version number somehow, right? Or would this automatically happen if we'd bump the version to 0.2? Can I simply change the file format in-place or does it need some kind of migration?

PatrickFranken:
Let me briefly elaborate a bit on why we prefer slot commands in our drill files compared to cutouts on the board outline.
First of all, it doesn't matter if it's done as G85 or in routing mode. Both are usually normalized to one of these methods during CAM.
A huge advantage of drill files compared to milling / board outline / edge cuts is that there are usually separate drill files for PTH and NPTH. This avoids the ambiguity which you've when merging both together. The CAM applications then try to decided based on the copper layer whether to make the hole plated or non-plated. This could go wrong in some cases.

To understand why plated slots have to be drilled (not milled / routed) we've to see how the PCBs are manufactured. Everything plated has to be done (obviously) before copper plating. As there is no milling / routing step before plating it has to be done during the drilling step. So how do we do slots using a drill? The process is called "Nibbling". It means putting a multitude of drills in a row very close to each other. This has do be defined in the drill file!

So if an application like EAGLE put slots on the board outline a CAM process has to make two error prone decisions
1) Is the slot plated or not plated?
2) Is it possible to nibble the slot using the tools we have? If not is has to be reject or gets overly expensive as we an additonal routing / milling step before plating.

Please don't be like EAGLE and be explicit about your slots. Use two drill files for PTH and NPTH and use slots commands, either Excellon routing mode or G85.

Patrick

ubruhin:
> how do I need to proceed regarding the file format? We need to bump the version number somehow, right? Or would this automatically happen if we'd bump the version to 0.2? Can I simply change the file format in-place or does it need some kind of migration?

@dbrgn let me do that in a separate PR. Since we never had a file format version bump until now, there are still some things missing for such a migration...

@PatrickFranken Thanks for the explanation, makes sense to me. Since LibrePCB already generates separate drill files for NPTH and PTH, we can put the slots into the corresponding drill file too.

dbrgn:
@ubruhin ok, then I'll wait for your PR :slightly_smiling_face:

ubruhin:
@dbrgn Since #652 is now merged, could you rebase this branch to master and finish the file format upgrade for holes?

dbrgn:
Closing in favor of #1071.

PR #511, opened 2019-09-02
Add Gerber export test

ubruhin:
Since there were some issues regarding the Gerber export, I added a test which now automatically checks the Gerber export, hopefully avoiding/reducing new such bugs.

The concept is simple: In the test data directory there is a project which uses as many features as possible (all pad shapes, planes, mirrored footprints, ...):

grafik

(yes, it looks a bit strange :grin:)

Then I exported the board to Gerber files and reviewed them manually (in Gerbv):

grafik

If everything looks good, those Gerber files are added to the test data repository. Now the test just exports Gerber data of the same project and compares the generated files with the expected data. If the data differs, the test fails.

Of course now we have to do the review again every time the Gerber output changes in some way.

Btw, there are also many other tests modified because I refactored the test data directory (see the separate commits for details).


ubruhin:
Hmm interestingly the Gerber output on Windows seems to be slightly different :sob:

dbrgn:
Good idea to do this test. The thing with different output on Windows is a very nice catch thanks to this approach :slightly_smiling_face: 

Is there a simple way to turn a Gerber file into a PNG on the command line? Then we could even show a diff if the generated and expected files are different.

ubruhin:
> The thing with different output on Windows is a very nice catch thanks to this approach

Yeah, but it also introduces new issues ;) It seems that on AppVeyor some floating point calculations return slightly (~1nm) different results than on my Linux machine. For the Gerber output that's not a big issue (1nm doesn't really matter), but of course it makes the new test fail.

At least one of the difference in Gerber files is caused by different results of the method `Toolbox::arcCenter()`. I temporary added debug output to see the value of each variable:

https://github.com/LibrePCB/LibrePCB/blob/73f853129f2f15266310c466fe43464ec18ecb07/libs/librepcb/common/toolbox.cpp#L77-L113

The output from my local machine (left side) vs. AppVeyor (right side):

grafik

After rounding to our internal resolution of 1nm, this makes a difference of 1nm and thus affects the Gerber files.

I'm not sure if and how we could fix this issue... Any idea @rnestler @dbrgn ?

> Is there a simple way to turn a Gerber file into a PNG on the command line? Then we could even show a diff if the generated and expected files are different.

I don't know, never looked for it...

dbrgn:
Maybe it's a compiler issue and not a platform issue. But in any case, it sucks...

I don't have a good idea right now that works as a long-term solution.

ubruhin:
> Maybe it's a compiler issue and not a platform issue.

Yes, that's possible. For example in a Windows 10 VM I couldn't reproduce the issue.

> But in any case, it sucks...

Floating point types = problems for sure (platform independent rule of thumb) :grin:

If we don't see a solution now, I think I would just disable that test on Windows and open an issue to keep this in mind. It's ugly, but better than not adding this new test at all (I think it's a very valuable test since it's basically our only "quality check" of the Gerber output).

dbrgn:
> If we don't see a solution now, I think I would just disable that test on Windows and open an issue to keep this in mind.

I agree! Make sure to add a comment stating why it was disabled.

ubruhin:
> Make sure to add a comment stating why it was disabled.

Done, and #516 created to track this issue.

rnestler:
> Maybe it's a compiler issue and not a platform issue. But in any case, it sucks...

It may also be an issue with the formatting of the floating point variable, which differ between standard library implementations.

PR #518, opened 2019-09-18
proposal for documenting fields and files

ouabache:
initial proposal

### DIFF
--- a/dev/doxygen/pages/library.md
+++ b/dev/doxygen/pages/library.md
@@ -152,6 +152,101 @@ There exist two types of categories, while all regular element types use exactly
     - librepcb::library::Package
 
 
+# Field Formats {#doc_library_field_formats}
+
+
+
+- TEXT: strings are always enclosed in double quotes. "TEXT"
+    - Use A through Z,a through z and 0-9.
+    - Use comma,space,dash,underscore and period.
+    - Any field may impose additional restrictions (IE: no spaces).
+    - Special character for LINEFEED not yet defined.
+
+- NUMERIC: numbers  are never enclosed in double quotes.
+    - Use  0-9, dash and period
+
+
+- UUID: Universal unique identifiers are used to track objects:
+    - version 4 (random).
+    - use lower case hexidecimal digits.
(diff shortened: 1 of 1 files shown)


ubruhin:
Thanks for working on this :smiley: 

My thoughts so far:
- I would move the basic S-Expressions documentation (general structure, charset, EOL character, formatting, ...) into a separate page (e.g. called "File Format") since it's not specific to libraries. Or even move everything into this new page, so all *.lp files are documented in one page to avoid clutter up basic documentation of libraries and projects.
- There are a lot of empty lines - please only use 1 empty line to separate paragraphs and 2 empty lines to separate chapters.
- I don't really understand the right column of the file format documentation :wink: What are the `LP_*` terms used for?
- I have not much experience with documenting file formats, but maybe it would make sense to use something like [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) for that? @dbrgn @rnestler, any opinion about that?

dbrgn:
BNF (or [EBNF](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form)) would work, but since we don't actually parse the file format spec, we don't benefit that much from a structured format. On the other hand, the BNF would give us re-usable types, so a `(pos x y)` type only needs to be defined once and can then be re-used.

@rnestler what do you think?

ouabache:
The right column was my attempt to cover all the use cases:

1) A field must be defined only one time
2) A field must be defined but can have more than one definition
3) A field may be defined zero or one time only
4) A field may be defined zero or multiple times.

The LP_ names are simply to have a name that you can refer to for that
field from the documentation


I don't have a lot of experience with file formats either. If I see an
example then I can understand it easier than trying
to parse some formal definition.



I also put in a pull request for 10 bga packages for the LibrePCB_Base
library.

web-coder-of-rpi:
> initial proposal

Remember @ouabache, make sure not to make to significant changes to file format. But it's probably fine if LibrePCB ignores the messages, like <code> < !-- comment -- > </code> does in Markdown and HTML.

PR #521, opened 2019-09-24
Package editor: Add more widgets to text drawing command toolbar

ubruhin:
When the "Add Text" tool is active, stroke width and text alignment can now be edited too in the toolbar:

grafik


dbrgn:
Nice :slightly_smiling_face:

PR #523, opened 2019-09-26
Library editor: Adjust default values of sym/pkg elements

ubruhin:
When creating symbols or packages, some properties were initialized with default values which aren't reasonable IMHO. Thus I reviewed and documented (in code) all default values and adjusted some of them.

To comply with our [library conventions](https://docs.librepcb.org/#libraryconventions):
- Footprint circle line width changed from 0.254mm to 0.2mm

To avoid unnecessary grab areas "by accident":
- Footprint circle/polygon grab area disabled

To avoid inch-based numbers with many decimal places where metric numbers could be used instead (most spinboxes have an interval of 0.1mm, thus it's more reasonable to use default numbers which are a multiple of 0.1mm):
- Footprint THT pad size changed from 1.27mm x 0.635mm to 1.5mm x 0.7mm
- Footprint SMT pad size changed from 2.54mm x 1.27mm to 2.5mm x 1.3mm
- Symbol text height changed from 2.54mm to 2.5mm
- Symbol circle/polygon line width changed from 0.254mm to 0.2mm

The latter change is actually the one with the most impact; it would require to adjust all already existing symbols to use an outline width of 0.2mm instead of 0.254mm. Actually the symbol outline width is not documented yet in our library conventions, but I would now recommend to use 0.2mm. Same for text height, I would recommend 2.5mm instead of 2.54mm. Or what do you think @dbrgn?

### DIFF
--- a/libs/librepcb/eagleimport/symbolconverter.cpp
+++ b/libs/librepcb/eagleimport/symbolconverter.cpp
@@ -122,7 +122,7 @@ std::unique_ptr<library::Symbol> SymbolConverter::generate() const {
     }
     PositiveLength height(Length::fromMm(text.getSize()) * 2);  // can throw
     if ((textStr == "{{NAME}}") || (textStr == "{{VALUE}}")) {
-      height = PositiveLength(2540000);
+      height = PositiveLength(2500000);
     }
     Point     pos = Point::fromMm(text.getPosition().x, text.getPosition().y);
     Angle     rot = Angle::fromDeg(text.getRotation().getAngle());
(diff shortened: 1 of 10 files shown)


dbrgn:
> it would require to adjust all already existing symbols to use an outline width of 0.2mm instead of 0.254mm.

It doesn't *require* us to do so, but of course it would be nice to adjust them.

I like the new metric defaults.

ubruhin:
OK, the new defaults are now documented in the library conventions (see https://github.com/LibrePCB/librepcb-doc/pull/38).

> It doesn't require us to do so, but of course it would be nice to adjust them.

Correct :)

PR #531, opened 2019-10-05
Board editor: Allow to hide copper planes

### LINKED ISSUE #481: Allow to hide copper planes
##### SUMMARY
Currently there is no way to hide copper planes. If they are always visible they are distracting while doing the rest of the layout.

##### SOLUTION

Provide a way to hide the planes completely. Optimally it would also provide an outline mode, where only the outline of the copper plane is visible.

ubruhin:
- Adding a context menu item to planes to allow hiding them:
grafik
- Adding menu items in the main menu allow to hide/show all planes at once:
grafik
- When opening the properties dialog of a plane, the plane will automatically be shown while the dialog is open (to see the effect when you change some plane properties).
- The outlines of planes are always visible, only the area can be hidden.
- By default, all planes are visible.
- I'll open an issue to store the planes visibility in user settings (needs to be delayed for the 0.2.0 release because of the file format change, but the feature itself can already be included in 0.1.3).

Fixes #481


dbrgn:
Cool, very nice!

Issue #43, opened 2016-01-07
Fix doxygen build warnings


ubruhin:
Since all classes were moved into a new namespace "librepcb" (1d361e6a33e4ff12ede683da00f7211ea93b100f), doxygen now throws lot of warnings like `warning: explicit link request to 'SomeClass' could not be resolved`

I would mentor this issue if someone wants to work on it.

javsanpar:
I would like to work on this issue 😄. I'm going to build doxygen to see the warnings.

ubruhin:
> I would like to work on this issue

Great! :+1: 

Just let me know if you have any questions :)

ubruhin:
Fixed in #537.

PR #537, opened 2019-10-08
Fix doxygen build warnings

javsanpar:
Add librepcb namespace to doxygen comments. #43 

Some warnings are already fixed if you want to take a look.

### DIFF
--- a/dev/doxygen/Doxyfile
+++ b/dev/doxygen/Doxyfile
@@ -208,7 +208,7 @@ TAB_SIZE               = 4
 # You can put \n's in the value part of an alias to insert newlines.
 
 ALIASES                =
-ALIASES += undocmd{1}="@note This method must be called only from inside an #UndoCommand!\nUndo Commands for this method: \1"
+ALIASES += undocmd{1}="@note This method must be called only from inside an #librepcb#UndoCommand!\nUndo Commands for this method: \1"
 
 # This tag can be used to specify a number of word-keyword mappings (TCL only).
 # A mapping has the form "name=value". For example adding
--- a/dev/doxygen/pages/code_style_guide.md
+++ b/dev/doxygen/pages/code_style_guide.md
@@ -48,7 +48,7 @@ This page describes the code style guide for LibrePCB developers.
 # Naming {#doc_code_style_guide_naming}
 
 - File Names: Exactly same naming as classes, but all characters lowercase. Example: `ses_select.cpp`, `ses_select.h`
-- Include guards: Use the namespace as prefix before the class name. Example: `#ifndef NAMESPACE_CLASSNAME_H`
+- Include guards: Use the namespace as prefix before the class name. Example: `NAMESPACE_CLASSNAME_H`
 - Namespaces: lowercase (with underline as separator if needed)
 - Classes: UpperCamelCase (for common prefixes, use the underline as separator. Example: librepcb::project::editor::SES_Select)
 - Interfaces: Like Classes, but with prefix "IF_". Example: librepcb::IF_GraphicsViewEventHandler
--- a/dev/doxygen/pages/mainpage.md
+++ b/dev/doxygen/pages/mainpage.md
@@ -18,14 +18,6 @@ LibrePCB Developers Documentation {#mainpage}
 - @ref doc_server_api
 
 
-# Related Pages {#doc_main_related_pages}
-
-- @ref todo
-- @ref bug
-- @ref deprecated
-- @ref test
-
-
 # External Links {#doc_main_external}
 
 - Contribution Guidelines: https://github.com/LibrePCB/LibrePCB/blob/master/CONTRIBUTING.md
--- a/dev/doxygen/pages/repository.md
+++ b/dev/doxygen/pages/repository.md
@@ -13,7 +13,7 @@ A repository is a server which holds a list of available libraries. The librarie
 somewhere as Zip files, for example on the Releases page of a GitHub repository. The library manager
 retrieves the library list from the repositories to display the available libraries. If the user
 wants to install such a library, the library manager just downloads these Zip files and extracts
-them into the [remote libraries directory](@ref doc_workspace_remote_dir) in the @ref workspace.
+them into the [remote libraries directory](@ref doc_workspace_remote_dir) in the @ref doc_workspace.
 After an automatically triggered rescan of the @ref doc_workspace_library_database, the library is
 ready to use.
 
--- a/dev/doxygen/pages/workspace.md
+++ b/dev/doxygen/pages/workspace.md
@@ -34,7 +34,7 @@ projects in an arbitrary, hierarchical system inside that directory. But of cour
 possible to store projects outside this directory.
 
 Each project is represented by a directory, see its
-[file structure documentation](@ref doc_project_file_structure) for details.
+[file structure documentation](@ref doc_project_filestructure) for details.
 
 ## v<VERSION>/ {#doc_workspace_version_dir}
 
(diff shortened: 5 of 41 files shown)


ubruhin:
Thanks for working on this! :smiley: 

Just one comment so far: It would be great to start all the links with the `#` character, e.g. `#librepcb#AttrTypeString` instead of `librepcb#AttrTypeString`. This helps to avoid broken links if we rename/move/remove classes in future since Doxygen warns about the broken links if the link starts with `#`. So `#librepcb#SomeNonExistentClass` would throw a warning, while `librepcb#SomeNonExistentClass` does not.

You don't have to adjust the already fixed links, but it would be great if you could do it that way from now on :-)

Btw, actually I no longer like the syntax with `#`, actually Doxygen allows to use `::` instead (e.g. `::librepcb::AttrTypeString`), which I like much more since it's C++ style :wink: But no need to change that (now), we could still apply a global search&replace with a regex some time later...

javsanpar:
> Thanks for working on this! smiley

You're welcome.
 
> Just one comment so far: It would be great to start all the links with the `#` character, e.g. `#librepcb#AttrTypeString` instead of `librepcb#AttrTypeString`. This helps to avoid broken links if we rename/move/remove classes in future since Doxygen warns about the broken links if the link starts with `#`. So `#librepcb#SomeNonExistentClass` would throw a warning, while `librepcb#SomeNonExistentClass` does not.
> 
> You don't have to adjust the already fixed links, but it would be great if you could do it that way from now on :-)

I have adjusted the already fixed links :grimacing:. It was easy with this: https://stackoverflow.com/questions/28697799/how-to-search-and-replace-a-string-in-the-lines-added-in-the-last-x-commits

> Btw, actually I don't like the syntax with `#`, actually Doxygen allows to use `::` instead (e.g. `::librepcb::AttrTypeString`), which I like much more since it's C++ style wink But no need to change that (now), we could still apply a global search&replace with a regex some time later...

Okay!

Do you know why UndoCommand appears as "@undocmd" in some files (for example: project.h)?

ubruhin:
> I have adjusted the already fixed links grimacing. It was easy with this: https://stackoverflow.com/questions/28697799/how-to-search-and-replace-a-string-in-the-lines-added-in-the-last-x-commits

Perfect! :smiley:  

> Do you know why UndoCommand appears as "@undocmd" in some files (for example: project.h)?

Yes, `@undocmd` is an alias I have defined here:

https://github.com/LibrePCB/LibrePCB/blob/2e747216114042648c9677d7c18301b793b80918/dev/doxygen/Doxyfile#L211

It is rendered special in the HTML output, but it seems this command is broken too, the namespace for `UndoCommand` is missing ;)

grafik

javsanpar:
> It is rendered special in the HTML output, but it seems this command is broken too, the namespace for `UndoCommand` is missing ;)

Nice, I will fix it!

Also, I have compared the Build Doxygen logs with # (`#librepcd`) and without # (`librepcb`). In the logs with # there are more warnings. Do you know why?

You can take a look on these links:

- With # (`#librepcd`) https://dev.azure.com/LibrePCB/LibrePCB/_build/results?buildId=379&view=logs

- Without # (`librepcb`) https://dev.azure.com/LibrePCB/LibrePCB/_build/results?buildId=371&view=logs

ubruhin:
> Also, I have compared the Build Doxygen logs with # (#librepcd) and without # (librepcb). In the logs with # there are more warnings. Do you know why?

Yes, that's what I tried to explain in my comment above, but I guess it was not clear enough, sorry ;) 

- Without the `#` prefix (e.g. `librepcb#UndoCommand`), Doxygen is lazy - if there exists such a class, it creates a hyperlink to that class. But if such a class does not exist, Doxygen simply ignores it and you will find the raw string `librepcb#UndoCommand` in the HTML output, without hyperlink.
- With the `#` prefix, Doxygen is strict - if the specified class exists, a hyperlink is created. But if it doesn't exist, Doxygen throws a warning in the console output so we can see that our Documentation contains invalid references.

In addition, Doxygen has a configuration key called [`WARN_AS_ERROR`](http://www.doxygen.nl/manual/config.html#cfg_warn_as_error). If we set this to `true`, Doxygen will immediately abort with an error code if there are warnings (like invalid references), and thus our CI would fail too. This makes it much simpler to keep the documentation up to date since we are forced to fix documentation warnings.

Currently we can't enable `WARN_AS_ERROR` because we have a lot of warnings. But once you fixed all warnings, I'd like to enable `WARN_AS_ERROR` since this will lead to a higher quality of our Doxygen documentation.

javsanpar:
Sorry, I didn't get it. I didn't notice there were other namespaces. :sweat_smile:

I have been fixing the hyperlinks, but I am having some difficulties with these links:

https://github.com/LibrePCB/LibrePCB/blob/cefa7e771b238ebcefdf78191274538b92705042/libs/librepcb/workspace/workspace.h#L255
https://github.com/LibrePCB/LibrePCB/blob/dd2dd1bc6366f05940b74843436537df4c8ba2ac/libs/librepcb/common/units/lengthunit.h#L119
https://github.com/LibrePCB/LibrePCB/blob/cefa7e771b238ebcefdf78191274538b92705042/libs/librepcb/common/undostack.h#L199

ubruhin:
Ah, those are broken references to members of the same class. Here the fixes:

- "to lock the version directory (#mVersionPath)" should be changed to "to lock the metadata directory (#mMetadataPath)"
- "Use #toString() and #fromString() instead for this purpose." -> just remove this sentence :-)
- "#isCommandActive()" should be changed to "#isCommandGroupActive()"

ubruhin:
@javsanpar What's actually the state of this?

javsanpar:
Hello, @ubruhin! Sorry for the absence

> @javsanpar What's actually the state of this?

There are several parameters which I changed and need a description:
https://github.com/javsanpar/LibrePCB/blob/ea8b126735f59815df2c707316ffb72ee4e7717e/libs/librepcb/common/attributes/attributesubstitutor.h#L77
https://github.com/javsanpar/LibrePCB/blob/ea8b126735f59815df2c707316ffb72ee4e7717e/libs/librepcb/common/attributes/attributesubstitutor.h#L78
https://github.com/javsanpar/LibrePCB/blob/ea8b126735f59815df2c707316ffb72ee4e7717e/libs/librepcb/project/project.h#L91
https://github.com/javsanpar/LibrePCB/blob/ea8b126735f59815df2c707316ffb72ee4e7717e/libs/librepcb/project/project.h#L92
https://github.com/javsanpar/LibrePCB/blob/ea8b126735f59815df2c707316ffb72ee4e7717e/libs/librepcb/project/project.h#L453
https://github.com/javsanpar/LibrePCB/blob/ea8b126735f59815df2c707316ffb72ee4e7717e/libs/librepcb/project/project.h#L454

They still are warnings in some Markdown files:
image
Do you have any clue on how to fix those warnings?

ubruhin:
Hi @javsanpar,

> Sorry for the absence

No problem :)

>  There are several parameters which I changed and need a description:
>  They still are warnings in some Markdown files:

I could fix them by myself and push a commit to your branch, is this OK for you?

And as explained in https://github.com/LibrePCB/LibrePCB/pull/537#issuecomment-540170170, I'd like to apply a global search&replace to replace `#` by `::`. Then I'd say this branch is ready to merge :)

javsanpar:
> I could fix them by myself and push a commit to your branch, is this OK for you?

Great! No problem.

> And as explained in #537 (comment), I'd like to apply a global search&replace to replace # by ::. Then I'd say this branch is ready to merge :)

Perfect. Anything I can help you with?

ubruhin:
I fixed the remaining warnings and rebased this branch to master. If CI succeeds, I think this is ready to merge.

Thanks for your contribution @javsanpar! :tada:

PR #556, opened 2019-10-20
Library editor: Set top pins text same orientation as bottom ones

### LINKED ISSUE #534: Library Editor: Wrong top pin text rotation
##### SUMMARY & SOLUTION
Set top pins text same orientation as bottom ones, like classic texts
(I know that it is preferable to make functional symbol, but I will do both)

image

mexchip:
Fixes #534

### DIFF
--- a/libs/librepcb/common/graphics/primitivetextgraphicsitem.cpp
+++ b/libs/librepcb/common/graphics/primitivetextgraphicsitem.cpp
@@ -27,6 +27,8 @@
 #include <QtCore>
 #include <QtWidgets>
 
+#include <iostream>
+
 /*******************************************************************************
  *  Namespace
  ******************************************************************************/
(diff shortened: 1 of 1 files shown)


ubruhin:
Hi,

Unfortunately this doesn't work properly ;) For example if the angle of a pin is -1°, the text is drawn upside down:

grafik

An angle of ]-90°, +90°] (i.e. excluding -90°) should be drawn with normal text orientation, and an angle of ]+90°, +270°] (i.e. excluding +90°) should be drawn with 180° rotated text.

mexchip:
I think I've finally got it working. I'm attaching some pictures so that it might be easier to confirm before pulling the code.

The -1° commented previously:
image

> An angle of ]-90°, +90°] (i.e. excluding -90°) should be drawn with normal text orientation, and an angle of ]+90°, +270°] (i.e. excluding +90°) should be drawn with 180° rotated text.

So -90° should be rotated:
image

But -89° not:
image

Also, 90° would be drawn normally:
image

But 91° would be rotated:
image

270° rotated:
image

271° normal orientation:
image

ubruhin:
Now it seems to work in the symbol editor, but fails for other use-cases of the `PrimitiveTextGraphicsItem` (that class is very generic, thus it is used at several locations) :wink:

I see two issues:
- The method `parentItem()` returns a pointer to the parent item, or `nullptr` if there is no parent item. Since you dereference that pointer without checking it, the application crashes if there is no parent item (for example in the package editor if you remove all footprints, a text without parent is shown and thus makes the application crash).
- Now you only take the rotation of the direct parent item into account. The rotation of `PrimitiveTextGraphicsItem` itself, and the rotation if all other parent items (graphics items can be recursively nested) is ignored. But to decide whether the text needs to be rotated by 180° we have to know the *overall* rotation of the graphics item, i.e. including rotation of all parent items. Thus currently the "trick" with `mapToScene()` is used since this respects the overall transformation. Maybe [`sceneTransform()`](https://doc.qt.io/qt-5/qgraphicsitem.html#sceneTransform) could also be helpful to determine the overall rotation (?).

mexchip:
Sorry for not checking your response earlier, I got a bit busy and will be able to work on this till the weekend. Thanks for your time.

mexchip:
I'll be working with this a little bit, I don't know if I should keep this pull request open or close and create it again once I'd like you to review it.

ubruhin:
OK, you can leave the PR open and notify me as soon as it's ready to review.

mexchip:
> 
>     * Now you only take the rotation of the direct parent item into account. The rotation of `PrimitiveTextGraphicsItem` itself, and the rotation if all other parent items (graphics items can be recursively nested) is ignored. But to decide whether the text needs to be rotated by 180° we have to know the _overall_ rotation of the graphics item, i.e. including rotation of all parent items. Thus currently the "trick" with `mapToScene()` is used since this respects the overall transformation. Maybe [`sceneTransform()`](https://doc.qt.io/qt-5/qgraphicsitem.html#sceneTransform) could also be helpful to determine the overall rotation (?).

What would be a way to test the overall rotation of the graphics item? What could be other required tests for PrimitiveTextGraphicsItem?

ubruhin:
I'm not sure what would be the best solution for determining the overall rotation. But probably the simplest variant would be to create a helper function (e.g. in toolbox.cpp/h) which traverses up through all the parents of a `QGraphicsItem` and sums up all their rotations...

mexchip:
I've updated the branch, could you please help me testing it?

ubruhin:
Hmm to be honest, that looks a bit "complicated" to me :wink: When looking at the code, I have no idea what all the comparisons actually check...

What about my idea from above, did you try that already?

> But probably the simplest variant would be to create a helper function (e.g. in toolbox.cpp/h) which traverses up through all the parents of a QGraphicsItem and sums up all their rotations...

(actually the method doesn't need to be in the toolbox yet - a method like `GetTotalRotation()` within `PrimitiveTextGraphicsItem` would be fine for now)

ubruhin:
I have now fixed the rotation issue with a cleaner and simpler implementation in #967 (auto-rotation detection is no longer done in `PrimitiveTextGraphicsItem` at all). Thanks anyway for helping and sorry for closing this PR :slightly_smiling_face:

PR #561, opened 2019-10-20
Schematic editor: Fix device when adding component multiple times

### LINKED ISSUE #547: Schematic Editor : only first (or no) device is good when adding multiple times the same
```
LibrePCB Version: 0.1.3-unstable
Git Revision:     2e747216114042648c9677d7c18301b793b80918
Build Date:       2019-10-10T18:14:18
Qt Version:       5.12.3 (built against 5.12.3)
CPU Architecture: x86_64
Operating System: Ubuntu 18.04.3 LTS
Platform Plugin:  xcb
```

##### SUMMARY
When adding multiple time the same device, only the first one is good int the "Place Devices" dock the other seems to be the default device of the component.
When clicking "Auto-Add-All", none are good.

##### STEPS TO REPRODUCE
1. Open the "Add component" window
2. select the 1608 resistor device
3. insert it multiple times in the schematic
image

4. go to board and checks devices

image
***********
image
***********
image

5. add all devices
=> all are bad ones
image

6. remove all devices from the board
=> it goes to step 4
7. now add the first one with "Add" button
=> a 1608 resistor is added
=> now all others have the good device in the "Place Devices" dock
image

8. But if you add all of them, none is good
image

:angry:

##### EXPECTED RESULTS
All are 1608 resistors

##### ACTUAL RESULTS
Just the first one is 1608.
Others are through hole ones.

ubruhin:
When adding a component with a pre-selected device multiple times to a schematic, the device was set only on the first component placed. All successive components didn't have a device selected.

Fixes #547

### DIFF
--- a/libs/librepcb/projecteditor/schematiceditor/fsm/ses_addcomponent.cpp
+++ b/libs/librepcb/projecteditor/schematiceditor/fsm/ses_addcomponent.cpp
@@ -304,10 +304,13 @@ SES_Base::ProcRetVal SES_AddComponent::processSceneEvent(
                   mCurrentComponent->getLibComponent().getUuid();
               Uuid symbVarUuid =
                   mCurrentComponent->getSymbolVariant().getUuid();
+              tl::optional<Uuid> defaultDeviceUuid =
+                  mCurrentComponent->getDefaultDeviceUuid();
               mUndoStack.commitCmdGroup();
               mIsUndoCmdActive = false;
               abortCommand(false);  // reset attributes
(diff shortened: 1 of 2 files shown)


ubruhin:
> If I understand this correctly, the device was passed to the first "add" command, but not to subsequent ones, right?

Correct :)

doom-fr:
Think there is still a problem as my U11 is ATmega324PB (ATmegaxx4PB of my Microchip lib).
You can check my pull request to see any problem but I don't noticed anything.

image

But in board editor, it is a ATmegaxx8PB

image

PR #563, opened 2019-10-25
LibraryEditor: Fix missing message box when closing tab

### LINKED ISSUE #492: Library editor: Changes in text field with focus not saved when closing tab
```
LibrePCB Version: 0.1.2
Git Revision:     acdd94d9d2310f79215125b999153e9da88a9376
Build Date:       2019-07-13T21:11:00
Qt Version:       5.12.3 (built against 5.12.3)
CPU Architecture: x86_64
Operating System: Ubuntu 18.10
Platform Plugin:  xcb
```

##### SUMMARY
Does not prompt to save after library changes are made

##### STEPS TO REPRODUCE
- Edit a library part
- change the description
- close tab

##### EXPECTED RESULTS
it should prompt to save data

##### ACTUAL RESULTS
data lost

##### ADDITIONAL INFORMATION
If you change the description AND the keywords then it will prompt. It only
fails if a single thing is changed

ubruhin:
When closing a tab in the library editor while some property (like name,
description, ...) was just modified, the "save changes?" message box did
not appear. Now the focus is explicitly moved out from the editor widget
to ensure the "dirty" state of editors gets updated when closing a tab.

Fixes #492.

### DIFF
--- a/libs/librepcb/libraryeditor/libraryeditor.cpp
+++ b/libs/librepcb/libraryeditor/libraryeditor.cpp
@@ -564,6 +564,15 @@ bool LibraryEditor::closeTab(int index) noexcept {
     return false;
   }
 
+  // Move focus out of the editor widget to enforce updating the "dirty" state
+  // of the editor before closing it. This is needed to make sure the
+  // "save changes?" message box appears if the user just edited some property
+  // of the library element and the focus is still in the property editor
+  // widget. See https://github.com/LibrePCB/LibrePCB/issues/492.
+  if (QWidget* focus = focusWidget()) {
+    focus->clearFocus();
+  }
+
   // Handle closing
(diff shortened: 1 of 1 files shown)


dbrgn:
Cool! Does this only apply to closing, or also when pressing "apply"?

It regularly happens to me that I update a value (e.g. a polygon point), then press "apply", but nothing happens because the focus is still on the input field.

ubruhin:
> Does this only apply to closing, or also when pressing "apply"?

It applies only to closing tabs, it's not related to dialogs.

> It regularly happens to me that I update a value (e.g. a polygon point), then press "apply", but nothing happens because the focus is still on the input field.

Good point, but I'll try to fix it in a separate PR since it's a different topic.

PR #564, opened 2019-10-25
Board editor: Tool to measure wire length

dbrgn:
Allow measuring the length of the selected wire segment.

img

If not all selected netlines are directly connected, a warning is shown.

img

The functionality is currently quite limited. It only measures from start point to end point and does not consider the width of the netline. However, it's quite hard to define what the correct behavior should be.

Other changes in this PR:

- `Point::getLength` now returns `UnsignedLength`
- `BI_NetLine` now has a `getLength` method
- In the board editor, right clicking will not clear the current selection anymore
- Added a link to https://icons8.com/ to the about dialog, so we can now use all their icons for free


libs/librepcb/common/dialogs/aboutdialog.ui
```diff
@@ -312,7 +314,7 @@
            <property name="textInteractionFlags">
             <set>Qt::TextSelectableByKeyboard|Qt::TextSelectableByMouse</set>
            </property>
-           <property name="placeholderText" stdset="0">
+           <property name="placeholderText">
```
ubruhin: This should be reverted since it breaks compatibility with older Qt versions.

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -759,6 +771,91 @@ bool BES_Select::removeSelectedItems() noexcept {
   }
 }
 
+void BES_Select::measureLengthInDirection(bool              directionBackwards,
+                                          const BI_NetLine& netline,
+                                          QSet<Uuid>&       visitedSegments,
+                                          UnsignedLength&   totalLength) {
+  auto currentNetline = std::cref(netline);
```
ubruhin: Hmm why not simply using a pointer to netline? i.e.

```cpp
const BI_NetLine* currentNetline = &netline;
```

Same for `currentAnchor`.

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -759,6 +771,91 @@ bool BES_Select::removeSelectedItems() noexcept {
...
+  auto currentAnchor  = std::cref(directionBackwards ? netline.getStartPoint()
+                                                    : netline.getEndPoint());
+
+  for (;;) {
+    tl::optional<BI_NetLine&> nextNetline = tl::nullopt;
```
ubruhin: Same here - IMHO the `optional` is just useless complexity, a pointer should be enough (and simpler):

```suggestion
    const BI_NetLine* nextNetline = nullptr;
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -759,6 +771,91 @@ bool BES_Select::removeSelectedItems() noexcept {
...
+    for (BI_NetLine* nl : currentAnchor.get().getNetLines()) {
```
ubruhin: I guess here you could use `const`(?):

```suggestion
    for (const BI_NetLine* nl : currentAnchor.get().getNetLines()) {
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -759,6 +771,91 @@ bool BES_Select::removeSelectedItems() noexcept {
...
+      currentNetline = std::cref(*nextNetline);
+      currentAnchor =
+          std::cref(*nextNetline->getOtherPoint(currentAnchor.get()));
+    } else {
+      break;
+    }
+  }
+}
+
+bool BES_Select::measureSelectedItems(BI_NetLine& netline) noexcept {
```
ubruhin: Probably the netline could be `const`(?):

```suggestion
bool BES_Select::measureSelectedItems(const BI_NetLine& netline) noexcept {
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -759,6 +771,91 @@ bool BES_Select::removeSelectedItems() noexcept {
...
+  Board* board = mEditor.getActiveBoard();
+  Q_ASSERT(board);
+  if (!board) return false;
+
+  Q_ASSERT(netline.isSelected());
+
+  // Store UUIDs of visited segments
+  QSet<Uuid> visitedSegments;
```
ubruhin: For such use-cases I usually use a set of pointers (`QSet<const BI_NetLine*>`) since a pointer is even "more unique" than a UUID and probably also more efficient :wink: However, `QSet<Uuid>` is also fine for me...

But I'd really rename the variable since it doesn't contain segments (in the meaning of `BI_NetSegment`), but netlines:

```suggestion
  QSet<Uuid> visitedNetLines;
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -759,6 +771,91 @@ bool BES_Select::removeSelectedItems() noexcept {
...
+
+  // Get the netline length. Then traverse the selected segments first in one
+  // direction, then in the other direction.
+  UnsignedLength totalLength = netline.getLength();
+  try {
+    measureLengthInDirection(false, netline, visitedSegments,
+                             totalLength);  // can throw
+    measureLengthInDirection(true, netline, visitedSegments,
+                             totalLength);  // can throw
+  } catch (const LogicError& e) {
```
ubruhin: I'd catch all exception types here since it's safer and has no disadvantages:

```suggestion
  } catch (const Exception& e) {
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -759,6 +771,91 @@ bool BES_Select::removeSelectedItems() noexcept {
...
+      // Don't visit a segment twice
+      if (visitedSegments.contains(nl->getUuid())) {
+        continue;
+      }
+      // Only visit selected segments
+      if (nl->isSelected()) {
+        if (nextNetline.has_value()) {
+          // There's already another connected and selected netline
+          throw LogicError(__FILE__, __LINE__,
+                           tr("Selected segment may not have branches"));
```
ubruhin: Add a period (or exclamation mark) at the end of sentences (it's shown in a message box and looks strange without the period):

```suggestion
                           tr("Selected segment may not have branches."));
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -759,6 +771,91 @@ bool BES_Select::removeSelectedItems() noexcept {
...
+  auto query = board->createSelectionQuery();
+  query->addSelectedNetLines();
+  int totalSelectedNetlines = query->getNetLines().size();
+
+  // Show result
+  QString title = tr("Measurement Result");
+  QString text  = tr("Total length of %1 segments: %2 mm / %3 in")
+                     .arg(QString::number(visitedSegments.count()),
+                          QString::number(totalLength->toMm()),
+                          QString::number(totalLength->toInch()));
```
ubruhin: `QString::number()` does not format numbers according the user's locale, and it's not needed anyway when using `.arg()` multiple times:

```cpp
                     .arg(visitedSegments.count())
                     .arg(totalLength->toMm(), 0, 'f', 6)
                     .arg(totalLength->toInch(), 0, 'f', 6));
```

Unfortunately this will show trailing zeros. To avoid them, one could use `Toolbox::floatToString()` instead.

libs/librepcb/common/dialogs/aboutdialog.ui
dbrgn: QtCreator did that automatically... I'll revert.

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
dbrgn: I'm allergic to raw pointers :wink: :crab: 

I'll change it.

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
dbrgn: Good point, I'll rename.

In theory there could be two pointers to the same element, right? Using UUIDs seems safer and any performance difference should be negligible.

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
dbrgn: In general I'd say that catching general exceptions is an antipattern (since specific exceptions might require specific handling), but in this case we're showing a generic error message anyways, so you're right :slightly_smiling_face:

dbrgn:
I think al comments should be addressed now.

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
ubruhin: > In theory there could be two pointers to the same element, right?

Yes there could be multiple instances (and thus different pointers) of the same netline.

> Using UUIDs seems safer

Generally yes, but in this case it doesn't really matter since you wouldn't dereference these pointers - the pointers would just be a unique identification of the netlines.

Anyway, I'm fine with `QSet<Uuid>`.

ubruhin:
> I think all comments should be addressed now.

You forgot one :)

> In addition to the comments, could you swap the order of the two new methods in bes_select.cpp to have consistent order in *.h and *.cpp?

And CI seems to be broken now ;)

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -383,6 +390,9 @@ BES_Base::ProcRetVal BES_Select::proccessIdleSceneRightMouseButtonReleased(
       menu.addSeparator();
       QAction* aSelectSegment = menu.addAction(
           QIcon(":/img/actions/bookmark.png"), tr("Select Whole Segment"));
+      menu.addSeparator();
+      QAction* aMeasureSelection = menu.addAction(
+          QIcon(":/img/actions/ruler.png"), tr("Measure Selected Segment"));
```
ubruhin: Somehow I'm not sure about this menu item text. With "Segment" actually LibrePCB means something different (see menu item above it: "Select Whole Segment"). I think this can be confusing for users. Maybe better call it something like "Measure Selected Traces", "Measure Traces Length", "Measure Length of Trace(s)"?

dbrgn:
> You forgot one slightly_smiling_face:

WTF, I did fix that, but for some reason it got lost during rebasing...

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
dbrgn: Changed to `Measure Selected Trace Length`.

dbrgn:
Hopefully CI should be fixed now. It was due to a unused-variable warning.

ubruhin:
Thanks! :+1:

PR #566, opened 2019-10-30
Change clang-format formatting rules

dbrgn:
Let me make my case for changed auto-formatter rules :slightly_smiling_face: 

Rationale: To preserve the functionality of tools like `git blame` and to make diffs easier to read, we should try not to touch any lines due to formatting rules that are not affected by the code change itself.

This is *not* about perceived beauty of the code, I think what we consider as beautiful is mostly the style we're used to. I remember finding PHP style with a line break before opening curly braces beautiful, now I hate it. I remember finding camel case method names in C# nice and well-readable, now I find them ugly. After working with a certain style for a while, it grows on you. (Well, except for Whitesmiths style. Let's ignore that for now :stuck_out_tongue_closed_eyes:)

Now a few examples on why I'd consider the current style problematic:

## Example: Horizontal alignment

img

Only a single character was changed. Can you quickly spot which one?

Besides making it easier to view diffs, not doing horizontal alignment is also a problem for `git blame` because all lines appear to be modified in a commit that did not really change them at all (besides whitespace).

And what's also important: Modifying multiple lines will increase the probability of a conflict between two PRs a lot! Not only is there a conflict, but resolving it is a pain because after conflict resolution reformatting is required. When doing this during a rebase, it will require resolving the conflict for every commit and then running the formatter again after every conflict resolution. This really is a royal pain, and git auto-merge-algorithms cannot do anything about it if a change is not confined to a single line.

## Example: Binpacking

img

A new argument was added to the function declaration. Can you quickly spot it?

Again, even though I usually like compact code, I would advise against using binpacking. If all arguments don't fit on a single line, put every argument on a single line.

(In general I also prefer to use trailing commas in argument lists, with the closing braces on a separate line, because it reduces unrelated changes even more. However, that doesn't seem to be possible with clang-format 6.)

## Next steps

@ubruhin maybe you can look through the generated changes to see if you could live with the changed configuration. It would be good if you could do the reformatting yourself, because then you retain "ownership" of changed lines in `git blame`. Feel free to push into my branch.

Reformatting the whole codebase sucks (it alters the history, it causes conflicts in open PRs, etc) but if we want to do it, then waiting will not make it easier. On the contrary, the longer we wait, the more negative impact a global code-reformatting will have.

Feedback by others is also welcome! Especially by @rnestler, who loves to repaint bikesheds! :stuck_out_tongue_winking_eye:


ubruhin:
Thanks for the detailed explanation :)

My thoughts after a quick review:
- I think I'm fine with disabling alignment of variables and comments, since the alignment affects multiple, unrelated statements.
- But somehow I don't like `AlignAfterOpenBracket: AlwaysBreak`... ;)
- I'm not yet sure about disabling bin packing. I agree it's better for diffs, but it leads to many more lines of code (+5k LOC in total!).

> @ubruhin maybe you can look through the generated changes to see if you could live with the changed configuration. It would be good if you could do the reformatting yourself, because then you retain "ownership" of changed lines in git blame. Feel free to push into my branch.

Done.

ubruhin:
Btw, with a good Git client it's no problem to see the diff even with the current settings :wink:

grafik

grafik

So in my opinion the diff is not really an issue; only the higher probability of merge conflicts is an issue...

dbrgn:
> Btw, with a good Git client it's no problem to see the diff even with the current settings

That might be true, but the diff still won't show up properly with the regular command-line git client, nor will it be rendered nicely by GitHub diffs. Those are the tools that most of the developers will use. Furthermore, `git blame` will still show the wrong commit, and – as you mentioned – the conflict potential also remains.

I do currently prefer bin packing visually, but it has too many practical downsides in my opinion. And as I wrote above, I think what we consider nice-looking can change quickly.

> Somehow I don't like AlignAfterOpenBracket: AlwaysBreak... ;)

Current (`Align`):

```c++
  connect(mManualLibraryDownload.data(),
          &LibraryDownload::progressState,
          mUi->lblDownloadZipStatusMsg,
          &QLabel::setText);
```

My suggestion (`AlwaysBreak`):

```c++
  connect(
      mManualLibraryDownload.data(),
      &LibraryDownload::progressState,
      mUi->lblDownloadZipStatusMsg,
      &QLabel::setText);
```

The third option (`DontAlign`):

```c++
  connect(mManualLibraryDownload.data(),
      &LibraryDownload::progressState,
      mUi->lblDownloadZipStatusMsg,
      &QLabel::setText);
```

I agree that `Align` looks best, but again it has the problem that renaming the method or namespace will affect all lines, not just the changed line. So it should not be used.

Regarding `DontAlign` vs `AlwaysBreak`, it seemed more consistent to me when all lines are aligned to the same column. Of course if no line break is necessary because the line is short enough, then no line break is inserted. (My preferred style would be to put the closing parentheses on its own line as well to make it look balanced, but that's not supported by clang-format :frowning_face:)

But I don't have a strong opinion on those two variants, so we could change it to `DontAlign` if you prefer that.

Of course it's also a bit dependent on bin packing. With bin packing, the `DontAlign` option might look more natural than without bin packing.

> but it leads to many more lines of code

In my view the readability is improved, which I consider much more important than lines of code.

If LOC really is important, the number of cases that require line breaks could be reduced considerably by increasing the max line width to 99. Even if you have two windows open side-by-side, 99 characters should still be easily viewable on today's screens.

I did a quick test and changed the config in this branch form 80 to 99 characters. C++ line count went from 182'113 to 175'694 (-3.5%) and a lot of lines look nicer than before:

```diff
   restoreState(s.value("mainwindow/state").toByteArray());
   mlastInputDirectory = s.value("mainwindow/last_input_directory").toString();
-  ui->edtDuplicateFolders->setText(
-      s.value("mainwindow/last_library_directory").toString());
+  ui->edtDuplicateFolders->setText(s.value("mainwindow/last_library_directory").toString());
   ui->input->addItems(s.value("mainwindow/input").toStringList());
   ui->output->setText(s.value("mainwindow/output").toString());
```

rnestler:
> That might be true, but the diff still won't show up properly with the regular command-line git client, nor will it be rendered nicely by GitHub diffs. 

To be fair for some of the cases `--ignore-all-space` should work. But not for all of them.

Regarding Binpacking: I find single argument per line *way* easier to parse with the eye since there is always one argument per line instead of 1..x
So one has either all arguments on 1 line or 1 argument per line.

dbrgn:
Comment by @ubruhin in #758:

> True :) If I remember correctly, we agreed on changing the alignment rules since everyone agrees. Afterwards we could still change more rules which will then be easier to review if alignment was already reformatted.

I'll rebase this PR accordingly.

dbrgn:
@ubruhin @rnestler branch updated!

- I removed binpacking for now (I can open a separate PR for discussions later)
- I changed `AlignAfterOpenBracket` back to `Align`

So in summary, this PR now only changes horizontal alignment of assignments (and a few other things) without touching binpacking.

I reformatted all code in the last commit, with authorship set to @ubruhin.

5n8ke:
@ubruhin sounds good! I don't have anything ready to be merged in the pipeline right now.

dbrgn:
@ubruhin rebased!

ubruhin (approved):
Thanks! :tada:

PR #571, opened 2019-11-06
CI: Create archive for deployment on Linux

ubruhin:
Currently we provide two ways to install LibrePCB binaries on Linux: Installer and AppImage. But the installer is a pain for automated installation, and the AppImage does not work out-of-the-box in Docker containers (because FUSE is missing). To avoid these issues, I think it makes sense to also provide the binaries in a *.tar.gz archive for Linux. It's basically "for free", since we just need to bundle the already available binaries in a *.tar.gz.

Interestingly I also realized that the `librepcb-cli` currently doesn't work on Linux when installed with the installer, because that binary was not patched with `linuxdeployqt`. That's also fixed with this PR, so it should now work on Linux too.

### DIFF
--- a/ci/build_linux_archive.sh
+++ b/ci/build_linux_archive.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+# set shell settings (see https://sipb.mit.edu/doc/safe-shell/)
+set -euv -o pipefail
+
+# run linuxdeployqt to bundle Qt libs
+linuxdeployqt "./build/install/opt/bin/librepcb-cli" -bundle-non-qt-libs -always-overwrite
+linuxdeployqt "./build/install/opt/bin/librepcb" -bundle-non-qt-libs -always-overwrite
+
+# copy to artifacts
+cp -r "./build/install/opt" "./artifacts/nightly_builds/librepcb-nightly-linux-x86_64"
--- a/ci/azure-jobs-linux.yml
+++ b/ci/azure-jobs-linux.yml
@@ -37,6 +37,9 @@ jobs:
   - bash: ./ci/build_linux_appimage.sh
     displayName: Build AppImage
     condition: and(succeeded(), eq(variables['DEPLOY'], 'true'))
+  - bash: ./ci/build_linux_archive.sh
+    displayName: Build Archive
+    condition: and(succeeded(), eq(variables['DEPLOY'], 'true'))
   - bash: ./ci/build_installer.sh
     displayName: Build Installer
     condition: and(succeeded(), eq(variables['DEPLOY'], 'true'))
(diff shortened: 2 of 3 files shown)


dbrgn:
> Interestingly I also realized that the librepcb-cli currently doesn't work on Linux when installed with the installer, because that binary was not patched with linuxdeployqt. That's also fixed with this PR, so it should now work on Linux too.

Cool, that probably fixes the deployment issues that @PatrickFranken had as well, right?

ci/build_linux_archive.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# set shell settings (see https://sipb.mit.edu/doc/safe-shell/)
+set -euv -o pipefail
+
+# run linuxdeployqt to bundle Qt libs
+linuxdeployqt "./build/install/opt/bin/librepcb-cli" -bundle-non-qt-libs -always-overwrite
+linuxdeployqt "./build/install/opt/bin/librepcb" -bundle-non-qt-libs -always-overwrite
+
+# copy to artifacts
+cp -r "./build/install/opt" "./artifacts/nightly_builds/librepcb-nightly-linux-x86_64"
```
dbrgn: How are these bundled into a `.tar.gz`? Is that handled by Azure somehow?

ubruhin: That's done on our server when the uploaded artifacts are processed ;)

ubruhin:
> Cool, that probably fixes the deployment issues that @PatrickFranken had as well, right?

Actually I don't know (yet), but if he used the installer, then probably yes :)

Btw, I forgot to mention that the new archive can be "reviewed" here: https://download.librepcb.org/nightly_builds/ci-build-linux-archive/librepcb-nightly-linux-x86_64.tar.gz

dbrgn:
> Actually I don't know (yet), but if he used the installer, then probably yes :)

No, the AppImage, which segfaulted on his server...

ubruhin:
> The archive currently contains a root directory called .. This results in the files being extracted into the current directory, without a containing root directory.

Hmm yeah that was indeed strange...

> It would be good if the archive contained a root directory called librepcb or librepcb-nightly-linux-x86_64.

Actually I don't like having a root directory in an archive (the purpose of an archive is to bundle multiple files into one file, so why move them all into a useless subdirectory). But (unfortunately) it seems to be common practice to do so, thus I added the root directory `librepcb-nightly-linux-x86_64` and now need to adjust the release scripts to replace "nightly" inside the *.tar.gz with the actual version number...

Or is it also accepted to put the actual files into the root of a tar archive?

dbrgn:
> Or is it also accepted to put the actual files into the root of a tar archive?

No, a root directory is more common. If you don't want the root dir, you can use `--strip-components=1` when unpacking. However, if you accidentally unpack an archive without a root directory, you got files all over the place...

ubruhin:
OK the archive now contains the root directory `librepcb-nightly-linux-x86_64`.

PR #573, opened 2019-11-09
ControlPanel::projectEditorClosed: Use toUnique on file path

### LINKED ISSUE #432: Segfault when re-opening project

```
LibrePCB Version: 0.1.0-unstable
Git Revision:     bc82905bb1b9abbee9ac4fc79018122e07a2c1be
Build Date:       2019-03-17T16:37:56
Qt Version:       5.12.1 (built against 5.12.1)
CPU Architecture: x86_64
Operating System: Arch Linux
Platform Plugin:  xcb
```

##### SUMMARY

When closing and re-opening a project, on re-opening the project LibrePCB segfaults.

##### STEPS TO REPRODUCE

Possibly this has to do with the user settings file.

Try this:

    echo "(librepcb_board_user_settings)" > boards/default/settings.user.lp

Then open the project, close it again and re-open it.

dbrgn:
Because the `toUnique` call was missing, projects on a symlinked path
were not removed from the `mOpenProjectEditors` hash map. This in turn
would trigger a segfault when reopening a project due to a
use-after-free.

Fixes #432.

### DIFF
--- a/apps/librepcb/controlpanel/controlpanel.cpp
+++ b/apps/librepcb/controlpanel/controlpanel.cpp
@@ -451,7 +451,9 @@ void ControlPanel::projectEditorClosed() noexcept {
   if (!editor) return;
 
   Project* project = &editor->getProject();
-  mOpenProjectEditors.remove(project->getFilepath().toStr());
+  Q_ASSERT(mOpenProjectEditors.contains(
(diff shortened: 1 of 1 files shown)


ubruhin:
Nice catch, thanks! :smiley:

PR #581, opened 2019-11-16
Board editor: Add initial design rule check

### LINKED ISSUE #539: Board editor: Add DRC (Design Rules Check)
##### SUMMARY
In Board Editor, it seems that net clearance constraints are missing.
:fire::fire::fire: I succeeded in connecting VCC net to GND net ! :fire::fire::fire: 

##### SOLUTION
Implement net clearance, if possible in live.

(Think this is a main pcb software feature that is missing, and so it needs a high priority)

ubruhin:
Actually I planned to implement only a very simple DRC for now, but it became much better than expected :grin:

- Add "Design Rule Check" dialog to the board editor, to run the DRC.
- Add "DRC" dock widget to the board editor, to see the messages even after closing the dialog. By clicking on a message, the issue is highlighted and centered on the board view.

drc

Limitations:
- The "missing connections" check does not really check the copper layers, but only emits a message for each existing airwire. If there is a false-positive airwire, you will also get a false-positive DRC message.
- Settings (like clearance) are not yet permanently saved in the board file (needs file format change, will do in the next major release).
- Messages cannot be suppressed yet (needs file format change, will do in the next major release).
- Not yet integrated into the CLI (does not make sense as long as settings and suppressions are not saved).
- Not yet tested intensively, i.e. there might still be some bugs.

Fixes #539


EdizonTN:
Perfect!
Some comments:

> Add "DRC" dock widget to the board editor, to see the messages even after closing the dialog. By clicking on a message, the issue is highlighted and centered on the board view.

DRC settings and Run DRC should be separate dialog.
Why? - DRC settings are valid for design rule check but for gerber output (export) also!

> the issue is highlighted and centered on the board view.
- the issue is highlighted **OR** centered on the board view.

ubruhin:
> DRC settings and Run DRC should be separate dialog.
> Why? - DRC settings are valid for design rule check but for gerber output (export) also!

I agree - thus there is still the separate dialog "Design Rules" in the "Board" menu. But since the new design rules are not yet saved in files, you would have to check/adjust the design rules every time after opening a project, and afterwards opening the DRC dialog to run the checks. I think this is too cumbersome, thus I added the new design rules to the new dialog instead.

We can fix this issue once design rules are really saved to files, i.e. moving the new options into the "Design Rules" dialog.

>  the issue is highlighted OR centered on the board view.

I'm not sure what you mean with that, could you explain it?

> Settings should be saved into new separate file. Settings is manufacturer depends. Settings file can have name as manufacturer or PCB constuction density.
> This file can by in separate library - for sharing between users.

Generally I agree - but it's not the topic of this PR since the new options aren't saved to files at all. As explained above, this will be the topic of the next major release ;)

EternityForest:
This is amazing! Looks like LibrePCB is so close to being the premier FOSS editor!

I'm not sure checking the copper for disconnection is actually needed. 

IIRC eagle doesn't have "disconnected" DRC at all, they just directly call it "airsides".  If you have a false positive airwire, you treat it like a compiler warning and fix it even if you know it's a false positive.

EdizonTN:
> > the issue is highlighted OR centered on the board view.
> 
> I'm not sure what you mean with that, could you explain it?

I mean some option for select way of show error on board.
Option 1: Highlight - when  I click on error in DRC panel, error blikn (or highlight) on board without zoom. This option is good for locate error on board. Ideal for big board design.

Option 2: Zoom (jump) - when  I click on error in DRC panel, error will zoomed on board to center to screen. This option is good for show directly error. Zoomed part of PCB is litlle bit lost on whole design.

Example (highligh vs zoom/jump):
drc

p.s. in this example, highlight mode is not ideal (and this is commercial, not cheap software). It's hard to find on bigger design this small error (even highlighted).

And next question: Is any way howe to I find which error just see on PCB on screen (reverse way as you present on video from start of this thread)? 
For example: when I move cursor over error mark on PCB, in status bar I see error description for this mark or error is highlighted in DRC toolbox. Can be applied this?

dbrgn:
Very cool! I ran it on my current project, and it found this gem :sweat_smile: 

short

So, thanks for saving me some time and hardware costs.

Some of the errors are a bit strange though:

triangle

Why this irregularly shaped triangle? I assume the algorithm approximates some irregular shapes with rough triangles, right?

ubruhin:
> I'm not sure checking the copper for disconnection is actually needed.

IMHO that's one of the most important checks, since a missing connection could make a board useless. In my experience it's easy to forget checking the airwires before generating the fabrication data, especially if there are short airwires which you could overlook, or if the airwires layer is hidden.

> I mean some option for select way of show error on board.
> Option 1: Highlight - when I click on error in DRC panel, error blikn (or highlight) on board without zoom. This option is good for locate error on board. Ideal for big board design.
>
> Option 2: Zoom (jump) - when I click on error in DRC panel, error will zoomed on board to center to screen. This option is good for show directly error. Zoomed part of PCB is litlle bit lost on whole design.

Good idea! I implemented it with a checkbox in the DRC dock. And a double-click on a message item will always zoom to the corresponding location, independent of the checkbox.

> And next question: Is any way howe to I find which error just see on PCB on screen (reverse way as you present on video from start of this thread)?
> For example: when I move cursor over error mark on PCB, in status bar I see error description for this mark or error is highlighted in DRC toolbox. Can be applied this?

No that's currently not possible. And it doesn't make much sense since only a single DRC violation is shown in the board editor (instead of all at the same time). We might change that behavior in future...

> Why this irregularly shaped triangle? I assume the algorithm approximates some irregular shapes with rough triangles, right?

An expanded rectangular pad results in a rounded rectangular, but ClipperLib (which I use for clearance checks) doesn't support arcs, thus I had to approximate arcs by straight line segments (like we already do for copper planes).

Btw, now I also added a check for overlapping courtyards of packages, with a user-defined additional offset. And I moved the missing connection messages to the bottom since during the design phase these messages are a bit annoying.

board_editor_drc

EternityForest:
In Eagle at least, an Airwire *is* a DRC issue that shows in the same list,
so it's pretty hard to forget them, and easy to find any short ones you
left behind, but without that, yeah, I could easily see forgetting.

dbrgn:
What's this white line when selecting a courtyard clearance error in the DRC? A leftover from debugging?

2019-11-24-234211_1123x823_scrot

doom-fr:
I think it is to highlight the error place.
Without this it will be hard to see the location of the problem.

ubruhin:
Yes, that's to easily locate even very small DRC issues independent of the zoom level. Some other EDA tools also do it like this, thus I just did it the same way :wink:

dbrgn:
Ok, never saw this before but seems reasonable for now. I think there could be better ways to visualize, but that can still be done later on :slightly_smiling_face:

doom-fr:
Is it possible to have a board view with all DRC ?
It will be helpful to solve problems when you have multiple DRC in same area.

ubruhin:
> Is it possible to have a board view with all DRC ?

No that's currently not possible. But we can add it in a later release.

doom-fr:
Yes sure the actual DRC is already more that just a simple DRC. Thanks for that.

As improvement, I would say we can :
1. view all DRC in the board editor (maybe via DRC layer)
2. make live DRC (DRC change each time we move something or draw something)
3. add component spacing clearance ;-)

PR #583, opened 2019-11-17
CLI: Add "--strict" option

ubruhin:
With the `--strict` option, the CLI reports files which are not canonical, i.e. files which would be modified when saving the opened project or library element. Useful for continuous integration of libraries, mainly to check library elements which were generated (i.e. not made with LibrePCB).

Example:

```bash
$ librepcb-cli open-library LibrePCB_Connectors.lplib --all --strict
Open library 'LibrePCB_Connectors.lplib'...
    - Non-canonical file: LibrePCB_Connectors.lplib/library.lp
Process 0 component categories...
Process 0 package categories...
Process 198 symbols...
Process 514 packages...
Process 198 components...
Process 514 devices...
    - Non-canonical file: LibrePCB_Connectors.lplib/dev/0b53aefa-f8fb-4f31-bfa2-24325a58428b/.librepcb-dev
    - Non-canonical file: LibrePCB_Connectors.lplib/dev/0b53aefa-f8fb-4f31-bfa2-24325a58428b/device.lp
Finished with errors!
```

### DIFF
--- a/apps/librepcb-cli/commandlineinterface.cpp
+++ b/apps/librepcb-cli/commandlineinterface.cpp
@@ -131,6 +131,10 @@ int CommandLineInterface::execute() noexcept {
   QCommandLineOption saveOption(
       "save",
       tr("Save project before closing it (useful to upgrade file format)."));
+  QCommandLineOption prjStrictOption(
+      "strict", tr("Fail if the project files are not strictly canonical, i.e. "
+                   "there would be changes when saving the project. Note that "
+                   "this option is not available for *.lppz files."));
 
   // Define options for "open-library"
   QCommandLineOption libAllOption(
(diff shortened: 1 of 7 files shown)


rnestler:
> Please, what do you mean with "canonical / non-canonical" files?

Well @ubruhin mentioned it above: "files which would be modified when saving the opened project or library element".

So other ways to say it may be "non-standard" or "not properly formatted" / "ill-formatted". I guess it is mostly about order and whitespace issues, right @ubruhin?

EdizonTN:
OK, few next questions:
It's, error, warning or notice (LibrePCB can accept this file)? 
Can LibrePCB repair these file or not?
It needs some user's action (ex. repair manually)?
Word "canonical" - it's standard for this type of message? 

Canonical are occasionaly used in folder/file path string before are unified but in file format.... I don't know (I don't met until now).
Proposals: "corrupted", "bad integrity", "file changed by another process", "bad format"... 

Thanks.

_Ed

dbrgn:
A non-canonical file is not wrong, it just means that LibrePCB would format it differently. This means that if you change something in a library element that would just affect a single line, the whole file will be reformatted, causing a lot of changes and a large diff in version controlled files.

When you generate your own files for private use, non-canonical files are not a problem at all. When contributing files to the standard library, they should be in canonical format.

EdizonTN:
.... a Genuine LibrePCB file  :-)

Ok, ok,  thanks for an explanation.

PR #613, opened 2019-12-01
Update appdata file

dbrgn:
It now matches the one generated in the flathub package.

Should be added to the release checklist. (Is that checklist inside the repo?)


share/metainfo/org.librepcb.LibrePCB.appdata.xml
```diff
@@ -64,5 +55,9 @@
     <binary>librepcb-cli</binary>
   </provides>
 
+  <releases>
+    <release version="0.1.3" date="2019-11-30"/>
```
ubruhin: Hmm since the tag is called `releases`, I wonder whether we should actually list *every* release instead of just the most recent one?

ubruhin:
> Is that checklist inside the repo?

Yes: https://github.com/LibrePCB/LibrePCB/blob/master/dev/doxygen/pages/release_workflow.md

share/metainfo/org.librepcb.LibrePCB.appdata.xml
dbrgn: You're right! https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases

I'll update.

dbrgn:
@ubruhin updated

ubruhin:
Nice, thanks! :+1:

Will you also update https://github.com/flathub/org.librepcb.LibrePCB/pull/5 accordingly? If yes, I'd wait with merging this to see if the Flatpak build succeeds.

dbrgn:
Can do!

ubruhin:
It might be a bit strange to insert an URL to the blogpost *before* the blogpost gets published, but OK, let's try it that way when creating the next release :wink:

Thanks @dbrgn!

dbrgn:
Ah, we can always add partial entries on publication and add more details later on!

But I think in this case we already know the URL in advance.

PR #623, opened 2019-12-05
README: Add BSD to the Features section

yurivict:


README.md
```diff
@@ -19,7 +19,7 @@ about the currently available features, limitations and known bugs.
 
 ### Features
 
-- Cross-platform (Unix/Linux, Mac OS X, Windows)
+- Cross-platform (Unix/Linux/FreeBSD, Mac OS X, Windows)
```
dbrgn: The compatibility is not specific to FreeBSD, right?

Also, if you're at it, you could fix the macOS name :slightly_smiling_face: 

```suggestion
- Cross-platform (Unix/Linux/BSD, macOS, Windows)
```

yurivict:
> The compatibility is not specific to FreeBSD, right?

It is specific to FreeBSD. The port is only available on FreeBSD, and not on other BSDs.

dbrgn:
> It is specific to FreeBSD. The port is only available on FreeBSD, and not on other BSDs.

Yes, but the "cross-platform" thing is mostly about API compatibility. If you compile it yourself, it will most probably run on OpenBSD or NetBSD as well. LibrePCB isn't packaged for every Linux distribution, but it doesn't make sense to list them all when listing supported platforms.

I agree that FreeBSD should be explicitly listed on the [download page](https://librepcb.org/download/) (which it is). (Although I just noticed that it's in the "Linux" category, which is probably wrong :slightly_smiling_face:)

dbrgn:
> Although I just noticed that it's in the "Linux" category, which is probably wrong

Fixed: https://github.com/LibrePCB/LibrePCB.github.io/pull/38

README.md
ubruhin: Since there is no reaction on this suggestion, I'll apply it by myself ;)

Thanks @yurivict for updating the README!

ubruhin: Hmm it seems I have no permissions to push to this branch, so I can't apply the suggestion.

@yurivict could you either apply the suggestion or grant me write permissions on this branch?

yurivict:
@ubruhin I updated the PR as suggested.

ubruhin:
Perfect, thanks! :+1:

I'll squash merge the commits since the branch is based on a quite old revision of `master`.

Issue #631, opened 2019-12-15
Board Editor: Prevent error "Nothing to connect."


mrm2m:
##### SUMMARY

If you miss the trace which you want to draw to by just a few mm you will see the message "Nothing to connect".

##### SOLUTION

Find out if there are trace nodes in reasonable distance (like 1cm) around the click position and select the one which is nearest if there is no direct match.

miniwark:
I have a similar case of "nothing to connect error", with a handmade component, even if the wire was apparently at the same place than the connection pin. I suspect than the pin was 1 or 2 pixels away from the grid, and the trace try to "snap to grid" instead of "snap to pin".

Also., I have solved the issue by recreating the component from start in the library. Wile doing this, there seem to have a steep, in the Wizard that not editable in the library: In the "Pin-Signal-Map" i have clearly indicated "pin" int the Designator in Schematics" column, and suspect then this steep is important to connect correctly in the schematics (but i do not have find a way to edit this in the library).

5n8ke:
Simply choosing the closest NetPoint would not suffice in my opinion. Pads, Vias, and even the closest distance to the traces must be taken into consideration.
And should the currently selected layer for routing be considered as well?

PR #632, opened 2019-12-15
Select the next reasonable node with "Draw trace" tool

mrm2m:
If the "Draw trace" tool is active the next reasonable trace node is selected to
connect the new trace to it. #631

### DIFF
--- a/libs/librepcb/project/boards/items/bi_netpoint.h
+++ b/libs/librepcb/project/boards/items/bi_netpoint.h
@@ -75,6 +75,7 @@ public:
   // General Methods
   void addToBoard() override;
   void removeFromBoard() override;
+  UnsignedLength distance(Point destination);
 
   /// @copydoc librepcb::SerializableObject::serialize()
   void serialize(SExpression& root) const override;
--- a/libs/librepcb/project/boards/items/bi_netsegment.cpp
+++ b/libs/librepcb/project/boards/items/bi_netsegment.cpp
@@ -194,6 +194,24 @@ int BI_NetSegment::getViasAtScenePos(const Point&    pos,
   return count;
 }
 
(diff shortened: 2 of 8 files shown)


libs/librepcb/project/boards/items/bi_netsegment.cpp
```diff
@@ -194,6 +194,24 @@ int BI_NetSegment::getViasAtScenePos(const Point&    pos,
...
+  foreach (BI_NetPoint* netpoint, mNetPoints) {
+    if (netpoint->isSelectable() && ((!layer) || (netpoint->getLayerOfLines() == layer))) {
+      UnsignedLength distance = netpoint->distance(pos);
+      if (! maxDistance || distance < *maxDistance) {
+        *maxDistance = distance;
+        *bestMatch = netpoint;
+      }
+    }
+  }
+  return 0;
```
ubruhin: This looks like the return value is actually not needed :wink: Maybe better return `bestMatch` instead of using an out-parameter for it? Or least change the return type to `void`.

libs/librepcb/project/boards/items/bi_netpoint.h
```diff
@@ -75,6 +75,7 @@ class BI_NetPoint final : public BI_Base,
   // General Methods
   void addToBoard() override;
   void removeFromBoard() override;
+  UnsignedLength distance(Point destination);
```
ubruhin: According our naming conventions, the method should be named `getDistance()`, or `getDistanceTo()` to be more expressive. But actually to be honest, I'd simply remove this method since it's trivial for callers to use `getPosition()` instead and calculate the distance to the other point "manually" since it's trivial :wink:

ubruhin:
Due to missing response in this PR, @5n8ke has taken over this feature and applied the suggestions in #753. So this PR can be closed.

Issue #630, opened 2019-12-12
BOM CSV Format


dbrgn:
##### VERSION / OS / ENVIRONMENT

```
LibrePCB Version: 0.1.3
```

##### SUMMARY

Both [GitHub](https://help.github.com/en/github/managing-files-in-a-repository/rendering-csv-and-tsv-data) and Digikey cannot handle CSV files that are semicolon separated.

It would be better if BOM export would use commas to separate fields, with quoting of all fields that actually contain commas.

##### STEPS TO REPRODUCE

- View [this BOM](https://github.com/gfroerli/hardware-v2/blob/2d5ac09174473047031ac4ce0273989ec1938180/output/v2/BOM.csv) on GitHub
- Upload the BOM linked above to Digikey

dbrgn:
With commas, the GitHub preview looks [like this](https://github.com/gfroerli/hardware-v2/blob/master/output/v2/BOM.csv).

Issue #621, opened 2019-12-04
Implement device placement file export for pick&place machines


doom-fr:
##### SUMMARY
Pick and Place machines need a file with position and angle of all devices on the board.
This file is currently not generated by the fabrication data generation.

##### SOLUTION
Generate the file during fabrication data generation.
If there is components on bottom, another file should be generated.

##### ADDITIONAL DATA

The file looks like that : (example generated by Kicad)
```
# Ref     Val       Package     PosX        PosY     Rot         Side
C11       22uF      0805        51.2000    -22.60      0.0000    top 
C12       22uF      0805        56.1000    -32.00    270.0000    top 
C13       100nF     0805        52.0000    -25.70    270.0000    top 
C14       100pF     0805        51.0000    -28.80      0.0000    top 
C21       47nF      0805        33.9600     -7.87    180.0000    top 
C22       220nF     0805        36.2500    -11.37    270.0000    top 
C23       22uF      0805        36.1000    -25.40    270.0000    top 
C24       2.2uF     0805        33.8000    -21.00    180.0000    top 
```

EdizonTN:
I absolutly agree with idea to generate placement data, but i have some notices.

Placement of the components (ex. reflow soldering) is not a part for fabrication process. It's a part of assembly process, and can be done in another company as fabrication. And into assembly house you will not send gerber data of course... :-)
Due this, is better, creating Pick&Place data package with BOM and Assembly prints, as independend process.

My proposal is create new Menu->File item "Generate Assembly data" which open a new window with options to export:
- BOM 
- PDF printout of Top Assembly
- PDF printout of Bot Assembly
- PDF printout of additional notification text (specialities in assembly process). All printouts can be in one PDF file, each on new side.
- Generate Pick&Place data (csv file)

Output path as "./output/{{VERSION}}/Assembly/{{PROJECT}}"

And Also Fabrication path will be good to little bit modified as "./output/{{VERSION}}/Fabrication/{{PROJECT}}"

And please thinking also, about option to saving THT and SMT devices to separate files.


Attachment - Another example of Pick&place .csv file content:
"Designator","Comment","Layer","Footprint","Center-X(mm)","Center-Y(mm)","Rotation","Description"
"IC12","PCA9955B","BottomLayer","HTSSOP65P640X970-28N_-_SOT1172-3_-_MO153","2.857","8.255","90","16-Channel LED CC driver with PWM current controller"
[...]

PR #633, opened 2019-12-15
Change BOM export CSV format

mrm2m:
As described in Issue #630 the CSV format with ';' and without string separators is not optimal for many use cases.

### DIFF
--- a/libs/librepcb/common/bom/bomcsvwriter.cpp
+++ b/libs/librepcb/common/bom/bomcsvwriter.cpp
@@ -50,13 +50,18 @@ QList<QStringList> BomCsvWriter::toStringList(const Bom& bom) noexcept {
   QList<QStringList> lines;
   // Don't translate the CSV header to make BOMs independent of the user's
   // language.
-  lines.append(QStringList{"Quantity", "Designators"} + bom.getColumns());
+  QStringList header = QStringList{"Quantity", "Designators"} + bom.getColumns();
+  QStringList headerCols;
+  foreach (const QString& headerItem, header) {
+    headerCols += '"' + headerItem + '"';
+  }
+  lines.append(headerCols);
   foreach (const BomItem& item, bom.getItems()) {
     QStringList cols;
     cols += QString::number(item.getDesignators().count());
-    cols += cleanStr(item.getDesignators().join(", "));
+    cols += '"' + cleanStr(item.getDesignators().join(", ")) + '"';
     foreach (const QString& attribute, item.getAttributes()) {
-      cols += cleanStr(attribute);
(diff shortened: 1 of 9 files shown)


libs/librepcb/common/bom/bomcsvwriter.cpp
```diff
@@ -50,13 +50,18 @@ QList<QStringList> BomCsvWriter::toStringList(const Bom& bom) noexcept {
...
+  QStringList headerCols;
+  foreach (const QString& headerItem, header) {
+    headerCols += '"' + headerItem + '"';
+  }
+  lines.append(headerCols);
   foreach (const BomItem& item, bom.getItems()) {
     QStringList cols;
     cols += QString::number(item.getDesignators().count());
-    cols += cleanStr(item.getDesignators().join(", "));
+    cols += '"' + cleanStr(item.getDesignators().join(", ")) + '"';
```
dbrgn: Will `cleanStr` ensure that all quotes inside the string are escaped? Otherwise we'll end up with broken CSV files if any of the strings contain quotes.

It might also be nice if double quotes are only added when necessary (i.e. when the value contains a comma).

ubruhin:
Thanks @mrm2m!

As @dbrgn mentioned, quotes inside the strings need to be escaped, otherwise invalid CSV might be generated. And actually we should refactor the `BomCsvWriter` anyway because it's a bit hacky and for #621 we need a CSV file writer too, therefore I started to implement a generic CSV writer class which can then be used for both, BOM export and pick&place export.

Thus I'd like to wait with this PR until the new CSV writer is ready to use. Would you be interested to do the refactoring of `BomCsvWriter` afterwards?

ubruhin:
> Thus I'd like to wait with this PR until the new CSV writer is ready to use.

Done - there is now a [`CsvFile`](https://github.com/LibrePCB/LibrePCB/blob/master/libs/librepcb/common/fileio/csvfile.h) class which uses comma as separator and handles escaping and quoting properly. In `PickPlaceCsvWriter` you can see how it is intended to use:

https://github.com/LibrePCB/LibrePCB/blob/09250c87be500cb09603cbe423ddb43d5dfcab5d/libs/librepcb/common/pnp/pickplacecsvwriter.cpp#L50-L99

ubruhin:
@mrm2m Any chance to get this updated soon? Or should we take over this PR? Actually now with the `CsvFile` class it should be very easy since you no longer have to care about escaping and quoting.

ubruhin:
Due to missing activity in this PR, I created #678 to implement the refactoring of `BomCsvWriter` -> closing this PR.

PR #638, opened 2019-12-26
Library Editor: Support moving elements to other library

ubruhin:
grafik

This is especially useful to move library elements from your own library to an official library, i.e. following workflow:
1. Add all new elements to your own library first
2. Use these elements in your PCB, manufacture and test the PCB
3. If needed, fix issues in the library elements
4. Move library elements to the corresponding official library
5. Open GitHub pull request to include your library element

Until now, step 4 was possible only by manually moving files around between the libraries. This PR simplifies this by integrating the action into the library editor.


ouabache:
Does it modify the UUID's ?

What happens if you try to load both libraries at the same time?

dbrgn:
Awesome, makes the workflow much easier :slightly_smiling_face: 

> What happens if you try to load both libraries at the same time?

If the version is identical, the element will be picked randomly from one of the libraries. That's fine if there are no changes. Otherwise, bump the version, then LibrePCB will pick the library element with the highest version.

ubruhin:
> Does it modify the UUID's ?

No, basically only the directory of the library element is moved to the other library, i.e. UUIDs are not modified.

> What happens if you try to load both libraries at the same time?

You mean, if the editors of both libraries are opened? That's no problem, the element will be closed automatically in the old library and after the library scan finished, the element appears in the new library.

ouabache:
And what happens the next time you run librePCB? It no longer knows which
one was old or new. How does it choose the correct one?

If it finds one in a remote lib and a duplicate in local then it should
choose local because that means you are editing it and need to be able to
test your edits.

ubruhin:
@ouabache I don't understand what you mean exactly. The library element is *not copied*, but *moved* to another library. So there is no issue with duplicate, old vs. new, version number or something else. Everything works exactly the same way as before moving, the element is just located in a different library.

ouabache:
OK, I was thinking about a copy

PR #649, opened 2020-01-04
Change macOS default target directory

### LINKED ISSUE #644: Install location on mac
This is for the mac platform only:
The default install location for apps should be /Applications
Not in a user's App folder like ~/"user"/Application
Users can override this default if need be, but almost 100% of apps usually go to the system wide /Applications location

dbrgn:
Instead of `~/Applications/`, use the system-wide `/Applications/` directory.

Fixes #644.

I haven't tested this. Will opening a PR automatically result in the branch being built in CI, or do we have to merge to master in order for a nightly to be created?

(I can test this on a macOS machine next week, but don't have any build toolchain on there to build the installer manually.)

### DIFF
--- a/dist/installer/update_metadata.sh
+++ b/dist/installer/update_metadata.sh
@@ -35,7 +35,7 @@ then
   DEFAULT_TARGET_DIR="@homeDir@/LibrePCB"
 elif [[ "$TARGET_NAME" == mac* ]]
 then
-  DEFAULT_TARGET_DIR="@homeDir@/Applications/LibrePCB"
+  DEFAULT_TARGET_DIR="/Applications/LibrePCB"
 fi
 
 # copy to destination directory


spookydd:
Can I help testing something?
I'm on mac and I could do this right away. No need to wait for next week or whatever.
Just tell me what to get and what to do.

dist/installer/update_metadata.sh
```diff
@@ -35,7 +35,7 @@ then
   DEFAULT_TARGET_DIR="@homeDir@/LibrePCB"
 elif [[ "$TARGET_NAME" == mac* ]]
 then
-  DEFAULT_TARGET_DIR="@homeDir@/Applications/LibrePCB"
+  DEFAULT_TARGET_DIR="/Applications/LibrePCB"
```
ubruhin: I just saw that there is also a [predefined variable](https://doc.qt.io/qtinstallerframework/scripting.html#predefined-variables) available for the applications directory:

```suggestion
  DEFAULT_TARGET_DIR="@ApplicationsDir@/LibrePCB"
```

ubruhin:
The installer built on CI can be downloaded here to test it: https://download.librepcb.org/nightly_builds/644-macos-default-target-dir/librepcb-installer-nightly-mac-x86_64.dmg

Note that it doesn't download the nightly LibrePCB application, but the official stable release. But this shouldn't matter, this change should still take effect.

spookydd:
Alright then. Just tried it. Problem solved!
Yippee!

spookydd:
Oh, and by the way: There is the matter of the unidentified developer.
Is this something hard to fix?

dist/installer/update_metadata.sh
dbrgn: Ah, nice find!

ubruhin:
@spookydd since there was one more change after you tested the installer, could you maybe quickly test it again to check if it still works? The link above now points to the updated installer.

> Oh, and by the way: There is the matter of the unidentified developer. Is this something hard to fix?

Probably Apple wants to see some money to allow us creating a signed installer, but I don't know how it works exactly...

spookydd:
Ok, just tried it again and it's back to broken. So the previous one worked fine.

I had no idea how this thing about identified developer works, but I would't put it past apple to make it a matter of money. They're expert at separating people from their money.
So I guess there is no way to prevent that install blocking every time then, it has to be overridden every single time...

dbrgn:
> Ok, just tried it again and it's back to broken

Can you explain what "broken" means? Which directory is being used?

spookydd:
Oh I'm sorry, I wasn't specific.
By broken again, I meant it went back to its own prior bad behavior, which defaults to the user's Applications folder instead of the system wide one. The behavior earlier was right, but now it's back to wrong.
Is there any chance soon to have the App do its own updates internally? That is automatically according to a preference setting to tell it how to handle it (like the majority of apps ).

spookydd:
btw: about that developer identification thing, I think there may be a somewhat better solution that may not involve giving any money to apple. I updated libreoffice lately and noticed the behavior is a tad different, although we still get the message about unidentified developer, there is no need to go to the control panel to tell it to allow it and then again to click on open in the popup that follows.
I'm thinking right now librepcb isn't signed at all, and libreoffice may be signed, although perhaps not having paid to be "identified", the signature, even though not fully official, causes a simpler procedure with a popup with an open button, so no need to open the control panel.
Maybe just sign the app in some way, not via a paid for solution, just sign it for free..

ubruhin:
Oops, it seems that's a [bug in the Qt installer framework](https://bugreports.qt.io/browse/QTIFW-1011) :sob: Fortunately @spookydd tested it again, I really didn't expect at all that the added commit breaks something!

Sorry @dbrgn, looks like you need to revert my suggestion :grin:

dbrgn:
Is the installer framework 4 not yet released?

ubruhin:
> Is the installer framework 4 not yet released?

https://lmgtfy.com/?q=qt+installer+framework+releases&s=d

SCNR :grin:

dbrgn:
Yeah, it was late and I was too lazy to DDG :stuck_out_tongue:

dbrgn:
Reverted.

ubruhin:
Thanks!

PR #650, opened 2020-01-04
Doxygen: Add database schema diagram

dbrgn:
Generated with [schemacrawler](https://www.schemacrawler.com/).

img


ubruhin:
Nice, thanks! :+1:

PR #652, opened 2020-01-08
Bump version to 0.2.0-unstable and add workspace migration

ubruhin:
To allow implementing file format changes, we need to bump the file format version to 0.2 and the application version to 0.2.0-unstable. But several other changes were needed in addition to just bumping the version:

## Add "Initialize Workspace" wizard

If you open a workspace with LibrePCB 0.2.x the first time, a wizard now helps you to initialize/migrate the workspace to the new file format version:

grafik

When choosing a version to import from:

grafik

Or when choosing not to import any data:

grafik

## Adjust first-run-wizard

Since the new wizard is now used to initialize a workspace, the "Settings" page was removed from the first-run-wizard. That wizard now only lets you choose the workspace, but does no longer initialize it.

## Enable warning about unstable file format

From now on, the `master` branch must not be used for productive work since the file format is unstable. Thus the following warning now appears when starting an unstable version of LibrePCB:

grafik

## Make CLI fail when trying to save files from unstable version

Same as in LibrePCB, it would be dangerous if files are saved with an unstable CLI version.

---

Since this is a very important change, it would be helpful to get some feedback!


spookydd:
Would it be helpful if I could also run this unstable version on mac?
Could it be run in parallel with the stable one?
I'm not doing any real project for now, so it's no trouble if something isn't working right, and if something does, then it's an opportunity to find out about it and then work on that.

ubruhin:
@spookydd I'm not sure if we are talking about the same thing ;) This PR does not add any new "real" feature, it only bumps the version number and introduces a workspace migration wizard. This wizard is shown only the first time when opening a 0.1.x workspace, so it's only the migration itself which could be tested. But yes, of course you could test that migration if you like :)

dbrgn:
Maybe I'll find time to do some testing on Sunday.

dbrgn:
Oh, sorry for not testing this, I totally forgot about the PR :slightly_smiling_face: 

I'll test it in the context of #510.

Issue #67, opened 2016-03-21
Strictly separate editor code from core logic (architecture issue)


ubruhin:
Some editor related code (`librepcbprojecteditor`) is currently integrated into the logic core classes (`librepcbproject`), which is a little bit ugly. We should fix this architecture issue someday.

Example:
The core class `NetSignal` has a property `isHighlighted` which is used to highlight items of a specific net signal in the schematic/board editors. As this is the only purpose of that property, it should be located in `librepcbprojecteditor` instead of `librepcbproject`.

PR #656, opened 2020-01-20
Add SVG export to schematic-/board editors

### LINKED ISSUE #345: Allow to export schematics as picture
##### SUMMARY
Currently the only means of exporting schematics are via PDF, which is very useful, but it would also be very beneficial to be able to export it as an image.

##### SOLUTION
Ideally to be able to export in both PNG and SVG, so that we have both a graphical format and a vectorial format.

ubruhin:
Allows to export schematics and boards as SVG files.

Unfortunately it adds the dependency to qt5-svg (probably only relevant on Linux), thus I'm not sure if this should be postponed to the next major release...

Fixes #345, although it does not provide *.png export yet. But with graphics software it's easy to convert the *.svg to *.png with desired resolution.


dbrgn:
Nice!

Some things I noticed:

- The yellow rectangles for a symbol appear above the text. The text is still readable due to alpha transparency of the symbol rectangle, but the text is grey instead of black and it cannot be selected without moving the symbol rectangle to the bottom. Not a big issue, but maybe it can be solved easily?

2020-01-21-112413_779x414_scrot

- The "grab crosses" are exported in the SVG:

2020-01-21-112319_675x513_scrot

2020-01-21-112657_387x306_scrot

That's probably unnecessary?

- In the board export, there are a *lot* of empty groups:

```xml
<g fill="none" stroke="#000000" stroke-opacity="1" stroke-width="1" stroke-linecap="square" stroke-linejoin="bevel" transform="matrix(0,3.52778,3.52778,0,684.774,392.282)"
font-family="Sans Serif" font-size="31.75" font-weight="400" font-style="normal"
>
</g>
```

They probably increase the file size considerably. (My export is 1.4 MiB.)

libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
```diff
@@ -533,6 +535,53 @@ void BoardEditor::on_actionExportAsPdf_triggered() {
...
+                                FilePath::ReplaceSpaces | FilePath::KeepCase);
+    QString projectVersion =
+        FilePath::cleanFileName(mProject.getMetadata().getVersion(),
+                                FilePath::ReplaceSpaces | FilePath::KeepCase);
+    QString relativePath =
+        QString("output/%1/%2_Board.svg").arg(projectVersion, projectName);
+    FilePath defaultFilePath = mProject.getPath().getPathTo(relativePath);
+    QDir().mkpath(defaultFilePath.getParentDir().toStr());
+    QString filename = FileDialog::getSaveFileName(
+        this, tr("SVG Export"), defaultFilePath.toNative(), "*.svg");
```
SchnWalter: I always liked the UX when exporting images in GIMP (`File » Export As...`), where you can change the file name to have a different extension (or choose a file type from the list) and then, depending on the selected file type, you where presented with a 2nd dialog with configuration for your format (for changing the DPI and what not), if the format supported any configuration.

P.S. Thanks for all the great work on this.

ubruhin: Yeah I agree regarding the "Export As...", actually I considered to do it that way too. But since SVG is currently the only option, I decided to create an explicit menu item instead, but we can change it to a more generic "Export As..." menu item once we support more export formats.

If there is a strong opinion, I could change it now - otherwise I would change it when it's needed :)

ubruhin:
> The yellow rectangles for a symbol appear above the text.

> In the board export, there are a lot of empty groups:

These things are handled by Qt (I just render the `QGraphicsScene` into an SVG), thus I don't know how to fix that easily. I think it's not worth the effort to investigate these issues...

> The "grab crosses" are exported in the SVG:

Yeah that's a limitation of the current concept. The SVG export just uses the layer stack of the schematic editor, and since the crosses are visible there, they are also visible in the SVG. I think we should completely separate the whole `QGraphicsScene/Item` stuff from the data structure (see also #67), then it would be much easier to use different rendering settings for the different exports. But that's a heavy task, it will take some time to do that refactoring. Afterwards we can make the SVG export (and many other features) much more powerful.

ubruhin:
I added the `qt5-svg` package to the Docker containers used on CI, now CI succeeds and this PR is ready so far.

Probably I'll still include this change in the 0.1.4 release since installing `qt5-svg` shouldn't be an issue for packagers...

Issue #659, opened 2020-01-26
More standard licenses


cjdg:
it should be possible to add more licenses on new projects

dbrgn:
Which licenses would you suggest?

My suggestions would be:

- TAPR OHL
- CERN OHL 1.2

CERN OHL 2.0 is also [very promising](https://www.youtube.com/watch?v=Lg0BcLdfV-I), but still in beta.

(Note that you can *always* add other licenses as you wish, the wizard simply suggests some common ones.)

cjdg:
Can be both? or if there are more would be better, maybe making a repo for libraries, include some, and allows the sofotware to add manually from server or localy?

dbrgn:
I don't understand your question. This is only about the licenses that LibrePCB suggests in the license picker.

You can always add your own licenses to your projects or libraries by adding the necessary license files.

cjdg:
Yes but its ideal for final users to have at hand the licenses, remember, not all the users know all the licenses, and would be nice so they can have a little bit more options

dbrgn:
Sure, I'm all for adding more options :slightly_smiling_face: However, before licensing your project under a license, you have to understand that license. This is a step that LibrePCB cannot solve for you. We can just help you to apply a license once you've picked one.

I'll add a PR with TAPR OHL and CERN OHL.

PR #666, opened 2020-02-03
Create license wrapper class

dbrgn:
Not all licenses can be applied by copying a single license file. Some licenses also require that you include additional files, for example a license application guide or a file with contact information. See for example the CERN OHL: https://ohwr.org/project/cernohl/uploads/8eaab7317d1719b8dda14bbe796c4ebc/cern_ohl_v_1_2_howto.pdf

Right now `LicenseBase` is only implemented in `SingleFileLicense`, but more complex license types can be added later (e.g. for the CERN OHL which requires multiple files).

@ubruhin this is a first draft. What do you think about the design? Do you think the `LicenseBase` type hierarchy is overengineered? As an alternative design, we could also implement all the logic in a single function that returns license information in a map, or something like that. It's a bit less structured though (and we could also extend `LicenseBase` to ask the user additional license-related questions in the future).

Also, this is my first attempt at polymorphism in C++ and it doesn't compile :slightly_smiling_face: 

```
libs/librepcb/projecteditor/../../../../LibrePCB/libs/librepcb/projecteditor/licenses/singlefilelicense.cpp:43: undefined reference to `librepcb::project::editor::LicenseBase::~LicenseBase()'
/usr/bin/ld: ../../output/liblibrepcbprojecteditor.a(singlefilelicense.o): in function `librepcb::project::editor::LicenseBase::LicenseBase()':
/mnt/data/Projects/librepcb/build-librepcb-Desktop-Debug/libs/librepcb/projecteditor/../../../../LibrePCB/libs/librepcb/projecteditor/licenses/license_base.h:42: undefined reference to `vtable for librepcb::project::editor::LicenseBase'
/usr/bin/ld: ../../output/liblibrepcbprojecteditor.a(singlefilelicense.o):(.data.rel.ro._ZTIN8librepcb7project6editor17SingleFileLicenseE[_ZTIN8librepcb7project6editor17SingleFileLicenseE]+0x10): undefined reference to `typeinfo for librepcb::project::editor::LicenseBase'
collect2: error: ld returned 1 exit status
```

Any ideas?

Addresses #659.


0xB767B:
You can solve the linker error by implementing the destructor in the base class.

dbrgn:
> You can solve the linker error by implementing the destructor in the base class.

Ah, thanks! Then I get a "weak vtable" warning (`-Wweak-vtables`) though, since the class has only virtual methods.

ubruhin:
> Ah, thanks! Then I get a "weak vtable" warning (-Wweak-vtables) though, since the class has only virtual methods.

Could you paste the whole compiler warning?

However, on CI the issue seems to be a missing `#include <memory>` in some file since `std::unique_ptr` was not found.

libs/librepcb/projecteditor/licenses/license_base.h
```diff
@@ -0,0 +1,66 @@
...
+ */
+class LicenseBase {
+public:
+  virtual ~LicenseBase() noexcept {}
+
+  /**
+   * @brief Return reference to a list of (source-filename, target-filename)
+   * pairs.
+   */
+  virtual const QList<std::pair<QString, QString>>& getFiles() = 0;
```
ubruhin: Should be marked as `const noexcept`:
```suggestion
  virtual const QList<std::pair<QString, QString>>& getFiles() const noexcept = 0;
```

Same for `getDescription()`, and the corresponding methods in the derived class.

dbrgn:
Got reminded of this when you made the change in  #1033. Can you give a comment on the general approach that this PR has? Should it be done differently?

If you generally agree, I can try to rebase this PR and maybe get it working.

ubruhin:
To be honest, for licenses which simply require to copy some files (without additional logic) it looks a bit overengineered to me :see_no_evil: I think it depends on  what features we really need in future whether this design makes sense or not.

I mean, for just copying files it should be enough (and easy to understand) to just create a structure like that:

```cpp
struct License {
  QString name;
  QMap<QString, QString> files;
};

QList<License> licenses = {
  {"None", {}},
  {"CC0-1.0", {{"cc0-1.0.txt", "LICENSE.txt"}}},
  ...
};
```

This list could then also be used to automatically populate the combobox instead of adding each item manually.

So my main question is: Do we need more than that? If yes, what do we need exactly to cover the OHL licenses (or other licenses of our interest)? Once we know what we need, we can talk about the implementation :slightly_smiling_face:

dbrgn:
Yeah, that would probably work for the CERN OHL: https://ohwr.org/project/cernohl/uploads/8eaab7317d1719b8dda14bbe796c4ebc/cern_ohl_v_1_2_howto.pdf

The base class based approach would allow for more customization (e.g. the user could be asked for certain information, e.g. "Contact point wishing to receive information about manufactured Products" as per the CERN OHL).  However, that could be implemented with a more functional / declarative style as well. For now, we can just add a note to the generated files for now.

dbrgn:
I'll close this. Maybe I find time to open a new proposal.

PR #668, opened 2020-02-12
Fix grammar: was crashed -> crashed

dbrgn:
Crashing is an active verb, not a passive one.

### DIFF
--- a/libs/librepcb/common/fileio/transactionalfilesystem.cpp
+++ b/libs/librepcb/common/fileio/transactionalfilesystem.cpp
@@ -68,8 +68,8 @@ TransactionalFileSystem::TransactionalFileSystem(const FilePath& filepath,
     if (restoreMode == RestoreMode::ASK) {
       QMessageBox::StandardButton btn = QMessageBox::question(
           0, tr("Restore autosave backup?"),
-          tr("It seems that the application was crashed the last time. "
-             "Do you want to restore the last autosave backup?"),
+          tr("It seems that the application crashed the last time you opened "
+             "this project. Do you want to restore the last autosave backup?"),
           QMessageBox::Yes | QMessageBox::No | QMessageBox::Cancel,
           QMessageBox::Cancel);
       switch (btn) {
--- a/dev/doxygen/pages/project.md
+++ b/dev/doxygen/pages/project.md
@@ -192,7 +192,7 @@ To avoid such problems, the constructor of librepcb::project::Project tries to l
 directory with a librepcb::DirectoryLock object (librepcb::project::Project::mLock). If the project
 is already locked by another application instance, the project cannot be opened again (the
 constructor of librepcb::project::Project throws an exception). If the lock file exists because the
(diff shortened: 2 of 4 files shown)


libs/librepcb/common/fileio/transactionalfilesystem.cpp
```diff
@@ -68,7 +68,7 @@ TransactionalFileSystem::TransactionalFileSystem(const FilePath& filepath,
     if (restoreMode == RestoreMode::ASK) {
       QMessageBox::StandardButton btn = QMessageBox::question(
           0, tr("Restore autosave backup?"),
-          tr("It seems that the application was crashed the last time. "
+          tr("It seems that the application crashed the last time you opened this project. "
```
ubruhin: Line looks a bit long, could you run clang-format? :wink:

dbrgn: Done.

ubruhin:
Hmm it seems now you reverted most of your changes :grin:

dbrgn:
Aaargh. Reformatting and rebasing is and remains a bad combination :roll_eyes: I think I still have the old commits in my reflog.

libs/librepcb/common/fileio/transactionalfilesystem.cpp
```diff
@@ -68,8 +68,8 @@ TransactionalFileSystem::TransactionalFileSystem(const FilePath& filepath,
     if (restoreMode == RestoreMode::ASK) {
       QMessageBox::StandardButton btn = QMessageBox::question(
           0, tr("Restore autosave backup?"),
-          tr("It seems that the application was crashed the last time. "
-             "Do you want to restore the last autosave backup?"),
+          tr("It seems that the application crashed the last time you opened "
+             "this project. Do you want to restore the last autosave backup?"),
```
ubruhin: Oh, I just realized that this change is actually not fully valid :( The class `TransactionalFileSystem` is not only used for projects, but also for libraries. So the message should be kept more generic. I'm so sorry for missing that point in my last review, that wasn't my intention :weary:

dbrgn: I wonder if we could parametrize the message somehow, so that it's more clear?

ubruhin: Hmm actually I don't like to open a `QMessageBox` in such a low level class anyway, thus I quickly tried to implement a different concept: Instead of passing an enum for choosing the restore mode (yes/no/ask/abort), a callback can be passed instead. The callback then returns a bool to choose whether the backup should be restored or not. Then the message box has to be implemented in a higher level class, and thus could be more specific.

My suggestion: We merge this PR as-is and afterwards I'll open a PR to apply the refactoring - do you agree @dbrgn?

dbrgn: Yep sounds great! Opening message boxes from low-level classes is a bad idea indeed :)

Issue #612, opened 2019-12-01
Board Editor: Segment vs Track


EdizonTN:
##### VERSION / OS / ENVIRONMENT
LibrePCB Version: 0.1.3
Git Revision:     56bc60d347ff67df0fe1d57807d03f0606de557f
Build Date:       2019-11-30T21:59:40
Qt Version:       5.11.3 (built against 5.11.3)
CPU Architecture: x86_64
Operating System: Windows 7 SP 1 (6.1)
Platform Plugin:  windows

##### SUMMARY
In PCB Editor is swapped thinking about track and segment of track.

Wrong (actual situation): 
Trace = part of segment.
Segment = finished connection between pads.

Correctly (expected):
Trace = finished connection between pads.
Segment = single part of trace.

##### STEPS TO REPRODUCE
Open PCB editor of routed PCB, and right click to part of trace.
Click "remove trace" - LibrePCB clear a part of the trace (segment).
Or
Click "remove whole segment" - LibrePCB clear whole trace not segment.

##### EXPECTED RESULTS
Trace: pad to pad connected copper (one or more segments = complette trace)
Segment: simple part of trace.
SEGMENT:
![SEGMENT](https://user-images.githubusercontent.com/48570749/69919768-c7e0f680-1480-11ea-9ae7-33806a67fd14.png)

TRACE:
![TRACE](https://user-images.githubusercontent.com/48570749/69919769-c8798d00-1480-11ea-9ac8-2090567febf6.png)

##### ACTUAL RESULTS
![TrackvsSegment](https://user-images.githubusercontent.com/48570749/69919698-2063c400-1480-11ea-937d-870ca2b29bb1.gif)

I warning also in [TRANSIFEX](https://www.transifex.com/librepcb/librepcb-application/translate/#sk/$/151805783?q=issue%3Aopen)

ubruhin:
Hi @EdizonTN,

Sorry for not responding yet on Transifex. It's still on my ToDo list to review your comments and fix the source strings where needed ;)

Regarding the naming: I agree it might be a bit confusing. The thing is, we call it "**net** segment", not "**trace** segment" or something like that. So it is a part of the net, not a part of a trace.

Basically there are three things to differentiate, for example when selecting elements:
- The whole net, i.e. all elements of the same net (currently called just "net")
- Several connected lines & vias, building a part of a net (currently called "net segment")
- Only a single line (currently called "trace")

[...]

PR #674, opened 2020-02-16
Fix various translation issues

ubruhin:
Fixing some of the translation issues reported at [Transifex](https://www.transifex.com/librepcb/):
- Fix inconsistent "rescan libraries" menu items
  - Rename "Rescan Library" to "Rescan Libraries"
  - Rename "Update Library Database" to "Rescan Libraries"
  - Use consistent icon and keyboard shortcut
- Stop translating low-level error messages (see [discussion](https://librepcb.discourse.group/t/translation-for-low-level-error-messages/119)).

Open question: Should we stop translating keyboard shortcuts (like "Ctrl+S")? There are a lot of keyboard shortcuts in the translation files, but I'm not sure if that makes sense. Unfortunately I didn't find a good documentation how translation of keyboard shortcuts should be handled...

### DIFF
--- a/apps/librepcb/controlpanel/controlpanel.cpp
+++ b/apps/librepcb/controlpanel/controlpanel.cpp
@@ -748,7 +748,7 @@ void ControlPanel::on_favoriteProjectsListView_customContextMenuRequested(
   }
 }
 
-void ControlPanel::on_actionRescanLibrary_triggered() {
+void ControlPanel::on_actionRescanLibraries_triggered() {
   mWorkspace.getLibraryDb().startLibraryRescan();
 }
 
--- a/apps/librepcb/controlpanel/controlpanel.h
+++ b/apps/librepcb/controlpanel/controlpanel.h
@@ -117,7 +117,7 @@ private slots:
   void on_recentProjectsListView_customContextMenuRequested(const QPoint& pos);
   void on_favoriteProjectsListView_customContextMenuRequested(
       const QPoint& pos);
-  void on_actionRescanLibrary_triggered();
+  void on_actionRescanLibraries_triggered();
 
 private:
   // make some methods inaccessible...
--- a/apps/librepcb/controlpanel/controlpanel.ui
+++ b/apps/librepcb/controlpanel/controlpanel.ui
@@ -477,7 +477,7 @@ QListView::item
     <string>Quit</string>
    </property>
    <property name="shortcut">
-    <string>Ctrl+Q</string>
+    <string notr="true">Ctrl+Q</string>
    </property>
    <property name="menuRole">
     <enum>QAction::QuitRole</enum>
@@ -486,6 +486,9 @@ QListView::item
   <action name="actionAbout_Qt">
    <property name="text">
     <string>About Qt</string>
+   </property>
+   <property name="shortcut">
+    <string notr="true"/>
    </property>
    <property name="menuRole">
     <enum>QAction::AboutQtRole</enum>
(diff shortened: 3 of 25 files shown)


dbrgn:
> Should we stop translating keyboard shortcuts (like "Ctrl+S")?

On German keyboards the key is often labelled "Strg" (Steuerung). I'm not sure if there are other languages that also localize that key.

Also, other key like "Space" or "Enter" are also frequently translated.

EdizonTN:
> "Should we stop translating keyboard shortcuts (like "Ctrl+S")?"

I think no. As dbrgn wrote, don't forget for German, and also for non latin languages (Russian, Japanese,...)

ubruhin:
I agree that e.g. "Ctrl" should be translated according the selected language. But currently the "problem" is that the *whole* shortcut (e.g. "Ctrl+S") is going through a translator, and the translated string is then passed to `QKeySequence` - so for example if the translator for some reason translates "Ctrl+S" to "Strg+X", the shortcut is changed to Ctrl+X (at least that's what I expect, but I didn't verify it yet). And if the translation is not understood by `QKeySequence`, the shortcut is probably even broken(?). Therefore IMHO it doesn't make sense to let translators changing the keyboard shortcuts.

But maybe I'm wrong and the translation is only used for displaying the shortcuts in the menu, but not for its functionality - I'll check this. But even in that case there is a risk of inconsistency, since the displayed shortcut might be different from the actually implemented shortcut...

EdizonTN:
Idea:
Exists any easy way how to divide this sequences?
I think, each sequence divide to two parts:
- modifier/s: "CTRL+", "SHIFT+", "ALT+SHIFT+"
- HotKey: "S", "F5", "DEL", "PageUp" - based on content - 1 char length as non-translable (ASCII character), or longer as translable (Fn keys, Cursors keys, ...)

For example: 
"CTLR+S" -> Translated("CTRL+")+Non-translated("S")
"SHIFT+ALT+HOME" -> Translated("SHIFT+ALT+")+Translated("HOME")
"ALT+F12" -> Translated("ALT+")+Translated("F12")
"ALT+2" -> Translated("ALT+")+Non-Translated("2")

ubruhin:
OK now I made some tests and the behavior is (more or less) how I expected it:

- Translators can change keyboard shortcuts (e.g. translating "Ctrl+S" to "F7" will change the shortcut).
- If the translated shortcut not understood by Qt, the shortcut is broken (disappears from the menu) -> this means Qt has to understand *translated* shortcuts in any language, which sounds very weird to me.

But the good thing is: The shortcut displayed in the menus is *not* the string translated by LibrePCB translators, but a string translated by Qt translators. So on a German system "Ctrl+S" appears as "Strg+S" even though LibrePCB translators did not translate the shortcut. Unfortunately currently this doesn't work with our binary releases for Linux (and maybe macOS?) because of https://github.com/probonopd/linuxdeployqt/issues/115, but I'm trying to work around this issue.

So I think we can just disable translation for keyboard shortcuts and everything is fine since the shortcuts will still be displayed in the chosen language -> commit added to disable all shortcut translations.

In addition, I also swapped the terms "trace" and "segment" in the board editor to fix #612.

ubruhin:
> Unfortunately currently this doesn't work with our binary releases for Linux (and maybe macOS?) because of probonopd/linuxdeployqt#115, but I'm trying to work around this issue.

It turned out the problem is not linuxdeployqt, but the fact that Qt translations are simply not installed in our Docker containers used on CI :sob: I'll fix that in a separate PR, so I think this PR is now ready.

Issue #362, opened 2018-10-27
Cannot change from millimeters to inches


ouabache:
##### VERSION / OS / ENVIRONMENT

```
LibrePCB Version: 0.1.0-rc1
Git Revision:     d49c0a70e23f5690e4d5785684ef828e04c2cb26
Build Date:       2018-10-21T12:05:18
Qt Version:       5.9.1 (built against 5.9.1)
CPU Architecture: x86_64
Operating System: Ubuntu 18.04.1 LTS
Platform Plugin:  xcb
```

##### SUMMARY
changing librePCB to inches has no effect on library or schematic editors

##### STEPS TO REPRODUCE
change workspace to inches

##### EXPECTED RESULTS
editors should configure to inches

##### ACTUAL RESULTS
editors stay in millimeters

ubruhin:
This is simply not implemented yet ;)

I think we should use this workspace setting only when creating new projects. Each schematic page and board stores the grid interval and unit in their files separately. This allows for example to use grid 0.1 Inches for schematics, but 1.0mm for boards.

But when creating a new project, its schematic and board should be initialized with the unit defined in the workspace settings.

In addition, the library editor should always use the unit defined in workspace settings (in contrast to schematics/board, grid interval is not stored in library elements).

ouabache:
So will you have libraries in either mm or inch and convert when used in the other domain?

Then you should have every symbol file declare its snap grid spacing in whatever units that it uses. Add a new field (snapgrid  2.54) to every symbol and you you will know that it is in metric. Once you add inches then you know to scale all the numbers before use. If you see (snapgrid 0.1) then you know it is in inches and you should scale if you are using metric.

This will also also avoid the problem that kicad often sees when you build a library using a snapgrid smaller than the schematic editor uses. Your pins can't connect.  This way the pins will always scale to whatever snapgrid the editor is using. The graphics may change size but the connectivity is intact.

You really do want to add this before your first official release. It may never get used but that's better than trying to deal with legacy code if you have to add it later.

John Eaton

ubruhin:
I don't think symbols and footprints need to know what grid interval was used to draw them. The snap issue wouldn't be solved with this, it's an issue related only to the editors. Board and schematic editors must work properly, independent of the grid interval used in symbols and footprints.

In files, LibrePCB always uses millimeters for everything. Switching to inches is only a matter of editors, they just convert all values between millimeters and inches.

ouabache:
[...]

PR #684, opened 2020-04-12
Respect configured workspace length unit in editors

ubruhin:
## Problem
Until now, all number edit widgets in all editors always used Millimeters, no matter which unit was configured in workspace settings. Also it was not possible to enter values in other units by explicitly entering the unit, e.g. by writing "100mils". See also #362.

## Changes
This PR heavily refactors these number input widgets in several ways:
- Code refactoring: Add new base class `LengthEditBase`, based directly on `QAbstractSpinBox`
  instead of `QDoubleSpinBox` -> this way the actual value hold by the widget is `Length` instead of `double`.
- By default, show the values in the unit configured in workspace settings.
- Add context menu to switch unit:
grafik
- Add support for entering mathematical expressions (using [muparser](https://beltoforion.de/article.php?a=muparser)), optionally including the unit:
Peek 2020-04-12 22-57
- Use automatic spinbox step value, determined from current value (for example if a multiple of 0.1mm is entered, the step is 0.1mm, but if a multiple of 10mils is entered, the step is 10mils):
Peek 2020-04-12 23-05
- Reduce number of displayed decimal places, depending on unit (converting values between different units can lead to very odd numbers with many decimal places, so for readability only a few decimal places are displayed).

# Open Questions

See https://github.com/LibrePCB/LibrePCB/issues/362#issuecomment-612683842.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -37,3 +37,6 @@
 [submodule "i18n"]
     path = i18n
     url = https://github.com/LibrePCB/librepcb-i18n.git
+[submodule "libs/muparser"]
+    path = libs/muparser
+    url = https://github.com/LibrePCB/muparser.git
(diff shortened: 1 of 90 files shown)


dbrgn:
Very cool! :tada:

Issue #588, opened 2019-11-18
Board editor: Sometimes air wires of same net disappear


doom-fr:
##### VERSION / OS / ENVIRONMENT

```
LibrePCB Version: 0.1.3-unstable
Git Revision:     bd616e14d1366486cf53cad8b8c7e3e67119d960
Build Date:       2019-11-17T15:33:39
Qt Version:       5.12.3 (built against 5.12.3)
CPU Architecture: x86_64
Operating System: Ubuntu 18.04.3 LTS
Platform Plugin:  xcb
```

##### SUMMARY
Sometimes when moving components, Air Wires (the yellow lines between equipotential pads) of same net disappear.

##### STEPS TO REPRODUCE
Seams to be when two consecutive air wires of same net are aligned or quite aligned.
Not always reproductible.

![image](https://user-images.githubusercontent.com/7809481/69100246-0b893880-0a5d-11ea-9022-ebf1ffd102fa.png)

Air wires are OK. And move a little bit and ...

![image](https://user-images.githubusercontent.com/7809481/69100195-ebf21000-0a5c-11ea-8229-8673d82ac88a.png)

##### EXPECTED RESULTS
Air wires are drawn

##### ACTUAL RESULTS
Sometimes, air wires are not drawn.

doom-fr:
Can you qualify this as a bug ?
It happens a lot.

doom-fr:
@ubruhin Can you have a look at this ?
It is really annoying, because when you align devices, the air wire disappears.

ubruhin:
Yeah I think I can try to fix this.

doom-fr:
Great !

doom-fr:
![Airwire_loss](https://user-images.githubusercontent.com/7809481/69906151-2b0a5480-13bf-11ea-8042-22aeafee5e97.gif)

doom-fr:
![Airwire_loss](https://user-images.githubusercontent.com/7809481/69906165-6e64c300-13bf-11ea-918b-3f7679e1cd11.gif)

doom-fr:
Just see that it also has a impact on DRC.
When the airwire is not drawn, the missing connexion error is not raised.

This is really a problem as it is so possible to forget to link some components.

Can you please fix it soon ??

ubruhin:
Yes yes, I added it to the 0.1.4 milestone ;)

doom-fr:
Sorry to be so impatient :-)

[...]

PR #687, opened 2020-04-19
Fix possibly disappearing airwires

ubruhin:
It's actually not a fix, but a workaround for issue #588. The Delaunay triangulation library is not able to handle colinear points in the input data (our net points) and silently ignores some of the input points, so there are edges (our airwires) missing in the output.

I spend quite some time on this issue, and I think it's not possible to actually fix this problem on our side, so the fix needs to be done in [delaunay-triangulation](https://github.com/Bl4ckb0ne/delaunay-triangulation). But actually I'm not sure if it's worth investigating and fixing the issue in that library - in my opinion the overall quality of that library is not acceptable for an EDA tool anyway. For long-term, we might switch to a different library, for example [delaunator-cpp](https://github.com/abellgithub/delaunator-cpp) which seems to be developed more seriously. But they currently also have a similar issue (https://github.com/abellgithub/delaunator-cpp/issues/8), so we should wait until this issue is resolved.

However, to get LibrePCB 0.1.4 released as soon as possible with disappearing airwires fixed, I simply implemented some kind of "fallback". Even if delaunay-triangulation failed to add some edges, we now still get all edges connected, but connections might not be "optimal", i.e. not the shortest distance is shown. But at least no airwires disappear now, so the risk of producing PCBs with missing connections is gone.

I also refactored the airwires builder to make it easier testable, and added some unittests which are able to detect the current bug.

### DIFF
--- a/libs/librepcb/common/algorithm/airwiresbuilder.cpp
+++ b/libs/librepcb/common/algorithm/airwiresbuilder.cpp
@@ -60,12 +60,25 @@ AirWiresBuilder::AirWires AirWiresBuilder::buildAirWires() noexcept {
   uint connectedEdges = mEdges.size();
 
   // determine additional edges between found points (candidates for airwires)
-  if (mPoints.size() >= 3) {  // minimum 3 points needed for triangulation
+  if (mPoints.size() == 2) {
+    mEdges.emplace_back(mPoints[0], mPoints[1], -1);
+  } else if (mPoints.size() == 3) {
+    // manually triangulate since it is easy and more stable than the
+    // delaunay-triangulation library
+    mEdges.emplace_back(mPoints[0], mPoints[1], -1);
+    mEdges.emplace_back(mPoints[1], mPoints[2], -1);
+    mEdges.emplace_back(mPoints[2], mPoints[0], -1);
+  } else if (mPoints.size() >= 3) {
+    // since delaunay-triangulation sometimes doesn't work well, add fallback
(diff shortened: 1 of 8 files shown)


libs/librepcb/common/algorithm/airwiresbuilder.cpp
```diff
@@ -0,0 +1,198 @@
...
+              return a.weight > b.weight;
+            });
+
+  while (mstSize < mstExpectedSize && !mEdges.empty()) {
+    auto& dt = mEdges.back();
+
+    int srcTag = tags[dt.p1.id];
+    int trgTag = tags[dt.p2.id];
+    // printf("mstSize %d %d %f %d<>%d\n", mstSize, mstExpectedSize,
+    // dt.weight, srcTag, trgTag);
```
dbrgn, line 135: Driveby comment (I didn't really look at the logic): There's quite some commented out debug code in here (more below). Do you want to keep that?

ubruhin: Yeah I don't like such comments too, but as stated above the function, this function is copied from Horizon (which copied it from KiCad) and I'd just like to keep it as-is to keep the diff to the upstream code as small as possible :wink:

dbrgn: Ah, good point, I missed that! :+1:

PR #690, opened 2020-04-26
Improve keyboard navigation in dialogs

ubruhin:
During testing of the 0.1.4 release candidate, I realized that the keyboard navigation of most dialogs with input fields was not very user friendly. This PR improves following things:

- Adjust order of widgets to always have the most important input field at the top of the dialog.
- Pre-select the value of the most important input field so one can immediately start typing a new value after opening the dialog.
- In multiline text edit widgets, make tabulator key navigating to the next widget instead of inserting a tab.
- Set reasonable tab order (usually from top to bottom).

Might make @dbrgn more happy when using LibrePCB :grin:


dbrgn:
Cool, I meant to address this for a long time 🙂 Especially the field tab order was quite random sometimes.

Issue #373, opened 2018-11-12
Workspace Settings: Custom PDF viewer


Symbian9:
##### SUMMARY

After "PDF Export" LibrePCB try immediately open PDF viewer (default).

##### SOLUTION

Add tab "External applications" to "Workspace Settings" where user could set which PDF viewer/app should be used for open exported PDF.

Here is how such functions shown in *KiCad* app:

![pic.1](http://i.imgur.com/vZl3UtA.png)

Also add option "Do not open exported PDF".

P.S.: Tab "External applications" in "Workspace Settings" could be later extended for choosing other apps such as *image editor* for opening exported images, *PDF editor*, etc.

ubruhin:
Good idea, makes sense :)

I just think that it would be nice to allow choosing "open exported PDF" every time when exporting (maybe sometimes you want to open it, and sometimes not). But I'm not sure how this could be implemented in the user interface... Any ideas?

Beside that user interface thing, I think it shouldn't be too hard to implement this feature, so I'll add the "easy" tag here ;)

ubruhin:
Sounds like a good idea! :+1: :smiley: 

For the workspace settings I would suggest following:
- Default value for PDF viewer is an empty string, which means to use the system's default PDF viewer.
- The input field accepts the placeholder "{{PDF}}" which LibrePCB then replaces with the path to the PDF (we could also use a shorter placeholder like "%p" like other applications, but I think for consistency it makes sense to use the same placeholder syntax like everywhere else in LibrePCB).

ubruhin:
Looks good! :smiley:

dbrgn:
As an improvement idea, we could even detect common PDF viewers on the system and list those as options (with "custom" as a fallback of course).

Some PDF viewers I know and use:

- evince
- mupdf
- zathura

ubruhin:
> As an improvement idea, we could even detect common PDF viewers on the system and list those as options (with "custom" as a fallback of course).

I wouldn't do that. It's just more complexity, and as @Symbian9 already wrote, a custom viewer is especially useful for portable viewers which can't be detected anyway. And users which want to use an installed viewer are probably fine with the system's default viewer (otherwise one should consider changing the default viewer system-wide :wink:).

@Symbian9 I would very welcome a pull request to add the workspace settings tab you posted above :)

Symbian9:
> I would very welcome a pull request

Well, ... I just *paint* draft above using image editor.

So, there is no code snippet for that :-(

[...]

PR #694, opened 2020-05-02
Support configuring PDF viewer in workspace settings

NaoPross:
Solution for #373

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
@@ -547,7 +547,16 @@ void BoardEditor::on_actionExportAsPdf_triggered() {
       board->print(printer);  // can throw
     }
 
-    QDesktopServices::openUrl(QUrl::fromLocalFile(filepath.toStr()));
+    // open pdf with system / custom reader accordingly
+    QString pdfReaderCommand =
+        mProjectEditor.getWorkspace().getSettings().pdfReaderCommand.get();
+
+    if (mProjectEditor.getWorkspace().getSettings().useCustomPdfReader.get()) {
+      QProcess::execute(pdfReaderCommand.replace("{{FILEPATH}}", filepath.toStr()));
+    } else {
+      QDesktopServices::openUrl(QUrl::fromLocalFile(filepath.toStr()));
+    }
+
   } catch (Exception& e) {
     QMessageBox::warning(this, tr("Error"), e.getMsg());
   }
--- a/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
@@ -548,15 +548,48 @@ void BoardEditor::on_actionExportAsPdf_triggered() {
     }
 
 
-    // open pdf with system / custom reader accordingly
-    if (mProjectEditor.getWorkspace().getSettings().useCustomPdfReader.get()) {
-      QString pdfCmd=
-          mProjectEditor.getWorkspace().getSettings().pdfReaderCommand.get();
-      QProcess::startDetached(pdfCmd.replace("{{FILEPATH}}", filepath.toStr()));
-    } else {
-      QDesktopServices::openUrl(QUrl::fromLocalFile(filepath.toNative()));
-    }
+    // Open PDF
+    {
+      namespace ws = ::librepcb::workspace;
+      const ws::WorkspaceSettings& workspaceSettings =
+          mProjectEditor.getWorkspace().getSettings();
 
+      ws::WorkspaceSettings::PdfOpenBehavior bhv
+          = workspaceSettings.pdfOpenBehavior.get();
+
+      if (bhv == ws::WorkspaceSettings::PdfOpenBehavior::NEVER) {
+        // do nothing
+      } else {
+        // TODO: refractor into FileUtils::askUserAndOpen or something similar
+        //       to avoid duplicated code in SchematicEditor
+
+        bool doOpenPdf = true;
+        if (bhv == ws::WorkspaceSettings::PdfOpenBehavior::ASK) {
+          int openPdf = QMessageBox::information(
+                          this, tr("PDF Export"),
+                          tr("PDF exported successfully"),
+                          QMessageBox::Ok | QMessageBox::Open);
+
+          if (openPdf == QMessageBox::Ok)
+            doOpenPdf = false;
+        } else if (bhv != ws::WorkspaceSettings::PdfOpenBehavior::ALWAYS) {
+          // this is the last possible state, otherwise there is an error
+          throw LogicError(__FILE__, __LINE__);
+        }
+
+        if (doOpenPdf) {
+          if (workspaceSettings.useCustomPdfReader.get()) {
+            QString pdfCmd = workspaceSettings.pdfReaderCommand.get();
+            // TODO: make it smarter to detect or insert quotes for path
+            //        containing spaces
+            QProcess::startDetached(pdfCmd.replace(
+                                      "{{FILEPATH}}", filepath.toNative()));
+          } else {
+            QDesktopServices::openUrl(QUrl::fromLocalFile(filepath.toNative()));
+          }
+        }
+      }
+    }
   } catch (Exception& e) {
     QMessageBox::warning(this, tr("Error"), e.getMsg());
   }
--- a/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
@@ -547,14 +547,14 @@ void BoardEditor::on_actionExportAsPdf_triggered() {
       board->print(printer);  // can throw
     }
 
-    // open pdf with system / custom reader accordingly
-    QString pdfReaderCommand =
-        mProjectEditor.getWorkspace().getSettings().pdfReaderCommand.get();
 
+    // open pdf with system / custom reader accordingly
     if (mProjectEditor.getWorkspace().getSettings().useCustomPdfReader.get()) {
-      QProcess::execute(pdfReaderCommand.replace("{{FILEPATH}}", filepath.toStr()));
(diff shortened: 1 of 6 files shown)


libs/librepcb/workspace/settings/workspacesettingsdialog.ui
```diff
@@ -280,6 +280,143 @@
...
+              </property>
+              <attribute name="buttonGroup">
+               <string notr="true">pdfOpenGroup</string>
+              </attribute>
+             </widget>
+            </item>
+           </layout>
+          </item>
+          <item row="1" column="2">
+           <widget class="QLineEdit" name="pdfCustomCmdEdit">
```
ubruhin, line 366: For this line edit, it would be good to set following properties:
- `maxLength` to something reasonable, e.g. 200 or so (there were already strange behavior of the GUI reported when pasting extremely long texts into such fields, this can be prevented by limiting its length).
- `placeholderText` to show a usage example like `e.g. "evince {{FILEPATH}}"` to make it more intuitive.
- `clearButtonEnabled` to `true` (sometimes useful, and shouldn't hurt :wink:)

libs/librepcb/workspace/settings/workspacesettingsdialog.ui
```diff
@@ -280,6 +280,143 @@
...
+      <attribute name="title">
+       <string>External Applications</string>
+      </attribute>
+      <layout class="QVBoxLayout" name="verticalLayout_2">
+       <item>
+        <widget class="QGroupBox" name="pdfBox">
+         <property name="title">
+          <string>PDF Reader</string>
+         </property>
+         <layout class="QGridLayout" name="gridLayout">
```
ubruhin, line 293: For this use-case, a `QFormLayout` is actually more suitable than `QGridLayout`. And the horizontal spacer should be removed :wink: As an example, you could take a look at the other tabs of that dialog.

libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
```diff
@@ -547,7 +547,16 @@ void BoardEditor::on_actionExportAsPdf_triggered() {
       board->print(printer);  // can throw
     }
 
-    QDesktopServices::openUrl(QUrl::fromLocalFile(filepath.toStr()));
+    // open pdf with system / custom reader accordingly
+    QString pdfReaderCommand =
+        mProjectEditor.getWorkspace().getSettings().pdfReaderCommand.get();
```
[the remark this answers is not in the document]
NaoPross: Done!

libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
```diff
@@ -547,7 +547,16 @@ void BoardEditor::on_actionExportAsPdf_triggered() {
...
+
+    if (mProjectEditor.getWorkspace().getSettings().useCustomPdfReader.get()) {
+      QProcess::execute(pdfReaderCommand.replace("{{FILEPATH}}", filepath.toStr()));
```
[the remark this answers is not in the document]
NaoPross: Done

libs/librepcb/workspace/settings/workspacesettingsdialog.ui
NaoPross: Done. Though I was wondering if I should put a file picker button on the right of the TextEdit, because on Windows "to run a custom command" is pretty much synonymous for pick an exe somewhere.

libs/librepcb/workspace/settings/workspacesettingsdialog.ui
NaoPross: Done

libs/librepcb/workspace/settings/workspacesettingsdialog.ui
ubruhin: Yeah actually I was also thinking about that :wink: Probably such s button to pick an executable would indeed be useful. Also because you could then auto-fill the TextEdit with the choosen path and just append "{{FILEPATH}}", so for typical applications the user don't have to specify the arguments manually...

If you'd like to add such a button - perfect! If not - no problem, I'd still merge this PR :)

ubruhin:
Well done with the open behavior enum! :+1:

I saw following things which should be changed to get this PR merged:
- CI fails because the `[[fallthrough]]` statements raises a compiler warning
- The term "behaviour" seems to be British English. Should be changed to American English, i.e. "behavior" (at least in the settings file, since files are intended to be in American English).
- I realized that paths with spaces in it need special attention. Either LibrePCB needs to add quotes around the substituted filepath, or the user needs to write `"{{FILEPATH}}"` instead of `{{FILEPATH}}`. Actually I'd prefer the latter since it provides more flexibility - maybe there are PDF viewers which require different quotes or so (e.g. `viewer.exe /cmd="open '{{FILEPATH}}'"`). But then we should at least add the quotes to the placeholder and tooltip text to make sure the user doesn't forget to add them...

Things which would be nice to have, but not needed to get this merged:
- As discussed above, a button to choose an executable would be nice to have.
- I'd rename `pdf_reader_cmd` to `custom_pdf_reader_command` to make the settings file more expressive (for this reason, we usually don't use abbreviations in files).
- I think the code in BoardEditor and SchematicEditor could be simplified by replacing the rather complex switch-case by simpler `if` statements, and optionally by even factoring our the code into a new class like `FileOpener` since it is exactly identical for both editors.
- In the message box, the buttons "Open" and "Cancel" are not really answers for the contained question "Open exported PDF?". "Yes" and "No" might be more intuitive :wink: In addition, since it is a question, why not using `QMessageBox::question()` instead of `QMessageBox::information()`?
- In the TextEdit placeholder, I'd start the text with "e.g." to make clear it is only an example, not some kind of default value or so (sorry if this wasn't clear from my suggestion I made previously :wink:).

NaoPross:
> The term "behaviour" seems to be British English. Should be changed to American English, i.e. "behavior" (at least in the settings file, since files are intended to be in American English).

Yeah I thought about it as well, but too late. Easy fix.

> CI fails because the `[[fallthrough]]` statements raises a compiler warning

Yep, apparently `[[fallthrough]]` was not standardized until C++17, so gcc < 7.1 fails to build it. But I'm using clang 6 so I didn't notice. I'll replace it with `if`s.

> I realized that paths with spaces in it need special attention.

> But then we should at least add the quotes to the placeholder and tooltip text to make sure the user doesn't forget to add them...

> I think the code in BoardEditor and SchematicEditor could be simplified by replacing the rather complex switch-case by simpler if statements, and optionally by even factoring our the code into a new class like FileOpener since it is exactly identical for both editors.

Umh right, I'll update the tooltip for now, and then maybe in the future if we were to refractor it into an external class we could implement a smarter logic that detects spaces and quotes to place `'` or `"` (or nothing if the user already placed them) accordingly.

> In the message box, the buttons "Open" and "Cancel" are not really answers for the contained question "Open exported PDF?". "Yes" and "No" might be more intuitive wink In addition, since it is a question, why not using `QMessageBox::question()` instead of `QMessageBox::information()`?

It made more sense to me because I was thinking around the cause of the message box, that is the information "The PDF has been exported". And so even if one doesn't read the contents of the box, it would be very clear what the "open" button does (against a YN question). Maybe I should change the title to "PDF Export" and the content to "The PDF has been successfully exported" to be consistent with the UX.

> In the TextEdit placeholder, I'd start the text with "e.g." to make clear it is only an example, not some kind of default value or so (sorry if this wasn't clear from my suggestion I made previously wink).

Makes sense.

ubruhin:
> Yep, apparently [[fallthrough]] was not standardized until C++17, so gcc < 7.1 fails to build it. But I'm using clang 6 so I didn't notice. I'll replace it with ifs.

:+1:

> Agree, except that I'd go with pdf_custom_reader_command to be consistent with the other pdf options, having the pdf_ prefix.

Hmm but `use_custom_pdf_reader` does not have this prefix either... Anyway, for me it doesn't really matter if there is a common prefix or not - as long as the settings keys are expressive/self-explaining, it's fine for me :wink:

> Umh right, I'll refractor and update the tooltip for now, and then maybe in the future if we were to move it into an external class we could implement a smarter logic that detects spaces and quotes to place ' or " (or nothing if the user already placed them) accordingly.

Agree :+1:

> It made more sense to me because I was thinking around the cause of the message box, that is the information "The PDF has been exported". And so even if one doesn't read the contents of the box, it would be very clear what the "open" button does (against a YN question). Maybe I should change the title to "PDF Export" and the content to "The PDF has been successfully exported" to be consistent with the UX.

Sounds good, however I'd then use buttons OK+Open instead of Cancel+Open (cancel sounds like you want to abort the triggered action, but if the message box says "PDF successfully exported", the action is already done, i.e. cannot be aborted anymore).

libs/librepcb/workspace/settings/workspacesettingsdialog.ui
NaoPross: Done

NaoPross:
Done. If the CI passes I think it's (almost) ready to merge.

ubruhin:
> Done. If the CI passes I think it's (almost) ready to merge.

It passed :smiley: With "almost" you mean if *we* request to change something, or do you have something in mind *you* would like to change?

From my side, it looks good so I'd merge it. I'll do a squash-merge since the branch is outdated and contains several fixup commits. If you insist on not squash-merging, you could rebase & cleanup the commits :slightly_smiling_face:

NaoPross:
I've cleaned it up a bit, but it's still far from perfect. If you find them ok I'd like you to keep them around (don't know exactly why), if not, you can squash them no problem.

ubruhin:
> I've cleaned it up a bit, but it's still far from perfect. If you find them ok I'd like you to keep them around (don't know exactly why), if not, you can squash them no problem.

The branch was still not based on the current `master` (your `master` is outdated), and honestly I'm not fully happy with the history (sorry, I'm a bit picky :grin:) thus I squash-merged it :wink:

So, thanks a lot for your contribution! :rocket: It will be contained in the 0.2.0 release (because of the file format change).

PR #698, opened 2020-05-15
Allow unbundling of some vendored libraries

### LINKED ISSUE #385: Unbundle statically linked libraries
##### SUMMARY
I'm packaging LibrePCB for Guix, a GNU/Linux distribution. LibrePCB has several libraries bundled into its source under `libs/`. This makes packaging less efficient with dependencies duplicated across several projects. It also causes dependencies used in LibrePCB to lag behind their latest released version.

##### SOLUTION
Unbundle all dependencies, and link to them as dynamic libraries. Guix and most other GNU/Linux distributions prefer to deal with dependencies as separate packages. This lets the distro's package manager handle dependency management and avoids the problems I mentioned earlier.

Thank you for your consideration!

dbrgn:
After about 6 hours of reading and changing qmake files, a first version finally works! :tada: 

The general approach is as follows:

- By default, everything is built as a static binary using vendored libs
- You can choose to unbundle certain libraries by specifying `UNBUNDLE+=quazip` as a qmake parameter
- If any library is unbundled (`!isEmpty(UNBUNDLE)`), this will lead to dynamic builds instead of static builds
- Linking parameters are determined using `pkg-config`
- Some libs like quazip require different imports depending on the header source. In those cases, I conditionally define a `SYSTEM_QUAZIP` preprocessor variable which can then be used to import the correct headers

To build with unbundled quazip:

```
mkdir build
cd build
qmake -r ../librepcb.pro UNBUNDLE+=quazip CONFIG+=ccache
make -j8
```

This results in a lot of dynamic libraries:

The `librepcbcommon` library now links against system quazip. The binaries link against the generated libs.

The binaries seem to work!

There's still some work to do:

- [x] Ensure that the static (default) build still passes CI
- [x] Verify whether linking is correct (right now `librepcb-cli` links against every librepcb lib while `librepcb` only links against `liblibrepcbcommon.so`, which seems a bit fishy)
- [x] Show the linking type (static or dynamic) in the about dialog
- [x] Allow unbundling other libraries
- [x] Allow unbundling all libraries without specifying them all
- [x] Test with dynamic build as well

Fortunately CI and build targets like AppImage / Flatpak / etc don't need to be updated, those use the static builds anyways. We might want to add a dynamic build to CI though, to ensure that nothing breaks.

Feedback welcome, @arunisaac @ktrace @FFY00

This will fix #385.


README.md
```diff
@@ -151,9 +151,26 @@ on the command line:
...
+
+```bash
+qmake -r ../librepcb.pro CONFIG+=ccache
+```
+
+By default, all binaries will be linked statically using vendored libraries. If
+you would prefer to unbundle some libraries, set the `UNBUNDLE` variable:
+
+```bash
+qmake -r ../librepcb.pro UNBUNDLE+=quazip
```
FFY00, line 168: Would it be possible to add an option to unbundle everything?

Otherwise LGTM, I did not test however.

dbrgn: I added support for `UNBUNDLE=all`.

dbrgn:
About dialog:

static dynamic_000

dbrgn:
@ubruhin could you test whether building this branch with `UNBUNDLE+=quazip` works for you? It seems to work for me, but in CI there are linker errors when building...

FFY00:
I started drafting up a PKGBUILD, I am currently hitting this error:

```
g++ -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -o ../../output/uuid-generator main.o mainwindow.o qrc_images.o moc_mainwindow.o   -L../../output -llibrepcbcommon -lsexpresso -lclipper -lmuparser /usr/lib/libQt5Widgets.so /usr/lib/libQt5Gui.so /usr/lib/libQt5Core.so -lGL -lpthread
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestComponent(librepcb::Uuid const&) const'
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestPackageCategory(librepcb::Uuid const&) const'
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestDevice(librepcb::Uuid const&) const'
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestSymbol(librepcb::Uuid const&) const'
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestComponentCategory(librepcb::Uuid const&) const'
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestPackage(librepcb::Uuid const&) const'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:161: ../../output/librepcb-cli] Error 1
make[2]: Leaving directory '/build/librepcb/src/LibrePCB/build/apps/librepcb-cli'
```

dbrgn:
@FFY00 yeah, we're hitting that in CI as well. Will have to find the reason once I find time to do so (maybe tomorrow).

Regarding the PKGBUILD, if it's of any help, here's the current AUR package: https://aur.archlinux.org/packages/librepcb (This doesn't do unbundling though.(

dbrgn:
I found the reason for the build errors.

This is the error message when linking `librepcb-cli` dynamically:

```
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestComponent(librepcb::Uuid const&) const'
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestPackageCategory(librepcb::Uuid const&) const'
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestDevice(librepcb::Uuid const&) const'
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestSymbol(librepcb::Uuid const&) const'
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestComponentCategory(librepcb::Uuid const&) const'
/usr/bin/ld: ../../output/liblibrepcblibrary.so: undefined reference to `librepcb::workspace::WorkspaceLibraryDb::getLatestPackage(librepcb::Uuid const&) const'
```

The reason is that `libs/librepcb/library` has an implicit dependency on `libs/librepcb/workspace`. We could fix that by reordering the library ordering. However, if I read https://developers.librepcb.org/de/d34/doc_architecture_overview.html correctly, this dependency is actually a violation of the architecture/layering, right?

The dependency:

```
[unbundle]danilo@x2000:LibrePCB$ rg workspace/ libs/librepcb/library
libs/librepcb/library/libraryelementcache.cpp
28:#include <librepcb/workspace/library/workspacelibrarydb.h>

libs/librepcb/library/pro/Makefile
3275:		../../workspace/library/workspacelibrarydb.h
```

It was added in d44243c15. @ubruhin is this dependency intentional? Should it be fixed?

ubruhin:
> this dependency is actually a violation of the architecture/layering, right?

Good catch, that's indeed an architecture violation! :scream: I would say it should be fixed by moving `LibraryElementCache` to `libs/librepcb/libraryeditor`, but then we have to fix the dependencies from other classes in `libs/librepcb/library` to `LibraryElementCache`. So far I see two options:
- Either by adding an interface for `LibraryElementCache` to `libs/librepcb/library`, and let the other classes reference only the interface.
- Or by moving all classes accessing `LibraryElementCache` to `libs/librepcb/libraryeditor` too - those classes are all implementations of the [Qt Model/View Pattern](https://doc.qt.io/qt-5/model-view-programming.html) for various data structures. Somehow I think this would be the cleaner solution, but it requires much more effort and raises new questions like where should we put model-view classes from `libs/librepcb/common` then?

dbrgn:
I'm not familiar with that part of the code. It does sound like caching should be the editor's business and not part of the library abstraction, but I don't think it would hurt if the library would provide an interface.

Do you think you could work on these things before the next minor or major release, to unblock the unbundling? It would be awesome to get first-level package support from distros. If we get those packages sooner, then more time can be spent testing them.

ubruhin:
> Do you think you could work on these things before the next minor or major release, to unblock the unbundling?

Yes, I'm working on this right now.

dbrgn:
@FFY00 the build error should now be fixed! I also added expanded [build instructions](https://github.com/LibrePCB/LibrePCB/blob/unbundle/dev/doxygen/pages/building.md) that may be useful to you. I would recommend to explicitly specify that you want a release build (even if it should be the default config).

dbrgn:
Wohoo, tests finally pass! :rocket: 

Before we merge this, I want to add an unbundling option for fontobene-qt5 as well.

dbrgn:
@ubruhin I added unbundling support for fontobene-qt5 as well. This is now ready to review!

In order to find the fontobene-qt5 headers, the fontobene-qt5 system package must provide a pkg-config file. For Arch, I've added such a file to the AUR package: https://aur.archlinux.org/packages/fontobene-qt5/

ci/build_librepcb.sh
```diff
@@ -38,7 +38,18 @@ fi
 
 # build librepcb
 mkdir build && pushd build
-qmake ../librepcb.pro -r ${BUILDSPEC-} "QMAKE_CXX=$CXX" "QMAKE_CC=$CC" "QMAKE_CFLAGS=$CFLAGS" "QMAKE_CXXFLAGS=$CXXFLAGS" "PREFIX=`pwd`/install/opt"
+if [ "${UNBUNDLE-}" != "" ]; then
+  ADDITIONAL_ARGS="UNBUNDLE+=$UNBUNDLE"
+else
+  ADDITIONAL_ARGS=""
+fi
```
ubruhin, line 45: Could you maybe move these lines which determine `ADDITIONAL_ARGS` to above the translations download step, i.e. below the lines which determine the other qmake arguments (`CFLAGS`, `BUILDSPEC`, ...)? :)

common.pri
```diff
@@ -44,5 +44,28 @@ CONFIG += warn_on
...
+}
+
+# add linking information to defines
+isEmpty(UNBUNDLE) {
+    DEFINES += "LINKING_TYPE=\"\\\"static\\\"\""
+    DEFINES += "UNBUNDLE=\"\\\"-\\\"\""
+} else {
+    DEFINES += "LINKING_TYPE=\"\\\"dynamic\\\"\""
+    DEFINES += "UNBUNDLE=\"\\\"$${UNBUNDLE}\\\"\""
+}
```
ubruhin, line 71: These defines are only needed for the `Application` class, right? Then I'd prefer to move them to `common.pro` instead to keep their scope as small as possible.

dev/doxygen/pages/building.md
```diff
@@ -0,0 +1,97 @@
...
+    make -j8
+
+Then run the `librepcb` library in the `build/output/` directory.
+
+To install LibrePCB on your system, run `make install`.
+
+## Faster Builds with ccache
+
+To speed up subsequent compilation steps, it's recommended to install and
+enable [ccache](https://ccache.dev/) (this will work on Qt 5.9.2 and newer):
```
ubruhin, line 54: Hmm actually ccache only helps to speed up clean rebuilds, or am I missing something? I mean, usually you don't do clean rebuilds locally, ccache is useful mainly on CI to pass build artifacts between independent build jobs...

dev/doxygen/pages/building.md
```diff
@@ -0,0 +1,97 @@
...
+
+If you want to create a release build that still contains debug info:
+
+    qmake -r ../librepcb.pro CONFIG+=release CONFIG+=force_debug_info
+
+
+# Dynamic Linking / Unbundling {#doc_building_unbundling}
+
+By default, all binaries will be linked statically using vendored libraries. If
+you would prefer to unbundle some libraries, set the `UNBUNDLE` variable:
```
ubruhin, line 85: I guess "would" is not needed here:

```suggestion
you prefer to unbundle some libraries, set the `UNBUNDLE` variable:
```

dev/doxygen/pages/building.md
```diff
@@ -0,0 +1,97 @@
...
+
+    qmake -r ../librepcb.pro UNBUNDLE+=quazip
+
+You can either list the libraries one by one, or you can use `UNBUNDLE=all` to
+unbundle all libraries that support dynamic linking.
+
+Right now, the following libraries can be unbundled:
+
+* `quazip`
```
ubruhin, line 94:
```suggestion
* `quazip`
* `fontobene-qt5`
```

ci/azure-jobs-linux.yml
```diff
@@ -57,3 +57,35 @@ jobs:
...
+- job: Linux_Unbundled
+  pool:
+    vmImage: 'ubuntu-16.04'
+  variables:
+    OS: 'linux'
+    ARCH: 'x86_64'
+  strategy:
+    matrix:
+      Ubuntu_2004_GCC:
+        IMAGE: dbrgn/librepcb-dev:ubuntu-20.04-2
```
dbrgn, line 69: Note to self: Replace with official image

README.md
```diff
@@ -127,54 +127,9 @@ git submodule update --init --recursive
...
-open ./output/librepcb.app    # Mac OS X
-output\librepcb.exe           # Windows
-```
-
-### Installation
-
-On a Unix/Linux system, LibrePCB can be installed with `sudo make install`.
+You can either build LibrePCB using Qt Creator, or you can build on the command
+line using qmake. For instructions, see
+https://developers.librepcb.org/d5/d96/doc_building.html
```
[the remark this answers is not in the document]
dbrgn: Ok, I agree!

apps/EagleImport/EagleImport.pro
```diff
@@ -12,15 +10,20 @@ include(../../common.pri)
...
+# Note: The order of the libraries is very important for the linker!
+# Another order could end up in "undefined reference" errors!
+# Note that dependencies need to be listed *after* the dependent libs.
 LIBS += \
     -L$${DESTDIR} \
     -llibrepcbeagleimport \
-    -llibrepcblibrary \    # Note: The order of the libraries is very important for the linker!
-    -llibrepcbcommon \     # Another order could end up in "undefined reference" errors!
+    -llibrepcbworkspace \
+    -llibrepcbproject \
```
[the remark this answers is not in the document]
dbrgn: Hmm, probably a copy-paste error. I'll remove them.

apps/EagleImport/EagleImport.pro
```diff
@@ -12,15 +10,20 @@ include(../../common.pri)
...
+    -llibrepcblibrary \
+    -llibrepcbcommon \
     -lparseagle \
     -lsexpresso \
     -lclipper \
-    -lquazip -lz \
+    -lmuparser \
```
dbrgn, line 26: Hm, I wonder how `-lmuparser` got in here... Maybe a rebase error.

common.pri
dbrgn: Yep, right now that's the only place where these defines are used. I'll move them.

dev/doxygen/pages/building.md
dbrgn: If you switch between branches, then that may result in recompilation of large parts of LibrePCB, especially if commonly used files, e.g. in librepcbcommon are changed. I'm not sure how much it really helps, but subjectively on my machine it seems that working on LibrePCB is faster with ccache than without.

ci/azure-jobs-linux.yml
```diff
@@ -57,3 +57,35 @@ jobs:
       UPLOAD_USER: $(ARTIFACTS_UPLOAD_USER)
       UPLOAD_PASS: $(ARTIFACTS_UPLOAD_PASS)
       UPLOAD_SIGN_KEY: $(ARTIFACTS_UPLOAD_SIGN_KEY)
+- job: Linux_Unbundled
```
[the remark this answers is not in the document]
dbrgn: That might be worth a try. I don't remember why I split up the steps (probably because I overlooked that the appimage and archive are only built conditionally).

apps/EagleImport/EagleImport.pro
dbrgn: Oh, actually it *is* required, otherwise the linker complains.

ci/azure-jobs-linux.yml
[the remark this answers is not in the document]
dbrgn: Good thing you suggested that, the libraries were not actually being installed by `make install`. It should now be fixed. I also merged the jobs by adding some conditionals.

dbrgn:
@ubruhin rebased & squashed!

common.pri
ubruhin: Now these defines are in both files, `common.pri` and `common.pro` :wink:

ci/azure-jobs-linux.yml
```diff
@@ -30,7 +33,13 @@ jobs:
       # Pass modified PATH to following steps:
       echo "##vso[task.setvariable variable=PATH]$PATH"
     displayName: Install Dependencies
-  - bash: ./ci/build_librepcb.sh
+  - bash: rm -rf libs/quazip
+    displayName: Remove unbundled quazip
+    condition: and(succeeded(), contains(variables['UNBUNDLE'], 'quazip'))
```
ubruhin, line 38: What's the reason for removing the *bundled* (not *unbundled*, as written) quazip?

ci/azure-jobs-linux.yml
```diff
@@ -30,7 +33,13 @@ jobs:
...
+  - bash: |
+      ./ci/build_librepcb.sh
+      # Set LD_LIBRARY_PATH
+      echo "##vso[task.setvariable variable=LD_LIBRARY_PATH]$(pwd)/build/install/opt/lib"
```
ubruhin, line 42: Why not simply set the `LD_LIBRARY_PATH` right in the build matrix above? Then it would even be set only on the build job with unbundled libraries...

common.pri
dbrgn: Gahhh, rebasing was a bit messy due to conflicts. I resolved one of them, but it seems that git auto-resolved the other conflict the wrong way.

ci/azure-jobs-linux.yml
dbrgn: I remove the directory to ensure that the bundled library isn't used. The text is wrong though, it should say "bundled" :slightly_smiling_face:

ci/azure-jobs-linux.yml
dbrgn: Because it needs to be an absolute path, and Azure doesn't show the full directory paths... I don't think we can rely on the checkout directory to remain a constant, right?

common.pri
ubruhin: So could you fix it now? :wink:

dev/doxygen/pages/building.md
ubruhin: OK I see, but then it would be good to state this more precisely in the documentation :wink: "To speed up subsequent compilation steps, it's recommended..." sounds like compiling is slow by default, which is simply not true. A normal build is probably even slower with ccache enabled (not measured, but ccache of course also needs some processing time).

dev/doxygen/pages/building.md
```diff
@@ -0,0 +1,99 @@
...
+
+# Building on the Command Line {#doc_building_command_line}
+
+You can also build LibrePCB from the command line using qmake:
+
+    mkdir build && cd build
+    qmake -r ../librepcb.pro
+    make -j8
+
+Then run the `librepcb` library in the `build/output/` directory.
```
ubruhin, line 47:
```suggestion
Then run the `librepcb` application in the `build/output/` directory.
```

common.pri
dbrgn: I thought I did, but edited the wrong location again... It doesn't help that both files are called `common` :see_no_evil: 

Should be fixed now.

dev/doxygen/pages/building.md
dbrgn: But isn't that exactly what's stated? "To speed up subsequent compilation steps"...

Anyways, I reworded the text to make it clearer.

dev/doxygen/pages/building.md
dbrgn: Changed to `binary`, which is what I intended to write...

ci/azure-jobs-linux.yml
ubruhin: Sure, but you can set it right in the build matrix like that:

```
LD_LIBRARY_PATH: $(Agent.BuildDirectory)/build/install/opt/lib
```

dev/doxygen/pages/building.md
ubruhin: Thanks! :slightly_smiling_face:

ci/azure-jobs-linux.yml
dbrgn: Ah, I wasn't aware of those variables! Will update the config.

dbrgn:
Cool, finally :smile: 

@FFY00 @dvzrv @FedericoCeratto @arunisaac let me know if anything is unclear, or if you run across an issue when packaging!

If you need any other library to support unbundling, also let me know, it should be quite easy to do. (But besides quazip, none of the used libraries seems to be already packaged in popular distros...)

PR #699, opened 2020-05-16
Update fontobene-qt5

dbrgn:
The updated version changes the header path.

See https://github.com/fontobene/fontobene-qt5/pull/2

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,7 +17,7 @@
     path = libs/sexpresso
     url = https://github.com/LibrePCB/sexpresso.git
 [submodule "libs/fontobene"]
-    path = libs/fontobene
+    path = libs/fontobene-qt5
     url = https://github.com/fontobene/fontobene-qt5.git
 [submodule "share/librepcb/fontobene"]
     path = share/librepcb/fontobene
--- a/libs/librepcb/common/common.pro
+++ b/libs/librepcb/common/common.pro
@@ -21,7 +21,7 @@ CONFIG += staticlib
 
 INCLUDEPATH += \
     ../../ \
-    ../../fontobene \
+    ../../fontobene-qt5 \
     ../../quazip \
     ../../sexpresso \
     ../../type_safe/include \
--- a/libs/librepcb/common/font/strokefont.cpp
+++ b/libs/librepcb/common/font/strokefont.cpp
@@ -22,8 +22,8 @@
  ******************************************************************************/
 #include "strokefont.h"
 
-#include <fontobene/font.h>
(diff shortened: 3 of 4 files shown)


dbrgn:
@ubruhin updated, tests pass now!

.gitmodules
```diff
@@ -17,7 +17,7 @@
     path = libs/sexpresso
     url = https://github.com/LibrePCB/sexpresso.git
 [submodule "libs/fontobene"]
```
ubruhin, line 19: Probably this one should be changed too?

```suggestion
[submodule "libs/fontobene-qt5"]
```

Or alternatively we could just keep the path `libs/fontobene`. Actually I think I'd prefer that solution, but `libs/fontobene-qt5` is also fine for me :)

dbrgn: The change was done this way by `git mv`, I don't dare changing it :smile: 

I'd keep the name of the submodule consistent with the library name, if you don't mind. It might also avoid some confusion between fontobene the format, fontobene the fonts and fontobene-qt5 the parser library.

ubruhin: > The change was done this way by git mv, I don't dare changing it smile

Ah, I see. I suspect by keeping the submodule name, Git recognizes that only the path has changed and therefore it can keep its internal data (commits etc.). Generally that's a good thing, but I'm not sure if it's worth to create an inconsistency between submodule name and path...

> I'd keep the name of the submodule consistent with the library name, if you don't mind.

OK, no problem :+1:

dbrgn: Ok, I removed and re-added the submodule instead!

dbrgn: Ah, nope, that won't work:

```
$ git clone --recursive https://github.com/LibrePCB/LibrePCB/
$ git branch
* master
$ git checkout fontobene-update
fatal: not a git repository: ../../.git/modules/libs/fontobene-qt5
fatal: could not reset submodule index
```

I think the problem is that the repository name is the same in both cases, so it conflicts in the  `.git/modules/`.

I'd recommend to just go with the rename, nobody cares what the identification of the git submodule is called :slightly_smiling_face:

dbrgn: I updated the branch again. Note: I would recommend *not* to check out the previous commit, it seems that I have somehow broken my local repo and cannot switch away to another branch anymore :confused:

ubruhin: Okay, so then let's keep the submodule name...

But you should make the submodule pointing to the proper commit to make CI happy :smile:

dbrgn: Whoops :smile:

PR #702, opened 2020-05-24
Move model-view classes from library to library::editor

ubruhin:
Fixes the software architecture violation reported in https://github.com/LibrePCB/LibrePCB/pull/698#issuecomment-630493839.

@dbrgn Could you check if this also fixes the linker error?


dbrgn:
> Could you check if this also fixes the linker error?

Yes, it seems to work now! :tada:

I'll cherry-pick the commit into #698 and will then push again, to verify that it's fixed in CI as well.

ubruhin:
OK, thanks for testing!

Rebased -> will merge once CI succeeded.

PR #703, opened 2020-05-25
Doxygen: Show TOC on markdown pages

dbrgn:
From the Doxygen docs[1]:

> Note that the TOC_INCLUDE_HEADINGS has to be set to a value > 0
> otherwise no table of contents is shown when using Markdown Headers.

I set the value to `2`, so that 2 levels of headings are included in
the table of contents.

[1] http://doxygen.nl/manual/markdown.html#md_toc


dbrgn:
Extracted from #698.

PR #723, opened 2020-07-04
Implement "select all" (Ctrl+A) action in all editors

ubruhin:
Adding a menu item "select all" with shortcut Ctrl+A to the symbol-, package-, schematic- and board editor.


libs/librepcb/libraryeditor/pkg/fsm/packageeditorstate_select.cpp
```diff
@@ -230,6 +230,20 @@ bool PackageEditorState_Select::processGraphicsSceneRightMouseButtonReleased(
   }
 }
 
+bool PackageEditorState_Select::processSelectAll() noexcept {
+  switch (mState) {
+    case SubState::IDLE: {
+      if (auto item = mContext.currentGraphicsItem) {
+        item->setSelectionRect(
+            item->boundingRect().adjusted(-100, -100, 100, 100));
```
dbrgn, line 238: Are these the max bounds, or just "magic numbers that should cover everything"?

ubruhin: They are "magic numbers that should cover everything" :slightly_smiling_face: But note it takes the actual bounding rect and increases it a bit just to be sure everything is really *inside* that rect. I first wrote a `selectAll()` method which does really select all items, but somehow I was not happy with adding many lines of code for something which could be solved with that simple hack :see_no_evil:

dbrgn: Ahh, I missed the expansion part. Makes sense :slightly_smiling_face:

ubruhin: I added a comment to clarify it :)

PR #724, opened 2020-07-06
SchematicEditor: Implement clipboard cut/copy/paste

ubruhin:
As usual for new, complex features: it's not perfect, but much better than nothing :wink:

- Copy&paste works within a schematic or between different schematics, even between different projects.
- The library elements of the copied items are copied into the clipboard too. But they are only pasted if the target project doesn't contain them yet.
- Under the hood, an S-Expression file containing the copied elements is generated and packed together with the library elements (which are whole directories) into a ZIP file which is then stored in the operating systems clipboard.
- When cutting schematic elements, the corresponding elements on the boards (if any) are removed, but currently aren't re-added after pasting. Of course that's a big drawback when cutting elements which are already layouted since you'd have to do parts of the layout again. But it requires quite some more effort to also restore all the removed items in the boards...
- Component names (e.g. "R5") are kept if possible. On conflicts, new names are generated.
- Currently, net lines and pins are not connected to the net items below them when pasting. 

copypaste

Required for #13, but does not yet provide copy&paste in the board editor.


dbrgn:
Awesome! :tada:

PR #727, opened 2020-07-10
Enable pluralization for net renaming string

dbrgn:
As requested by a translator on Transifex.

Hope this is correct (as per https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals).

### DIFF
--- a/libs/librepcb/projecteditor/schematiceditor/renamenetsegmentdialog.ui
+++ b/libs/librepcb/projecteditor/schematiceditor/renamenetsegmentdialog.ui
@@ -57,7 +57,7 @@
    <item>
     <widget class="QRadioButton" name="rbtnRenameWholeNet">
      <property name="text">
-      <string extracomment="Number of segments (&gt;=1)">Rename whole net (%1 segments)</string>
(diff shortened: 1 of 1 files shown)


ubruhin:
Did you test it? I suspect this doesn't work properly. When using `%n`, the number must be passed as argument to `tr()` when translating that string. But the code which calls `tr()` is automatically generated from the *.ui file, but the generator can't know what number you want to pass. I don't know if it's somehow possible to use proper pluralization in *.ui files. Maybe it needs to be solved entirely within the *.cpp file...

dbrgn:
Previously the placeholder was `%1`. Where does that come from? Wouldn't that be a number?

ubruhin:
> Previously the placeholder was %1. Where does that come from? Wouldn't that be a number?

Yes it is a number, but it is manually substituted in https://github.com/LibrePCB/LibrePCB/blob/9821f984c506870e43260fbbcce78b58ed2847da/libs/librepcb/projecteditor/schematiceditor/renamenetsegmentdialog.cpp#L96.

So the English string in the *.ui is automatically translated to the target language (still containing the placeholder `%1`) and then the placeholder within the translated string is substituted by the actual number.

PR #729, opened 2020-07-11
Fix fontobene-qt5 config in libs.pro

dbrgn:
Probably a copy paste error.

### DIFF
--- a/libs/libs.pro
+++ b/libs/libs.pro
@@ -25,7 +25,7 @@ librepcb.depends = \
     librepcb.depends += quazip
 }
 
-!contains(UNBUNDLE, quazip) {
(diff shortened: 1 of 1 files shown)


dbrgn:
Why? There doesn't seem to be any conflict :slightly_smiling_face: 

Nevertheless, updated.

ubruhin:
> Why? There doesn't seem to be any conflict

Yeah but IMHO it's still good practice to only merge up-to-date branches (having no merge conflict doesn't mean CI will succeed after merging). Thus I configured GitHub to allow merging only up-to-date branches, which lead to a scary "confirm using administrator privileges to merge" warning :grin:

...I really miss the "Rebase" button which GitLab provides on MRs :wink:

PR #736, opened 2020-07-21
Board editor: Improve the "add via" tool

5n8ke:
General UX and Code improvements for the "add via" command

# General improvements
The via is now combined with NetPoints/NetLines of its specified signal. Adding a via at the position of a pad, other via or an intersection of multiple different (!) signals is no longer possible.

# Auto Mode
In "Auto" Mode, the via is added to an existing NetSignal at the target location. Otherwise, it defaults to the signal with the most segments but should be changed to the closest signal.

# CmdSplitNetline
A new UndoCommandGroup, that splits the NetLine at the given position. The position does not have to be **on** the netline.

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
@@ -69,6 +69,7 @@ BES_AddVia::BES_AddVia(BoardEditor& editor, Ui::BoardEditor& editorUi,
     mCurrentViaSize(700000),
     mCurrentViaDrillDiameter(300000),
     mCurrentViaNetSignal(nullptr),
+    mFindClosestNetSignal(true),
     // command toolbar actions / widgets:
     mSizeLabel(nullptr),
     mSizeEdit(nullptr),
@@ -79,7 +80,7 @@ BES_AddVia::BES_AddVia(BoardEditor& editor, Ui::BoardEditor& editorUi,
 }
 
 BES_AddVia::~BES_AddVia() {
-  Q_ASSERT(!mUndoStack.isCommandGroupActive());
+  Q_ASSERT(mSubState == SubState_Idle);
 }
 
 /*******************************************************************************
@@ -99,8 +100,7 @@ bool BES_AddVia::entry(BEE_Base* event) noexcept {
   Q_UNUSED(event);
   Board* board = mEditor.getActiveBoard();
   if (!board) return false;
-  if (mEditor.getProject().getCircuit().getNetSignals().count() == 0)
-    return false;
+  if (mCircuit.getNetSignals().isEmpty()) return false;
 
   // clear board selection because selection does not make sense in this state
   board->clearSelection();
@@ -162,14 +162,12 @@ bool BES_AddVia::entry(BEE_Base* event) noexcept {
   mNetSignalComboBox->setInsertPolicy(QComboBox::NoInsert);
   mNetSignalComboBox->setEditable(false);
   foreach (NetSignal* netsignal,
-           mEditor.getProject().getCircuit().getNetSignals()) {
+           mCircuit.getNetSignals()) {
     mNetSignalComboBox->addItem(*netsignal->getName(),
                                 netsignal->getUuid().toStr());
   }
   mNetSignalComboBox->model()->sort(0);
-  mNetSignalComboBox->setInsertPolicy(QComboBox::InsertAtTop);
-  while (mEditor.getProject().getCircuit()
-         .getNetSignalByName(mAutoText)) {
+  while (mCircuit.getNetSignalByName(mAutoText)) {
     mAutoText = "[" + mAutoText + "]";
   }
   mNetSignalComboBox->addItem(mAutoText);
@@ -182,12 +180,12 @@ bool BES_AddVia::entry(BEE_Base* event) noexcept {
             if (value == mAutoText) {
               mCurrentViaNetSignal = nullptr;
             } else {
-              mCurrentViaNetSignal = mEditor.getProject().getCircuit()
-                  .getNetSignalByName(value);
+              mCurrentViaNetSignal = mCircuit.getNetSignalByName(value);
               setNetSignal(mCurrentViaNetSignal);
             }
           });
 
+  mLastClosestNetSignal = mCircuit.getNetSignalWithMostElements();
   // add a new via
   return addVia(*board);
 }
@@ -240,7 +238,7 @@ BES_Base::ProcRetVal BES_AddVia::processSceneEvent(BEE_Base* event) noexcept {
                       .mappedToGrid(board->getGridProperties().getInterval());
       switch (sceneEvent->button()) {
         case Qt::LeftButton: {
-          fixVia(pos);
+          fixVia(*board, pos);
           addVia(*board);
           updateVia(*board, pos);
           return ForceStayInState;
@@ -293,11 +291,11 @@ BES_Base::ProcRetVal BES_AddVia::processSceneEvent(BEE_Base* event) noexcept {
 
 bool BES_AddVia::addVia(Board& board) noexcept {
   Q_ASSERT(mSubState == SubState_Idle);
-  Q_ASSERT(!mUndoStack.isCommandGroupActive());
 
   try {
-    mSubState = SubState_PositioningVia;
     mUndoStack.beginCmdGroup(tr("Add via to board"));
+    Q_ASSERT(mUndoStack.isCommandGroupActive());
+    mSubState = SubState_PositioningVia;
     CmdBoardNetSegmentAdd* cmdAddSeg = nullptr;
     if (mCurrentViaNetSignal) {
       cmdAddSeg = new CmdBoardNetSegmentAdd(board, *mCurrentViaNetSignal);
@@ -329,7 +327,6 @@ bool BES_AddVia::addVia(Board& board) noexcept {
 
 void BES_AddVia::abortPlacement(const bool showErrorMessage) noexcept {
   if (mSubState != SubState_Idle) {
-    Q_ASSERT(mUndoStack.isCommandGroupActive());
     try {
       mUndoStack.abortCmdGroup();
     } catch (Exception& e) {
@@ -338,7 +335,6 @@ void BES_AddVia::abortPlacement(const bool showErrorMessage) noexcept {
       }
     }
   }
-  Q_ASSERT(!mUndoStack.isCommandGroupActive());
   mSubState = SubState_Idle;
 }
 
@@ -359,23 +355,22 @@ bool BES_AddVia::updateVia(Board& board, const Point& pos) noexcept {
   }
 }
 
-bool BES_AddVia::fixVia(Point& pos) noexcept {
+bool BES_AddVia::fixVia(Board& board, const Point& pos) noexcept {
   Q_ASSERT(mSubState == SubState_PositioningVia);
   //TODO(5n8ke): handle user errors in a more graceful way without popup message
 
   try {
     mViaEditCmd->setPosition(pos, false);
 
-    Board* board = &mCurrentVia->getBoard();
     NetSignal* netsignal = mCurrentViaNetSignal;
     if (!netsignal) {
-        QSet<NetSignal*> netsignals = getNetSignalsAtScenePos(*board, pos,
+        QSet<NetSignal*> netsignals = getNetSignalsAtScenePos(board, pos,
         {mCurrentVia});
       if (netsignals.count() > 1) {
         throw RuntimeError(__FILE__, __LINE__,
                            tr("Multiple different signals at via position."));
       } else if (netsignals.count() == 0) {
-        netsignal = getClosestNetSignal(*board, pos);
+        netsignal = getClosestNetSignal(board, pos);
       } else {
         netsignal = netsignals.values().first();
       }
@@ -386,7 +381,7 @@ bool BES_AddVia::fixVia(Point& pos) noexcept {
 
     // Find stuff at the via position
     QSet<BI_NetPoint*> otherNetAnchors = {};
-    if (BI_Via* via = findVia(*board, pos, nullptr, {mCurrentVia})) {
+    if (BI_Via* via = findVia(board, pos, nullptr, {mCurrentVia})) {
       if (&via->getNetSignalOfNetSegment() != netsignal) {
         throw RuntimeError(__FILE__, __LINE__,
           tr("Via of a different signal already present at target position."));
@@ -394,7 +389,7 @@ bool BES_AddVia::fixVia(Point& pos) noexcept {
         abortPlacement();
         return true;
       }
-    } else if (BI_FootprintPad* pad = findPad(*board, pos)) {
+    } else if (BI_FootprintPad* pad = findPad(board, pos)) {
       if (pad->getCompSigInstNetSignal() != netsignal) {
         throw RuntimeError(__FILE__, __LINE__,
           tr("Pad of a different signal already present at target position."));
@@ -403,7 +398,7 @@ bool BES_AddVia::fixVia(Point& pos) noexcept {
         return true;
       }
     }
-    foreach (BI_NetPoint* netpoint, board->getNetPointsAtScenePos(pos)) {
+    foreach (BI_NetPoint* netpoint, board.getNetPointsAtScenePos(pos)) {
       if (&netpoint->getNetSignalOfNetSegment() != netsignal) {
         throw RuntimeError(__FILE__, __LINE__,
           tr("Netpoint of a different signal already present"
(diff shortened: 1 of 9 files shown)


ubruhin:
Nice! Some comments so far:

- If the schematic is empty (and thus no net signals exist), the application crashes when starting the "add via" tool. Previously we rejected starting the tool if no net signals exist to avoid this issue.
- When the net signal is set to "auto", performance is heavily reduced (at least on a larger board with many elements). It seems that the calculations in the MouseMoveEvent are quite heavy :( It might be worth considering a more efficient way to determine the closest netsignal. However, I didn't test it with a release build yet, but only with a debug build.
- I'd prefer to rename the `CmdSplitNetLine` class to something containing the term "Board" to make clear it only splits *board* netlines.
- Removing `mUndoCmdActive` and instead calling `mUndoStack.isCommandGroupActive()` is problematic since now you don't know whether the active undo command group was started by the "add via" tool, or by some other tool. It could lead to serious issues if another tool started the undo command group, but the "add via" tool "thinks" it was started by itself.
- If there exists a net with name "Auto", choosing that signal in the dropdown menu doesn't enforce the corresponding net because the "auto net" feature only matches the text in the dropdown menu :wink:

5n8ke:
@ubruhin 
Thanks for your feedback. I addressed almost all of your comments.
The low performance of the auto command is not yet considered. I think for now the addition of the auto mode is useful and an improved signal search algorithm would be part of a new project.

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
```diff
@@ -283,45 +292,65 @@ BES_Base::ProcRetVal BES_AddVia::processSceneEvent(BEE_Base* event) noexcept {
 }
 
 bool BES_AddVia::addVia(Board& board) noexcept {
-  Q_ASSERT(mUndoCmdActive == false);
-  Q_ASSERT(mCurrentViaNetSignal);
+  Q_ASSERT(mSubState == SubState_Idle);
+  Q_ASSERT(!mUndoStack.isCommandGroupActive());
```
ubruhin, line 296: This assertion fails (and thus LibrePCB crashes) if another tool (for example in the schematic editor) is active when the "add via" tool is started. It's totally legal to try starting a tool while another command is active, thus this assertion should be removed:

```suggestion
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
```diff
@@ -283,45 +292,65 @@ BES_Base::ProcRetVal BES_AddVia::processSceneEvent(BEE_Base* event) noexcept {
...
 
   try {
+    mSubState = SubState_PositioningVia;
     mUndoStack.beginCmdGroup(tr("Add via to board"));
```
ubruhin, line 300: The state must be set *after* starting the new undo command, since starting a command raises an exception if there is already a command active. Then we don't want to be in `SubState_PositioningVia` state since this would make `abortPlacement()` aborting a command which was started by another tool.

```suggestion
    mUndoStack.beginCmdGroup(tr("Add via to board"));
    mSubState = SubState_PositioningVia;
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
```diff
@@ -283,45 +292,65 @@ BES_Base::ProcRetVal BES_AddVia::processSceneEvent(BEE_Base* event) noexcept {
...
+    Q_ASSERT(mUndoStack.isCommandGroupActive());
+    try {
+      mUndoStack.abortCmdGroup();
+    } catch (Exception& e) {
+      if (showErrorMessage) {
+        QMessageBox::critical(&mEditor, tr("Error"), e.getMsg());
+      }
+    }
+  }
+  Q_ASSERT(!mUndoStack.isCommandGroupActive());
```
ubruhin, line 341: Same here - if we are not in `SubState_PositioningVia`, we must not rely on whether no command is active currently.

```suggestion
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
```diff
@@ -330,22 +359,115 @@ bool BES_AddVia::updateVia(Board& board, const Point& pos) noexcept {
...
-bool BES_AddVia::fixVia(const Point& pos) noexcept {
-  Q_ASSERT(mUndoCmdActive == true);
+bool BES_AddVia::fixVia(Point& pos) noexcept {
+  Q_ASSERT(mSubState == SubState_PositioningVia);
+  //TODO(5n8ke): handle user errors in a more graceful way without popup message
 
   try {
     mViaEditCmd->setPosition(pos, false);
+
+    Board* board = &mCurrentVia->getBoard();
```
ubruhin, line 369: That looks strange - if you get a reference to a board, you should store it as a reference since this way it's clear that it can never be null.

```suggestion
    Board& board = mCurrentVia->getBoard();
```

But actually I would even pass `Board& board` as a parameter to the `fixVia()` method (just like `updateVia()`) instead of fetching it from the currently placed via. The only function which calls `fixVia()` even already has a pointer to the board.

libs/librepcb/projecteditor/cmd/cmdboardsplitnetline.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+/*******************************************************************************
+ *  Namespace
+ ******************************************************************************/
+namespace librepcb {
+namespace project {
+namespace editor {
+
+CmdBoardSplitNetLine::CmdBoardSplitNetLine(BI_NetLine& netline, Point& pos)
+  noexcept : UndoCommandGroup(tr("Split netline")),
```
ubruhin, line 42: For translated strings, we don't use the term "netline" since translators and users might not know what it means. Instead, for board netlines we use the term "trace".

```suggestion
  noexcept : UndoCommandGroup(tr("Split trace")),
```

libs/librepcb/projecteditor/cmd/cmdboardsplitnetline.cpp
```diff
@@ -0,0 +1,81 @@
...
+                     mOldNetLine.getLayer(), mOldNetLine.getWidth());
+  cmdAdd->addNetLine(*mSplitPoint, mOldNetLine.getEndPoint(),
+                     mOldNetLine.getLayer(), mOldNetLine.getWidth());
+  QScopedPointer<CmdBoardNetSegmentRemoveElements> cmdRemove(
+        new CmdBoardNetSegmentRemoveElements(mOldNetLine.getNetSegment()));
+  cmdRemove->removeNetLine(mOldNetLine);
+  execNewChildCmd(cmdAdd.take()); // can throw
+  execNewChildCmd(cmdRemove.take()); // can throw
+
+  undoScopeGuard.dismiss();  // no undo required
```
ubruhin, line 71: FYI, since this undo command group does not contain logic which depends on the executed commands, it would be simpler (and thus possibly safer) to append the commands with `appendChild()` instead of `execNewChildCmd()` and then calling `UndoCommandGroup::performExecute()` at the end of the function. This way you don't need the undo scope guard at all. See for example `CmdRemoveSelectedSymbolItems`.

But it should work this way, so no need to change it now...

libs/librepcb/projecteditor/cmd/cmdboardsplitnetline.h
```diff
@@ -0,0 +1,76 @@
...
+namespace editor {
+
+/*******************************************************************************
+ *  Class CmdBoardNetLineSplit
+ ******************************************************************************/
+
+class CmdBoardSplitNetLine : public UndoCommandGroup {
+public:
+  // Constructors / Destructor
+  explicit CmdBoardSplitNetLine(BI_NetLine& netline, Point& pos) noexcept;
```
ubruhin, line 52: The `Point` should be passed as a const reference:

```suggestion
  explicit CmdBoardSplitNetLine(BI_NetLine& netline, const Point& pos) noexcept;
```

libs/librepcb/projecteditor/cmd/cmdboardsplitnetline.h
```diff
@@ -0,0 +1,76 @@
...
+  BI_NetPoint* getSplitPoint() noexcept { return mSplitPoint; };
+
+private: // Methods
+
+  /// @copydoc UndoCommand::performExecute()
+  bool performExecute() override;
+
+  // Private Member Variables
+  BI_NetLine& mOldNetLine;
+  Point& mSplitPosition;
```
ubruhin, line 64: This looks quite dangerous. The `Point` object passed to the constructor will most probably have a shorter object lifetime than this undo command object, which will lead to dangling references and thus application crashes. It's much safer to store it by value. But since it is actually never accessed at all, let's remove it completely:

```suggestion
```

libs/librepcb/projecteditor/cmd/cmdboardsplitnetline.h
```diff
@@ -0,0 +1,76 @@
...
+namespace project {
+
+class BI_NetLine;
+class BI_NetPoint;
+class CmdBoardNetSegmentAddElements;
+
+namespace editor {
+
+/*******************************************************************************
+ *  Class CmdBoardNetLineSplit
```
ubruhin, line 46: Inconsistent naming:

```suggestion
 *  Class CmdBoardSplitNetLine
```

libs/librepcb/projecteditor/cmd/cmdboardsplitnetline.h
```diff
@@ -0,0 +1,76 @@
...
+ ******************************************************************************/
+
+class CmdBoardSplitNetLine : public UndoCommandGroup {
```
ubruhin, line 49: For consistency, at least a very short doxygen comment should be added (I guess this would really be the first class with no doxygen comment at all):

```suggestion
/**
 * @brief Undo command to split a ::librepcb::project::BI_NetLine
 */
class CmdBoardSplitNetLine : public UndoCommandGroup {
```

ubruhin:
> The low performance of the auto command is not yet considered. I think for now the addition of the auto mode is useful and an improved signal search algorithm would be part of a new project.

Now I just tested it with a release build (on a decent PC) - unfortunately the board editor still lags on a medium complex board. I suspect on slower computers or with more complex boards, it will lag so much that it would be very annoying.

Therefore I'm really sorry, but I don't agree that the performance issue could be fixed at any time later. I'd really prefer to (at least try to) fix the performance issues right in this PR. But instead of optimizing the netsignal finder algorithm, for now we might simply run that algorithm less often. Example:

- When the "auto" mode is selected, start a `QTimer` with e.g. 100ms interval (to be determined by experiments)
- In the MouseMove event, do not immediately find the closest netsignal, but only set a flag to do so
- Within the timer callback, check that flag - if it is set, find the closest net signal, update the via and clear the flag.

I suspect this would at least avoid having a totally lagging UI, while keeping the implementation effort low. What do you think about that?

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
```diff
@@ -71,7 +79,7 @@ BES_AddVia::BES_AddVia(BoardEditor& editor, Ui::BoardEditor& editorUi,
 }
 
 BES_AddVia::~BES_AddVia() {
-  Q_ASSERT(mUndoCmdActive == false);
+  Q_ASSERT(!mUndoStack.isCommandGroupActive());
```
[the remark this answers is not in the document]
5n8ke: Thanks. I fixed it.

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
5n8ke: Will fix it there too

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
```diff
@@ -91,19 +99,12 @@ bool BES_AddVia::entry(BEE_Base* event) noexcept {
   Q_UNUSED(event);
   Board* board = mEditor.getActiveBoard();
   if (!board) return false;
+  if (mEditor.getProject().getCircuit().getNetSignals().count() == 0)
```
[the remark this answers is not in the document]
5n8ke: Found also some other places with the same issue.

libs/librepcb/projecteditor/cmd/cmdboardsplitnetline.cpp
5n8ke: Thanks for the info. Since I'm still working on it, I will change it.

libs/librepcb/projecteditor/cmd/cmdboardsplitnetline.h
5n8ke: I missed that...

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
5n8ke: Done.

5n8ke:
Thanks for the many comments. I fixed them all.

I improved the finder a bit by not always calculating the most used NetSignal, but instead storing the information in mLastClosestNetSignal. I also implemented a timer, that allows the closestNetSignal to only be calculated every 500ms. It still causes a noticeable choppiness but only twice a second :/ ...
The finder algorithm only stores a new NetSignal, when it is directly found below the cursor, as previously.

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
```diff
@@ -368,31 +485,116 @@ void BES_AddVia::sizeEditValueChanged(const PositiveLength& value) noexcept {
...
 }
 
 void BES_AddVia::setNetSignal(NetSignal* netsignal) noexcept {
+  Q_ASSERT(mSubState == SubState_PositioningVia);
   try {
-    if (!netsignal) throw LogicError(__FILE__, __LINE__);
-    mCurrentViaNetSignal = netsignal;
+    if (!netsignal) {
+      throw LogicError(__FILE__, __LINE__);
+    }
```
ubruhin, line 501: During testing, sometimes the via position was not updated at all when moving the cursor, which seemed to be very strange since `setPosition()` was called properly. After some debugging, it turned out that the `setNetSignal()` is now called every time in the MouseMove event, which constantly removes the whole net segment and re-adds it immediately, which seems to lead to graphical issues (in my case, it was just never repainted, although I'd expect flickering...). Updating the net signal only when needed, fixed that issue:

```suggestion
    }
    if (netsignal == &mCurrentVia->getNetSignalOfNetSegment()) {
      return;  // net signal is already correct, no update needed
    }
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
```diff
@@ -368,31 +485,116 @@ void BES_AddVia::sizeEditValueChanged(const PositiveLength& value) noexcept {
...
+  // Otherwise the last candidate is returned.
+  if (mFindClosestNetSignal) {
+    if (BI_NetLine* atPosition = findNetLine(board, pos)) {
+      mLastClosestNetSignal = &atPosition->getNetSignalOfNetSegment();
+    }
+    mFindClosestNetSignal = false;
+    QTimer *timer = new QTimer(this);
+    connect(timer, &QTimer::timeout, [this]() {
+      this->mFindClosestNetSignal = true;
+    });
```
ubruhin, line 558: This regularly instantiates new `QTimer` objects which will be deleted only when closing the board editor. So the memory consumption constantly increases. I'd suggest to either re-use the same timer instance (by storing it as a member variable), or delete it after it expired:

```suggestion
    QTimer *timer = new QTimer(this);
    connect(timer, &QTimer::timeout, [this, timer]() {
      this->mFindClosestNetSignal = true;
      timer->deleteLater();
    });
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
```diff
@@ -330,22 +355,114 @@ bool BES_AddVia::updateVia(Board& board, const Point& pos) noexcept {
...
+             "at target position."));
+      } else {
+        otherNetAnchors.insert(netpoint);
+      }
+    }
+    foreach (BI_NetLine* netline, board.getNetLinesAtScenePos(pos)) {
+      if (&netline->getNetSignalOfNetSegment() != netsignal) {
+        throw RuntimeError(__FILE__, __LINE__,
+          tr("Netline of a different signal already present"
+             "at target position."));
```
ubruhin, line 414: Missing space:

```suggestion
          tr("Netline of a different signal already present "
             "at target position."));
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_addvia.cpp
```diff
@@ -330,22 +355,114 @@ bool BES_AddVia::updateVia(Board& board, const Point& pos) noexcept {
...
+      } else {
+        abortPlacement();
+        return true;
+      }
+    }
+    foreach (BI_NetPoint* netpoint, board.getNetPointsAtScenePos(pos)) {
+      if (&netpoint->getNetSignalOfNetSegment() != netsignal) {
+        throw RuntimeError(__FILE__, __LINE__,
+          tr("Netpoint of a different signal already present"
+             "at target position."));
```
ubruhin, line 405: Missing space:

```suggestion
          tr("Netpoint of a different signal already present "
             "at target position."));
```

PR #737, opened 2020-07-23
Board Editor: Simplify Trace command

### LINKED ISSUE #710: Board Editor : Remove unnecessary nodes of wires when extending them
##### SUMMARY
When you draw a wire in multiple actions, unnecessary nodes remain there.

image

##### SOLUTION
During drawing the first segment of wire that start on another wire of same net and if segments are aligned, remove the start node as it is not useful.
Idem on last segment.

5n8ke:
A command that simplifies all the NetSegments of a NetSignal by removing duplicate items and combining disconnected parts of a NetSegment.
# Features
## Summary
- Combine NetPoints at the same location
- Replace NetPoints with Vias at the same location
- Combine NetLines forming a straight line
- Combine NetPoints and Vias with underlying NetLines
- Remove duplicate NetLines connecting the same two anchors

## Principle
1. Select a NetSignal to be simplified.
2. **(TODO)** Connect disjunct NetSegments
3. Simplify every NetSegment of the NetSignal
   1. Remove duplicate NetLines
   2. Combine duplicate NetPoints at the same location and layer
   3. Connect NetPoints with NetLines crossing their location
   4. **(TODO)** Connect NetLines crossing each other
   5. For every Via
      1. Replace NetPoints at the Via location with the Via
      2. Connect the Via with NetLines crossing its location
   6. Remove Duplicate NetLines
   7. Combine NetLines forming a straight line (Fixes #710)

During the operation, duplicate NetLines and NetPoints are combined multiple times, since some of the more complex simplifications can create these duplicate items.

# TODO
- Replace NetPoints with Pads at the same location
- Connect disjunct NetSegments
- Connect crossing traces
- Connect planes

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -16,3 +16,4 @@ yourself by creating a pull request (see [CONTRIBUTING.md](CONTRIBUTING.md)).
 - [@chrisgwerder](https://github.com/chrisgwerder)
 - [@0xB767B](https://github.com/0xB767B)
 - Josua Schmid ([@schmijos](https://github.com/schmijos))
+- Lucas Keune ([@5n8ke](https://github.com/5n8ke))
--- a/libs/librepcb/project/boards/items/bi_netpoint.cpp
+++ b/libs/librepcb/project/boards/items/bi_netpoint.cpp
@@ -140,11 +140,17 @@ void BI_NetPoint::removeFromBoard() {
 }
 
 void BI_NetPoint::registerNetLine(BI_NetLine& netline) {
-  if ((!isAddedToBoard()) || (mRegisteredNetLines.contains(&netline)) ||
-      (&netline.getNetSegment() != &mNetSegment) ||
-      ((mRegisteredNetLines.count() > 0) &&
-       (&netline.getLayer() != getLayerOfLines()))) {
-    throw LogicError(__FILE__, __LINE__);
+  if (!isAddedToBoard()) {
+    throw LogicError(__FILE__, __LINE__, "NetPoint is not added to the board.");
+  } else if (mRegisteredNetLines.contains(&netline)) {
+    throw LogicError(__FILE__, __LINE__,
+                     "NetLine already registered to NetPoint.");
+  } else if (&netline.getNetSegment() != &mNetSegment) {
+    throw LogicError(__FILE__, __LINE__,
+                     "NetLine is part of a different NetSegment.");
+  } else if ((mRegisteredNetLines.count() > 0) &&
+             (&netline.getLayer() != getLayerOfLines())) {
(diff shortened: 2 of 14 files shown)


ubruhin:
Are you actually waiting for my comments on this PR, or is it work in progress on your side? So far I didn't really review it because it's still a draft :see_no_evil: 

This will for sure conflict with #760, but hopefully it shouldn't be too hard to resolve since it mainly "only" adds an additional tool (not modifying existing tools) :slightly_smiling_face:

5n8ke:
It currently a work in progress on my side. If you've got time, I'd be happy about some initial comments but right now I've got enough stuff to work on.

ubruhin:
I quickly tested this feature. The simplification seems to work fine! :+1: However, I have some thoughts:

- Actually I don't understand why we need a separate tool for this feature. Wouldn't it make more sense to integrate it in the already existing tools?
  - In the "Draw Trace" tool, just automatically simplify the modified net segment after finishing a trace. Maybe this feature could be made configurable whether it is active or not, but so far I don't see a case where we don't want to simplify traces.
  - In the "Select" tool, a context menu entry for traces could be added to simplify the net segment of the selected trace.
- Why simplifying all net segments of a given net signal? Since net segments are independent from each other, I'd only simplify single net segments...
- Regarding implementation, it might make sense to create a new undo command class like `CmdSimplifyBoardNetSegment` (maybe replacing `CmdBoardCombineAnchors`?) to easily share the feature between different tools.

dbrgn:
> but so far I don't see a case where we don't want to simplify traces

When drawing a straight trace, you might want to place an anchor point at a specific location, so that you can snap to that location from another trace.

ubruhin:
> When drawing a straight trace, you might want to place an anchor point at a specific location, so that you can snap to that location from another trace.

Hmm but you can also connect a trace to another trace without any anchor point :thinking:

5n8ke:
Thank you very much for the feedback!
- I have no strong opinions about simplification being its own tool or part of the Select/draw tool.
- I second the implementation as a Commandgroup. It will be done with my next round of improvements.
- I simplified all NetSegments of a signal, since as a second (or preferably first) stage, I envisioned the connection of technically disjunct, but visually connected NetSegments.

dbrgn:
> Hmm but you can also connect a trace to another trace without any anchor point

Yes, but then it snaps to the grid, not to the existing anchor point. If you draw the second trace with a different grid setting than the first one, having an anchor point for snapping might be desirable.

Nevertheless, I'm not sure how important that really is... It can be worked around by changing the grid again.

ubruhin:
Closing this in favor of #1591.

PR #738, opened 2020-07-30
CI: Migrate Windows job from AppVeyor to Azure Pipelines

### LINKED ISSUE #462: CI: Migrate Windows jobs from AppVeyor to Azure Pipelines
##### SUMMARY

#458 migrated Linux/MacOS jobs from Travis-CI to Azure Pipelines, but not the Windows jobs. It would be nice to migrate them to Azure too to avoid having multiple CI systems.

##### SOLUTION

- [ ] Create a Windows Docker image and push it to https://hub.docker.com/r/librepcb/librepcb-dev
- [ ] Add Windows job(s) to [`ci/azure-pipelines.yml`](https://github.com/LibrePCB/LibrePCB/blob/master/ci/azure-pipelines.yml)
- [ ] Remove [`appveyor.yml`](https://github.com/LibrePCB/LibrePCB/blob/master/appveyor.yml) (and remove other references to AppVeyor if there are some)

ubruhin:
Advantages:
- It works (compared to AppVeyor) :wink:
- Allow using recent Qt and compiler versions. With AppVeyor this was not possible because after such a large modification we hit the maximum build time limit as ccache wouldn't speed up the build.
- Providing our own build environment instead of relying on the tools provided by AppVeyor (which has 52385 tools installed which we don't even need) -> more reproducible build.
- (Hopefully) get rid of the random failures and suddenly broken CI due to issues with MSYS or other tools installed on AppVeyor.

Disadvantages:
- <del>Currently CI takes much longer (~1.5h instead of ~40min).</del> I set up ccache, now CI time is around 45 minutes.
- The already very unreliable unit test `SQLiteDatabaseTest` now is completely unable to success. I spent many hours to find the root cause, but didn't find a better "solution" than reducing the number of parallel thread workers :hankey: 

ToDo:
- [x] Merge https://github.com/LibrePCB/docker-librepcb-dev/pull/7 and create a Git tag
- [x] Test if the built binaries work properly. I was not able (yet) to get DLLs like `msvcp120.dll`, or `ssleay32.dll` or `zlib1.dll` bundled together with our binaries. But actually I don't know if they are even needed...

Fixes #462


ubruhin:
I set up `ccache` to speed up builds and fixed several deployment issues. Now I would say everything works fine (tested the Windows builds on a fresh Windows installation) :rocket: Time to get rid of AppVeyor! :smiley:

tests/unittests/common/sqlitedatabasetest.cpp
```diff
@@ -247,8 +246,13 @@ TEST_F(SQLiteDatabaseTest, testConcurrentAccessFromMultipleThreads) {
         startWorkerThread(&pool, READING | TRANSACTION, 10000);
     QFuture<WorkerResult> r3 =
         startWorkerThread(&pool, READING | NO_TRANSACTION, 10000);
+#if (!defined(Q_OS_WIN32)) && (!defined(Q_OS_WIN64))
+    // 4 threads seem to be too much for unreliable operating systems. Thus
+    // on unreliable, less known, rarely used niche operating systems we only
+    // use 3 reader threads...
```
rnestler, line 252: So much cynicism in 3 lines :wink:

ubruhin: Yep, after many hours of struggling with such annoying Windows issues, I have to calm down somehow and it turned out writing such comments helps to do so :grin:

PR #745, opened 2020-08-03
ERC: Disregard schematic-only components in net pin count

### LINKED ISSUE #739: Schematic Editor : Raise a warning if one net is connected to only one pin
##### SUMMARY
I usually use nets to connect pins in schematic to make it more clear.
Unfortunately, there's a risk to miss an unconnected net.
(unfortunately, I experienced that !)

##### SOLUTION
Raise a warning if one net is connected to only one pin.

5n8ke:
Only components that are not schematic only are counted when determining if a NetSignal is connected to less than two pins.

- Fixes #739

### DIFF
--- a/libs/librepcb/project/circuit/netsignal.cpp
+++ b/libs/librepcb/project/circuit/netsignal.cpp
@@ -27,9 +27,11 @@
 #include "../erc/ercmsg.h"
 #include "../schematics/items/si_netsegment.h"
 #include "circuit.h"
+#include "componentinstance.h"
 #include "componentsignalinstance.h"
 #include "netclass.h"
 
+#include <librepcb/library/cmp/component.h>
 #include <librepcb/common/exceptions.h>
 
 #include <QtCore>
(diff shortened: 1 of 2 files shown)


libs/librepcb/project/circuit/netsignal.cpp
```diff
@@ -263,7 +265,13 @@ void NetSignal::updateErcMessages() noexcept {
     mErcMsgUnusedNetSignal.reset();
   }
 
-  if (mIsAddedToCircuit && (mRegisteredComponentSignals.count() < 2)) {
+  int registeredRealComponentCount = 0;//mRegisteredComponentSignals.count();
```
ubruhin, line 268: Instead of commented out code, I'd prefer a short comment why not all component signals are counted :wink:

```suggestion
  // Raise a warning if the net signal is connected to less then two component
  // signals. But do not count component signals of schematic-only components
  // since these are just "virtual" connections, i.e. not represented by a
  // real pad (see https://github.com/LibrePCB/LibrePCB/issues/739).
  int registeredRealComponentCount = 0;
```

ubruhin:
Makes sense, thanks! :+1:

PR #746, opened 2020-08-05
Board editor: Don't remove complete trace when changing device

### LINKED ISSUE #616: Board editor: Don't remove complete trace when changing device
##### SUMMARY
When you change of device in the board (right click -> change device), it deletes all traces connected to the device, including vias and goes far.
When it is a integrated circuit with also gnd and vcc, it suppresses a lot of things.

##### SOLUTION
Just delete the last segment of each trace connected to the device.
If the user needs more suppression, he will do it by hand with "remove trace" or "remove segment".

5n8ke:
- Fixes #616 
- Replace pad with NetPoints, one for each used layer, when more than one NetLine is connected
- Remove unused includes


ubruhin:
Looks good, thanks! :+1:

Issue #610, opened 2019-12-01
Components didn't follow the grid


petus:
Hi,
at first, thanks for your job. The PCB CAD looks better and better.
Now, I installed v 0.1.3 and I made example board with capacitors and resistors and I find out that the resistor didn't follow the grid.

![librePCB_0 1 3](https://user-images.githubusercontent.com/8709019/69915907-a833d900-1454-11ea-96f3-ef5015076b0e.JPG)

Windows 10.

doom-fr:
Right click on the component and choose "snap to the grid"

petus:
OK, goog point. Thanks, but it doesn't make sense to do it with each component

doom-fr:
Yes. This can be an improvement to add a snap grid to components that are selected

PR #748, opened 2020-08-06
Board editor: Refactor and improve context menu

### LINKED ISSUE #536: Selection of stacked items
##### SUMMARY
In Library Package editor (but maybe elsewhere), when there's different items stacked, you can only select the upper one.

image

For example, here, I can't select the Documentation Layer rectangle of the pin as it is completely inside the pad on Top Layer.

I have to move the pad away, select the pin and do what I have to do and move the pad back.

##### SOLUTION
2 different solutions : 
1. when you select a point with stacked items, open a menu to choose the one you need
2. when you select a point with stacked items, firstly select the one that is inside others

Any of this solution will solve the problem.
And a mix of both will be perfect !

(layer filtering can also solve this problem but will not be user friendly in Board Editor)

5n8ke:
Previously the menu was build and executed for each different item type separately. This resulted in duplicate code and increased effort when adding an already established action to an item. This PR mainly condenses duplicate code and creates an easy framework for new actions.

# Principle
A range of standard actions are already implemented, that can be used for building the context menu (remove, rotate, flip, snap, ...).
If an item needs a custom action, it must be defined locally and a callback must be connected to the QAction::triggered signal.

# Changes
- Extract and condense code for identical actions
- Add context menu for NetPoints
- Provide functions to create menu entries for some standard actions
- When holding Shift, while clicking on a position with multiple items, cycle through the available items
- Always look if any (not just the first) item at a position was previously selected (Fixes #536)
- create generic openPropertiesDialog function, that handles the type discrimination
- Add option for CmdDragSelectedBoardItems to move without restrictions by the grid
- Add isOnGrid() for Point

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
@@ -258,37 +258,42 @@ BES_Base::ProcRetVal BES_Select::processIdleSceneRightMouseButtonReleased(
   // If the right-clicked element is part of an active selection, keep it as-is.
   // However, if it's not part of an active selection, clear the selection and
   // select the right-clicked element instead.
-  if (!items.first()->isSelected()) {
+  BI_Base* selectedItem = items.first();
+  if (!selectedItem->isSelected()) {
     board->clearSelection();
-    items.first()->setSelected(true);
+    selectedItem->setSelected(true);
   }
 
   // build and execute the context menu
   QMenu menu;
-  switch (items.first()->getType()) {
+  QAction* aRemove = nullptr;
+  QAction* aRemoveTrace = nullptr;
+  QAction* aSelectTrace = nullptr;
+  QAction* aRotateCCW = nullptr;
+  QAction* aFlipH = nullptr;
+  QAction* aSnapToGrid = nullptr;
+  QAction* aProperties = nullptr;
+  QAction* action = nullptr;
+  bool menuWasExecuted = false;
+  switch (selectedItem->getType()) {
     case BI_Base::Type_t::Footprint: {
-      BI_Footprint* footprint = dynamic_cast<BI_Footprint*>(items.first());
+      BI_Footprint* footprint = dynamic_cast<BI_Footprint*>(selectedItem);
       Q_ASSERT(footprint);
       BI_Device&         devInst = footprint->getDeviceInstance();
       ComponentInstance& cmpInst = devInst.getComponentInstance();
 
       // build the context menu
-      QAction* aRotateCCW =
-          menu.addAction(QIcon(":/img/actions/rotate_left.png"), tr("Rotate"));
-      QAction* aFlipH = menu.addAction(
-          QIcon(":/img/actions/flip_horizontal.png"), tr("Flip"));
-      QAction* aRemove =
-          menu.addAction(QIcon(":/img/actions/delete.png"),
-                         QString(tr("Remove %1")).arg(*cmpInst.getName()));
+      aRotateCCW = ACTION_ROTATE_DEFAULT(menu);
+      aFlipH = ACTION_FLIP_DEFAULT(menu);
+      aRemove = ACTION_DELETE(menu, tr("Remove %1").arg(*cmpInst.getName()));
       menu.addSeparator();
-      QAction* aSnapToGrid =
-          menu.addAction(QIcon(":/img/actions/grid.png"), tr("Snap to grid"));
-      aSnapToGrid->setVisible(devInst.getPosition().mappedToGrid(
-                                  board->getGridProperties().getInterval()) !=
-                              devInst.getPosition());  // only show if needed
+      aSnapToGrid = ACTION_SNAP_DEFAULT(menu);
+      aSnapToGrid->setVisible(!devInst.getPosition()
+                  .isOnGrid(board->getGridProperties())); // only show if needed
       QAction* aResetTexts =
-          menu.addAction(QIcon(":/img/actions/undo.png"), "Reset all texts");
+        menu.addAction(QIcon(":/img/actions/undo.png"), tr("Reset all texts"));
       menu.addSeparator();
+
       QMenu* aChangeDeviceMenu =
           menu.addMenu(QIcon(":/img/library/device.png"), tr("Change Device"));
       foreach (const DeviceMenuItem& item, getDeviceMenuItems(cmpInst)) {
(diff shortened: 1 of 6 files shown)


ubruhin:
Generally I like the idea to refactor the context menu :+1: But to be honest, I don't like the macros :wink: IMHO in C++, preprocessor macros should be used almost never - for most use-cases, a better solution (without macros) exists. Actually I'm not even sure if the `tr()` calls within the macros are working as expected or if it could lead to translation issues.

For now, I see two better solutions:
- Instead of using macros, just create static methods within the `BES_Select` class. THen you could even use default function arguments instead of separate `_DEFAULT` macros.
- Maybe it would even be worth to create a separate class, e.g. `BoardEditorGraphicsViewContextMenu` which provides methods like `addDeleteAllAction()` so you don't even need to manually pass the `menu` object as parameter?

And regarding the execution of the corresponding actions: Did you already consider connecting the signal `QAction::triggered()` of the added actions directly to the corresponding functions to be executed? Maybe that would be a much cleaner solution than the huge if...else if...else if... construct after executing the menu...

And another thing: IMHO the `Point` class should really not have a dependency to `GridProperties`. Better pass the parameter `const PositiveLength& gridInterval` to `isOnGrid()`, which removes that dependency and is still very easy to use (see also `Point::mapToGrid()`).

5n8ke:
I'll take a look at it this evening. Personally I like macros for exactly this application, but if it doesn't fit the style, I'll use another way to incorporate it.

I kept the if... else... construct in place to keep the same style as used for the unique custom actions and to have the deciding logic available at a glance.

Having Point not depending on Gridproperties is understandable and quick to fix

5n8ke:
I worked a little bit more on the PR.
The Macros are now private functions and already connect the trigger signal to the desired logic. All the custom actions are also ported to use the trigger signal and the code lost some more redundancy due to this. This obviously also removed the if ... else ... construct.

5n8ke:
@ubruhin would this also be a place right now to implement the capabilities needed to fix #610 ?

Otherwise I see this PR as feature complete and would only do bug fixes.

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -309,259 +338,137 @@ BES_Base::ProcRetVal BES_Select::processIdleSceneRightMouseButtonReleased(
...
             footprintIcon, *footprint.getNames().value(
                                mProject.getSettings().getLocaleOrder()));
-        a->setData(footprint.getUuid().toStr());
         if (footprint.getUuid() ==
             devInst.getFootprint().getLibFootprint().getUuid()) {
           a->setCheckable(true);
           a->setChecked(true);
           a->setEnabled(false);
+        } else {
+          connect(a, &QAction::triggered, [this, board, &devInst, footprint](){
```
ubruhin, line 347: I think `footprint` can be passed by reference as well, to avoid copying each footprint object by value:

```suggestion
          connect(a, &QAction::triggered, [this, board, &devInst, &footprint](){
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -309,259 +338,137 @@ BES_Base::ProcRetVal BES_Select::processIdleSceneRightMouseButtonReleased(
...
+            try {
+              Uuid uuid = footprint.getUuid();
```
ubruhin, line 349: The variable `uuid` is actually not accessed, so it could be removed.

```suggestion
```

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -705,6 +574,79 @@ BES_Base::ProcRetVal BES_Select::processSubStateMovingSceneEvent(
...
+                                board->getGridProperties().getInterval()), false);
+        mUndoStack.execCmd(cmdMove.take());
+      } catch (Exception& e) {
+        QMessageBox::critical(&mEditor, tr("Error"), e.getMsg());
+      }
+    });
+  }
+}
+
+void BES_Select::addActionSelectAll(QMenu& menu, BI_NetSegment* netsegment,
```
ubruhin, line 642: Since `netsegment` is not allowed to be `nullptr`, please pass it as a reference to make callers clear that it must not be null.

```suggestion
void BES_Select::addActionSelectAll(QMenu& menu, BI_NetSegment& netsegment,
```

The same also applies to arguments of other `addAction*` methods.

libs/librepcb/projecteditor/boardeditor/fsm/bes_select.cpp
```diff
@@ -309,259 +338,137 @@ BES_Base::ProcRetVal BES_Select::processIdleSceneRightMouseButtonReleased(
...
+      BI_NetPoint* netpoint = dynamic_cast<BI_NetPoint*>(selectedItem);
+      Q_ASSERT(netpoint);
+
+      addActionDeleteAll(menu, &netpoint->getNetSegment());
+      menu.addSeparator();
+      addActionSelectAll(menu, &netpoint->getNetSegment());
+      menu.addSeparator();
+      addActionSnap(menu, netpoint->getPosition(), board, selectedItem);
+      menu.addSeparator();
+      addActionMeasure(menu, *netpoint->getNetLines().begin());
```
ubruhin, line 389: I agree it is unlikely, but at least theoretically it's possible that a netpoint has no netlines connected to it, so for safety reasons the netlines iterator should not be blindly dereferenced:

```suggestion
      if (!netpoint->getNetLines().isEmpty()) {
        menu.addSeparator();
        addActionMeasure(menu, *netpoint->getNetLines().begin());
      }
```

5n8ke:
@ubruhin Done!

ubruhin:
Thanks, after a quick review I would say it looks good now! :tada: 

Btw, since it seems you cleaned up the commit history (:+1:), just a small note: To be honest, I prefer to get rid of intermediate code states (in this case: the first change in `Point` which got changed again later, and the macros which were replaced later by functions). IMHO such intermediate states usually are not really helpful in future, but mainly lead to more "noise" in the history and (beside lower clearness) higher likelihood of conflicts when rebasing branches or cherry-picking commits onto the release branches. I'm really sorry about my nitpicking :see_no_evil: And no need to change it in this PR, since mainly only 1 module is affected, I'm fine with squash-merging it. Just wanted to share my thoughts about that topic ;)

5n8ke:
Thanks for the heads-up. I'll keep that in mind for future PRs

ubruhin (approved):

PR #751, opened 2020-08-07
Draw Trace: Allow connection to multiple anchors

### LINKED ISSUE #750: Board Editor: Placing a via while drawing traces only connects to one NetLine
```
LibrePCB Version: 0.2.0-unstable
Git Revision:     7c2c602e5efbd2d603f4e2cb610faff8411a6705
Build Date:       2020-08-07T10:41:46
Linking type:     static
Unbundled libs:   -
Qt Version:       5.15.0 (built against 5.15.0)
CPU Architecture: x86_64
Operating System: Windows 10 Version 1909
Platform Plugin:  windows
```

##### SUMMARY
When drawing a trace and a via will be placed at a position with NetLines on multiple layers, it is only connected to one of them.

##### EXPECTED RESULTS
The Via is connected to all NetLines of the same signal at the target position.

##### ACTUAL RESULTS
The Via is only connected to one NetLine at the target position.

5n8ke:
# Principle
1. Construct a List of all available anchors at the target position
2. Connect the new trace as previously for each anchor
3. Connect skipped NetPoints due to combining NetSegments

# Changes
- Fixes #750
- Fix error where finishing a trace at a Via or FootprintPad without NetLines was impossible

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/fsm/bes_drawtrace.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/fsm/bes_drawtrace.cpp
@@ -628,9 +628,6 @@ bool BES_DrawTrace::startPositioning(Board& board, const Point& pos,
 bool BES_DrawTrace::addNextNetPoint(Board& board) noexcept {
   Q_ASSERT(mSubState == SubState_PositioningNetPoint);
 
-  //TODO(5n8ke): Don't abort, when trying to add a via and selecting the start
-  // of the positioning
-
   // abort if no via should be added and p2 == p0 (no line drawn)
   if (!mTempVia && mTargetPos == mFixedStartAnchor->getPosition()) {
     abortPositioning(true);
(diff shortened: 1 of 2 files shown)


ubruhin:
Nice, but there seems to be at least one bug ;) Connecting to a netpoint leads to a LogicError and then leaves the tool in some intermediate state:

Peek 2020-08-23 23-15

The same situation works with the current `master`, so it must be an issue with these changes.

libs/librepcb/projecteditor/boardeditor/fsm/bes_drawtrace.cpp
```diff
@@ -651,38 +651,42 @@ bool BES_DrawTrace::addNextNetPoint(Board& board) noexcept {
...
+      if (netpoint == mPositioningNetPoint1 ||
+          netpoint == mPositioningNetPoint2) continue;
+      otherAnchors.append(netpoint);
+    }
+    foreach (BI_NetLine* netline, Toolbox::toSet(board.getNetLinesAtScenePos(
+                        mTargetPos, mAddVia ? nullptr : layer, netsignal))) {
+      if (netline == mPositioningNetLine1 ||
+          netline == mPositioningNetLine2) continue;
+      if (otherAnchors.contains(&netline->getStartPoint()) ||
+          otherAnchors.contains(&netline->getStartPoint())) continue;
```
5n8ke, line 684: The LogicError seems to stem from here since it's not checking for the endpoint

5n8ke (commented):
@ubruhin I think I found the bug.

5n8ke:
It is fixed. Also, the command is now aborted if an error occurs, to not be stuck in the intermediary state.

ubruhin:
Hmm yeah that issue seems to be fixed, but unfortunately now I noticed that there is even a more fundamental issue :sob: The thing is, when first determining a list of anchors, and then execute commands like `CmdCombineBoardNetSegments`, some anchors in the list could get invalid (e.g. replaced by newly created anchor objects) and thus must not be accessed anymore. It's hard to say what could happen if they are still accessed.

The statement `if (otherBase && !otherBase->isAddedToBoard()) continue;` looks like a workaround for this issue, but it's not working in every case :wink:

Example how to reproduce an issue (but it's not the only issue I was able to trigger. Depending on the exact situation, different errors occur):

Peek 2020-08-24 23-49

Wouldn't it be possible to immediately connect the found anchors one after each other, instead of memorizing a list of anchors to be connected?

5n8ke:
This seem to be two different problems.

The one in the video is when combining a netpoint with multiple anchors. Previously this case was only handled for vias.

The other is more deeply rooted in the system. I'll have to take a closer look if directly combining the anchors is possible.

5n8ke:
The first problem is fixed. Perhaps we want to extract combineAnchors() into its own undo command.

For the other problem, I don't have a solution yet. It is possible to create a set of anchors that contains only one for each disjunct segment. And afterward, all points at the target position must be of the same segment.
@ubruhin But I was wondering why checking isAddedToBoard() is not enough.

ubruhin:
> The first problem is fixed.

Great, now it seems to work fine! :+1: At least I found no issue during testing :)

> Perhaps we want to extract combineAnchors() into its own undo command.

Might be reasonable, but I'm happy with the current implementation.

> For the other problem, I don't have a solution yet. It is possible to create a set of anchors that contains only one for each disjunct segment. And afterward, all points at the target position must be of the same segment.

OK, let's keep it that way for the moment since it seems to be working fine.

> But I was wondering why checking isAddedToBoard() is not enough.

Hmm to be honest, I'm not sure anymore if I had critical issues in mind when using `isAddedToBoard()` :thinking: But I called it "workaround" since an anchor which is no longer added to the board might actually still exist, but is represented by a different object (e.g. `CmdCombineBoardNetSegments` copies all anchors of one netsegment into the other netsegment and removes their original). By ignoring them, they will not be combined of course. But maybe it actually doesn't lead to critical issues, but just ignores such anchors when combining anchors. In that case, it's fine for me.

After all I'd say this PR is a clear improvement and seem to be working stable, so I'm happy to merge it :) From your perspective this is ready to merge too, right?

5n8ke:
Yes, this is ready.
The problem I see might be with the life cycle of objects that are removed from boards. As long as the memory is not yet freed, isAddedToBoard() should always be a reasonable check. But I'm not familiar with the code enough to know when the memory is actually freed.

ubruhin:
> Yes, this is ready.

OK :+1: Thanks again for the contribution! :)

> The problem I see might be with the life cycle of objects that are removed from boards. As long as the memory is not yet freed, isAddedToBoard() should always be a reasonable check. But I'm not familiar with the code enough to know when the memory is actually freed.

An object which was added at least once to a board should not be freed until the project is closed, since it might be needed again when the user clicks "undo". So this should be safe.

5n8ke:
> An object which was added at least once to a board should not be freed until the project is closed, since it might be needed again when the user clicks "undo". So this should be safe.

Good to know. Then the check should definitely solve the problem.

PR #753, opened 2020-08-11
Board editor: Snap to nearest anchor when starting a trace

### LINKED ISSUE #631: Board Editor: Prevent error "Nothing to connect."

5n8ke:
When no anchor is found at the start location of drawing a trace, look in a radius of 1cm for other possible candidates.
Currently supports
- NetPoints
- Vias
- Pads

# Principle
The search function looks for anchors in the radius of maxDistance around the specified position on a given layer. When a candidate is found, the radius is reduced to the distance between the position and the center point of the anchor.
The function searches for NetPoints, Vias, and Pads in the given order. It returns the last found anchor since it laid within the smallest search radius.

## Caveat
Currently (similar to the rendering of airwires) NetLines are not considered, even if a new NetPoint would result in a closer anchor.

# Linked items
- Merge #632 with current master
- Fixes #631

### DIFF
--- a/libs/librepcb/project/boards/board.cpp
+++ b/libs/librepcb/project/boards/board.cpp
@@ -540,6 +540,60 @@ QList<BI_FootprintPad*> Board::getPadsAtScenePos(
   return list;
 }
 
+BI_NetPoint* Board::getNetPointNextToScenePos(
+    const Point& pos, UnsignedLength& maxDistance, const GraphicsLayer* layer,
+    const NetSignal* netsignal) const {
+  BI_NetPoint* bestMatch = nullptr;
+  foreach (BI_NetSegment* segment, mNetSegments) {
+    if ((!netsignal) || (&segment->getNetSignal() == netsignal)) {
+      BI_NetPoint* newMatch =
+          segment->getNetPointNextToScenePos(pos, layer, maxDistance);
+      if (newMatch) bestMatch = newMatch;
+    }
+  }
+  return bestMatch;
+}
+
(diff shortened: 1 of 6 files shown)


doom-fr:
I think that the distance used should depend of the zoom factor

5n8ke:
Since it is only the maximum distance of the search radius but in any way every anchor will be checked, it would even be feasible to not specify it at all.

In any case, what would be a good metric? 1cm in screen space, 5cm?

It comes down to the question if the error "nothing up connect" should even be shown at all when a viable candidate is present anywhere on the board.

ubruhin:
Thanks! Seems to work fine :+1: However, the order of the new methods is not ideal - sometimes they are placed *in between* the already existing `get*AtScenePos()` methods, and the order within *.h files is not consistent with the order in the *.cpp files. Could you clean up these inconsistencies?

5n8ke:
I'll fix that tonight.

5n8ke:
@ubruhin It's fixed and ready to be merged

ubruhin:
Hmm, `board.cpp` is still not consistent with `board.h`, but OK, I can turn a blind eye :see_no_evil: Just waiting for CI...

5n8ke:
@ubruhin Fixed board.cpp for you

PR #756, opened 2020-08-24
Improve search behavior in board- and schematic editor

5n8ke:
- Allow cycling through results beginning with the current search term
- Zoom depending on the symbol/device size, and showing the maximum dimension as 1/4 of the screen

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
@@ -104,7 +104,7 @@ BoardEditor::BoardEditor(ProjectEditor& projectEditor, Project& project)
   if (!mProject.getDirectory().isWritable()) {
     filenameStr.append(QStringLiteral(" [Read-Only]"));
   }
-  setWindowTitle(QString("%1 - LibrePCB Board Editor").arg(filenameStr));
+  setWindowTitle(tr("%1 - LibrePCB Board Editor").arg(filenameStr));
 
   // Add Dock Widgets
   mUnplacedComponentsDock = new UnplacedComponentsDock(mProjectEditor);
@@ -443,7 +443,7 @@ void BoardEditor::on_actionCopyBoard_triggered() {
   bool    ok   = false;
   QString name = QInputDialog::getText(
       this, tr("Copy Board"), tr("Choose a name:"), QLineEdit::Normal,
-      QString(tr("copy_of_%1")).arg(*board->getName()), &ok);
+      tr("copy_of_%1").arg(*board->getName()), &ok);
   if (!ok) return;
 
   try {
@@ -462,8 +462,7 @@ void BoardEditor::on_actionRemoveBoard_triggered() {
 
   QMessageBox::StandardButton btn = QMessageBox::question(
       this, tr("Remove board"),
-      QString(tr("Are you really sure to remove the board \"%1\"?"))
-          .arg(*board->getName()));
+      tr("Are you really sure to remove the board \"%1\"?").arg(*board->getName()));
   if (btn != QMessageBox::Yes) return;
 
   try {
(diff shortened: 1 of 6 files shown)


libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
```diff
@@ -88,6 +88,8 @@ BoardEditor::BoardEditor(ProjectEditor& projectEditor, Project& project)
     mProject(project),
     mUi(new Ui::BoardEditor),
     mGraphicsView(nullptr),
+    mLastSearch(""),
+    mNextDeviceIndex(1),
```
ubruhin, line 92: Although probably it doesn't really matter, wouldn't an initialization value `0` be more intuitive? ;)

libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
```diff
@@ -861,17 +862,36 @@ QStringList BoardEditor::getSearchToolBarCompleterList() noexcept {
...
+      if (device->getComponentInstance().getName()
+          ->startsWith(name, Qt::CaseInsensitive)) {
+        deviceCandidates.append(device);
+      }
+    }
+    std::sort(deviceCandidates.begin(), deviceCandidates.end(),
+              [](BI_Device* A, BI_Device* B){
+      return A->getComponentInstance().getName()
+          < B->getComponentInstance().getName();
+    });
```
ubruhin, line 880: Just to comply with naming conventions:

```suggestion
    std::sort(deviceCandidates.begin(), deviceCandidates.end(),
              [](BI_Device* a, BI_Device* b){
      return a->getComponentInstance().getName()
          < b->getComponentInstance().getName();
    });
```

libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
```diff
@@ -861,17 +862,36 @@ QStringList BoardEditor::getSearchToolBarCompleterList() noexcept {
...
+    if (deviceCandidates.count()) {
+      if (mNextDeviceIndex >= deviceCandidates.count()) {
+        mNextDeviceIndex = 0;
       }
+      BI_Device* device = deviceCandidates[mNextDeviceIndex++];
+      mNextDeviceIndex = mNextDeviceIndex % deviceCandidates.count();
```
ubruhin, line 886: Actually this line isn't needed, since the wrap around will be done anyway during the *next* call, or am I wrong?

```suggestion
```

libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
```diff
@@ -861,17 +862,36 @@ QStringList BoardEditor::getSearchToolBarCompleterList() noexcept {
...
+      board->clearSelection();
+      device->setSelected(true);
+      QRectF rect = device->getFootprint().getBoundingRect();
+      //Zoom to a rectangle relative to the maximum device dimension. The device
+      // is 1/4th of the screen
```
ubruhin, line 891: Missing space :see_no_evil: 

```suggestion
      // Zoom to a rectangle relative to the maximum device dimension. The
      // device is 1/4th of the screen.
```

ubruhin (commented):
Thanks! :smiley: 

Functionality is good IMO. I just added some comments about the code. In addition, actually I think it would be great to avoid code duplication by moving this functionality into the `SearchToolBar` class. That class could memorize the last search term, and pass the incrementing index as an additional argument to the `goToTriggered()` signal, e.g. like that:

```cpp
signals:
  void goToTriggered(const QString& name, int index);
```

Of course that class cannot know how many elements are in the loop, so it cannot do the wrap around. But the editors could easily do that with the modulo operator.

Or what do you think about that?

EDIT: I only commented to the changes in the board editor. Of course the same comments also apply to the schematic editor :)

libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
5n8ke: It should probably be 0. I don't know anymore why I chose 1

libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
5n8ke: Always the lower case variables... :D

5n8ke:
That would probably be a good course of action. This would also allow a better reset of the counter with every edit of the search field.

Then I'll refactor these changes further.

5n8ke:
I implemented the requested changes.
The index is now calculated in the SearchToolbar itself and resets when the text is edited.
I also created the function getSearchCandidate in the editors, to supply both the completion list and the goTo* function with identical lists of candidates.

libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
```diff
@@ -849,31 +848,52 @@ void BoardEditor::clearDrcMarker() noexcept {
...
-QStringList BoardEditor::getSearchToolBarCompleterList() noexcept {
-  QStringList list;
+QList<BI_Device*> BoardEditor::getSearchCandidates() noexcept {
+  QList<BI_Device*> candidates = {};
   if (Board* board = getActiveBoard()) {
-    foreach (const BI_Device* device, board->getDeviceInstances()) {
-      list.append(*device->getComponentInstance().getName());
+    foreach (BI_Device* device, board->getDeviceInstances()) {
+      candidates.append(device);
     }
```
ubruhin, line 856: Could be simplified:

```suggestion
    candidates += board->getDeviceInstances().values();
```

libs/librepcb/projecteditor/schematiceditor/schematiceditor.cpp
```diff
@@ -696,31 +696,51 @@ void SchematicEditor::renameSchematic(int index) noexcept {
...
-  QStringList list;
+QList<SI_Symbol*> SchematicEditor::getSearchCandidates() noexcept {
+  QList<SI_Symbol*> candidates = {};
   foreach (const Schematic* schematic, mProject.getSchematics()) {
-    foreach (const SI_Symbol* symbol, schematic->getSymbols()) {
-      list.append(symbol->getName());
+    Q_ASSERT(schematic);
+    foreach (SI_Symbol* symbol, schematic->getSymbols()) {
+      candidates.append(symbol);
     }
```
ubruhin, line 705: Could be simplified:

```suggestion
    candidates += schematic->getSymbols();
```

libs/librepcb/projecteditor/schematiceditor/schematiceditor.cpp
```diff
@@ -696,31 +696,51 @@ void SchematicEditor::renameSchematic(int index) noexcept {
...
+      symbolCandidates.append(symbol);
+    }
+  }
+
+  if (symbolCandidates.count()) {
+    index %= symbolCandidates.count();
+    SI_Symbol* symbol = symbolCandidates[index];
+    Schematic* schematic = &symbol->getSchematic();
+    if (setActiveSchematicIndex(mProject.getSchematics().indexOf(schematic))) {
+      schematic->clearSelection();
```
ubruhin, line 736: As mentioned in a previous PR, if you get a reference, better don't convert it to a pointer if not really needed:

```suggestion
    Schematic& schematic = symbol->getSchematic();
    if (setActiveSchematicIndex(mProject.getSchematics().indexOf(&schematic))) {
      schematic.clearSelection();
```

Of course it's not critical, but IMHO it's good habit to let references stay references :)

ubruhin (commented):
Nice, I like the refactorings! :+1:

I only found a few minor code improvement opportunities, see comments. If you agree and apply them, I'll merge this PR :)

> Should we create a fuzzy and an exact search signal? Where the second is only emitted when it matches one of the completion terms.

Hmm I wouldn't do that as long as that signal is not used somewhere. But of course that might be useful some day.

5n8ke:
@ubruhin All reviews incorporated and ready to be merged.

ubruhin (approved):
Perfect, thanks! :tada:

libs/librepcb/projecteditor/schematiceditor/schematiceditor.cpp
```diff
@@ -696,31 +696,49 @@ void SchematicEditor::renameSchematic(int index) noexcept {
   }
 }
 
+QList<SI_Symbol*> SchematicEditor::getSearchCandidates() noexcept {
+  QList<SI_Symbol*> candidates = {};
+  foreach (const Schematic* schematic, mProject.getSchematics()) {
+    Q_ASSERT(schematic);
+    camdidates += schematic->getSymbols();
```
ubruhin, line 703: Oooh damn, now there is a typo which makes CI fail :sob:

```suggestion
    candidates += schematic->getSymbols();
```

5n8ke:
I wonder why it compiled then in my local version... Something else to investigate.

PR #757, opened 2020-08-24
Editors: Improve context menu handling

### LINKED ISSUE #536: Selection of stacked items

5n8ke:
Improve context menu handling for the symbol, package, and schematic editor.

- Implement improvements as done in #748
- Implement findItemsAtPosition function where necessary
- Allow cycling through items while holding the shift key
- Fixes #536

### DIFF
--- a/libs/librepcb/libraryeditor/pkg/fsm/packageeditorstate_select.cpp
+++ b/libs/librepcb/libraryeditor/pkg/fsm/packageeditorstate_select.cpp
@@ -58,7 +58,10 @@ namespace editor {
  ******************************************************************************/
 
 PackageEditorState_Select::PackageEditorState_Select(Context& context) noexcept
-  : PackageEditorState(context), mState(SubState::IDLE), mStartPos() {
+  : PackageEditorState(context),
+    mState(SubState::IDLE),
+    mStartPos(),
+    mCurrentSelectionIndex(0) {
 }
 
 PackageEditorState_Select::~PackageEditorState_Select() noexcept {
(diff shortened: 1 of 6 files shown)


libs/librepcb/libraryeditor/pkg/fsm/packageeditorstate_select.cpp
```diff
@@ -369,103 +366,125 @@ bool PackageEditorState_Select::processAbortCommand() noexcept {
...
-  } else if (action == aFlipH) {
-    return mirrorSelectedItems(Qt::Horizontal, true);
-  } else if (action == aRemove) {
-    return removeSelectedItems();
-  } else if (action == aProperties) {
-    return openPropertiesDialogOfItemAtPos(pos);
-  } else {
-    return false;
-  }
+  return menu.exec(QCursor::pos());
```
ubruhin, line 424: If the menu was opened, the event should be considered as handled, no matter whether the user triggered one of the menu items or not. So we should always return true here.

```suggestion
  menu.exec(QCursor::pos());
  return true;
```

libs/librepcb/libraryeditor/sym/fsm/symboleditorstate_select.cpp
```diff
@@ -347,86 +347,108 @@ bool SymbolEditorState_Select::processAbortCommand() noexcept {
...
-  } else if (action == aMirrorH) {
-    return mirrorSelectedItems(Qt::Horizontal);
-  } else if (action == aRemove) {
-    return removeSelectedItems();
-  } else if (action == aProperties) {
-    return openPropertiesDialogOfItemAtPos(pos);
-  } else {
-    return false;
-  }
+  return menu.exec(QCursor::pos());
```
ubruhin, line 400: Same here:

```suggestion
  menu.exec(QCursor::pos());
  return true;
```

libs/librepcb/projecteditor/schematiceditor/fsm/schematiceditorstate_select.cpp
```diff
@@ -319,85 +311,60 @@ bool SchematicEditorState_Select::processGraphicsSceneRightMouseButtonReleased(
...
+      break;
     }
-  }
 
-  return false;
+    default:
+      return false;
+  }
+  // execute the context menu
+  return menu.exec(e.screenPos());
```
ubruhin, line 367: And one more time:

```suggestion
  menu.exec(e.screenPos());
  return true;
```

ubruhin (commented):
Looks and works very good! :smiley: 

So far I have only very few code suggestions, see comments :)

libs/librepcb/libraryeditor/pkg/fsm/packageeditorstate_select.cpp
5n8ke: As I see it, this would change the current behavior, which itself is inconsistent compared to the schematic editor.

The else statement returns false and represents the case when no action was selected.

ubruhin: Hmm you're right, in the library editor it indeed would change the behavior :thinking: So I would say that was actually wrong until now :see_no_evil: 

But both the schematic editor and board editor currently always return true if the context menu was shown, no matter whether an action was clicked or not. False is only returned if the menu was not even shown. That's the correct behavior I would say.

5n8ke: Ok, then I'll change it tomorrow and check for any other place where the wrong behavior might be present.

5n8ke:
It's done!

ubruhin (approved):
Thanks! :+1:

PR #758, opened 2020-08-30
Format & clean up source files

ubruhin:
## Make `./dev/format_code.sh` running even without tty

The script did not work when running from a recent QtCreator because it had no tty attached. Now it runs both, with and without tty.

## Remove unnecessary QString() wrappers around tr()

Just to clean up the code.

## Format all source files with `./dev/format_code.sh --all`

To fix all style inconsistencies.


dbrgn:
Reminds me of #566 :wink: The longer we wait, the more annoying it will be.

ubruhin:
> Reminds me of #566 wink The longer we wait, the more annoying it will be.

True :) If I remember correctly, we agreed on changing the alignment rules since everyone agrees. Afterwards we could still change more rules which will then be easier to review if alignment was already reformatted.

PR #759, opened 2020-08-30
Implement reusable via/pad shape selector

5n8ke:
As part of the ui redesign I created a two reusable widgets for choosing the shape of pads and vias.
The existing selector for pads was updated to create the buttons more generally and to follow the same style.

Both widgets also include the size selection.


5n8ke:
Seems QButtonGroup::idClicked() was only added in QT 5.15 I'll change it later tonight

ubruhin:
To be honest, I don't like to support both shape types with one widget - IMO it would be much cleaner, more flexible and easier to maintain if two separate widgets are created as long as the shape enums are separate types anyway. And (currently) we can't share the enum since they are not identical ("square" vs. "rect").

Beside that, the change itself introduces an architectural violation - `bi_via.h` must not be included from within `libs/librepcb/common/` (see [architecture overview](https://developers.librepcb.org/de/d34/doc_architecture_overview.html)).

> which oldest Qt version are we still supporting? Is it Qt 5.2 as I just found in the docs?

Currently 5.2, yes. The plan is to support the Qt version which is available in the oldest still supported Ubuntu or Debian release. Actually I'm not sure which Qt version that would be at the moment, so far it was very easy to keep support for Qt5.2, so I had no motivation to change it ;)

5n8ke:
Ok, then I'll split the code again and move it into the respective locations.
This would also allow the edits for size, drill,... To be included in the widget.
I'm sorry about the architectural violation. There is still a lot to learn for me about the code base.

I looked at the Qt history and 5.2 is from 2011 or so. But since I'm aware of that, it's easy to add the necessary switches for newer functions.

5n8ke:
@ubruhin The shape selectors for pads and vias are separated again.
I also included the width and height edit for pads and size and drill edit for vias.
I excluded the drill size for pads for now, since it is not used for SMD pads.
I also changed some GUI callbacks to lambdas when they were short and only used a single time.

5n8ke:
When the functionality is acceptable as is, then I see this as ready to merge as soon as I incorporate the refactor tomorrow.

5n8ke:
@ubruhin It's ready to merge.

ubruhin:
Sorry for not responding earlier, I couldn't find the time to review it properly until now :(

Unfortunately I found several issues:
- A toolbar widget should not be derived from `QToolBar` since it's not a toolbar, but only a widget which gets added to a toolbar. Such widgets should be derived from `QWidget` (or alternatively from `QWidgetAction`, but that's more complicated). 
- Somehow (at least on my PC) the toolbar with the new widgets are slightly higher that normal toolbars. This causes the UI to "jitter" around when switching between the different board editor tools. Maybe this is caused by deriving from `QToolBar`(?).
- I'm not sure if it actually makes sense to create one widget to control multiple, independent properties of a pad or via. Separate widgets for each property allows us to use them at more places, for example a shape selector widget could also be used in `FootprintPadPropertiesDialog` - but one widget representing multiple properties wouldn't be useful, since it doesn't fit into a form layout of a dialog.
  - Related issue: within the `PadShapeSelector` class, the key for storing the displayed unit is hardcoded to "package_editor/add_pads/{width,height}", so this class should only be used within the "Add Pads" tool, i.e. is not really reusable. Having separate widgets solves this problem "automatically" since the corresponding tools have to configure these keys on their own ;)

I totally agree it makes sense to create a new widget for the via shape selector to simplify the board editor tools "Add Via" and "Draw Trace", since these tools currently build the selector widgets on their own. But to be honest, I don't fully understand the motivation to combine multiple editor widgets into one widget. Do you want to further simplify the board editor tools? If yes, maybe it would make more sense to use the `ToolBarProxy` class which would take ownership over the toolbar widgets and provides a `clear()` method to destroy all toolbar items at once, so we could replace all widget pointer member variables by one `ToolBarProxy` object, and remove all the cleanup code within the `exit()` functions. What do you think about that?

5n8ke:
Thank you for the review. I had some time to think about your comments and you're correct.

- I didn't think about using the widget in a form, since at least for the vias, this is currently not implemented. I was thinking more about just using it in a toolbar, e.g when a via is selected, showing the respective widget to edit the item.
- In my opinion, the hardcoded key does not pose a problem. It should be renamed, that's true, but having the chosen selection carry across _all_ places, where the item is edited seems like a usability improvement for me.
- I chose to inherit from QToolBar to gain access to the convenience functions for adding actions, which then match in style to the other actions that are not added via the widget. This also gets rid of the usage of QToolButtons.

I'll take a look at the ToolBarProxy class and its use for the schematic and board editor.

ubruhin:
Since this PR is pretty outdated and somehow even obsolete since the shape of vias is no longer configurable, I'll close it. I'm sorry for the work not being merged. In the future I recommend to discuss any change first before implementing it as this allows to create a common understanding and agreement of the changes. Thanks!

PR #760, opened 2020-08-30
BoardEditor: Refactor whole state machine

ubruhin:
Migrating to the simpler, cleaner FSM design pattern as already used in the library editor and schematic editor. The reduced complexity should make it easier to implement new board editor features. In addition, the naming of the classes is now more expressive.

The behavior of the board editor should not be affected by this change, except a few tiny fixes.

Same as #720, but this time for the board editor.


5n8ke:
Works fine for meI found no errors/problematic differences to the previous behavior so far.

PR #763, opened 2020-08-31
Remove linking output message in qmake project file

ubruhin:
The message() causes the QtCreator (at least sometimes) to switch to a different output pane, which is quite annoying. And in colored console output it might be confusing too since it it the only output on stderr. Thus I'd say let's remove it.

@dbrgn is that fine for you? :slightly_smiling_face:


dbrgn:
Sure, this was mostly for debug purposes. If it has any downsides, remove it.

PR #764, opened 2020-08-31
BES_DrawTrace: Get rid of member variable mCurrentNetSignal

### LINKED ISSUE #762: Board editor: Assert is raised when starting trace at unconnected pad

```
LibrePCB Version: 0.2.0-unstable
Git Revision:     2040eac1afbc0cf1f135884d9c05bf62b156a21c
Build Date:       2020-08-31T20:59:09
Linking type:     static
Unbundled libs:   -
Qt Version:       5.5.1 (built against 5.5.1)
CPU Architecture: x86_64
Operating System: Ubuntu 16.04.7 LTS
Platform Plugin:  xcb
```

##### SUMMARY

When starting to draw a trace at a footprint pad which has no netsignal, an assert is raised (resp. crashes in release mode).

##### STEPS TO REPRODUCE

- Add a device to a board containing pads which are not connected
- Start the "draw trace" tool
- Click on the unconnected pad

##### EXPECTED RESULTS

The trace should start at the pad.

##### ACTUAL RESULTS

The following assert is raised:

https://github.com/LibrePCB/LibrePCB/blob/1f2c0d411a165ae42772b49549dbef78a3c91794/libs/librepcb/projecteditor/boardeditor/fsm/bes_drawtrace.cpp#L566

/cc @5n8ke

5n8ke:
Fix crash when trying to connect to a pad that is not part of a signal.

Fixes #762

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/fsm/bes_drawtrace.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/fsm/bes_drawtrace.cpp
@@ -75,7 +75,6 @@ BES_DrawTrace::BES_DrawTrace(BoardEditor& editor, Ui::BoardEditor& editorUi,
     mCurrentSnapActive(true),
     mFixedStartAnchor(nullptr),
     mCurrentNetSegment(nullptr),
-    mCurrentNetSignal(nullptr),
     mPositioningNetLine1(nullptr),
     mPositioningNetPoint1(nullptr),
     mPositioningNetLine2(nullptr),
@@ -514,8 +513,11 @@ bool BES_DrawTrace::startPositioning(Board& board, const Point& pos,
     layer->setVisible(true);
 
     // determine the fixed anchor (create one if it doesn't exist already)
-    mCurrentNetSignal  = nullptr;
-    mCurrentNetSegment = nullptr;
+    // if the selected item is not part of a NetSegment (e.g. device pads),
+    // netsignal must be set to a valid NetSignal, which is used to create the
+    // new NetSegment
+    NetSignal* netsignal = nullptr;
+    mCurrentNetSegment   = nullptr;
     if (fixedPoint) {
       mFixedStartAnchor  = fixedPoint;
       mCurrentNetSegment = &fixedPoint->getNetSegment();
@@ -534,7 +536,11 @@ bool BES_DrawTrace::startPositioning(Board& board, const Point& pos,
     } else if (BI_FootprintPad* pad = findPad(board, pos)) {
       mFixedStartAnchor  = pad;
       mCurrentNetSegment = pad->getNetSegmentOfLines();
-      mCurrentNetSignal  = pad->getCompSigInstNetSignal();
+      netsignal          = pad->getCompSigInstNetSignal();
+      if (!netsignal) {
+        throw Exception(__FILE__, __LINE__,
+                        tr("Pad is not connected to any signal."));
+      }
       if (pad->getLibPad().getBoardSide() !=
           library::FootprintPad::BoardSide::THT) {
         layer = board.getLayerStack().getLayer(pad->getLayerName());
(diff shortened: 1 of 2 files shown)


ubruhin:
Btw, #753 will introduce the same issue again just a few lines below. Maybe better move the exception to outside the if..else if..else if construct to make sure *any* case of a missing netsignal is catched?

5n8ke:
Actually the assert could be replaced with the exception.

But I want to take a closer look at mCurrentNetSignal, since it's only set when accessing a device pad and not for the other items.

5n8ke:
mCurrentNetSignal is now removed since the correct netsignal can always be gotten from mCurrentNetSegment.
I introduced the same check in #753 where it definitely caused crashes and would update this PR as soon as the other one is merged.

5n8ke:
Since #753 is merged, I'll update this PR tonight.

ubruhin:
> Since #753 is merged, I'll update this PR tonight.

Perfect! Afterwards I'll rebase and merge #760 if there are no objections.

5n8ke:
Not on my side.

5n8ke:
@ubruhin It's updated to the current master version

libs/librepcb/projecteditor/boardeditor/fsm/bes_drawtrace.cpp
```diff
@@ -918,23 +916,27 @@ void BES_DrawTrace::updateNetpointPositions() noexcept {
     // find anchor under cursor
     GraphicsLayer* layer = mPositioningNetPoint1->getLayerOfLines();
     Q_ASSERT(layer);
-    if (BI_Via* via =
-            findVia(board, mCursorPos, mCurrentNetSignal, {mTempVia})) {
+    NetSignal* netsignal = &mCurrentNetSegment->getNetSignal();
+    // NOTE(5n8ke): netsignal must not be nullptr, since a connection should
+    // only be made to the current NetSignal
+    Q_ASSERT(netsignal);
```
ubruhin, line 922: `netsignal` can't be `nullptr` since it was constructed by taking the address of an object, so no check is needed:

```suggestion
    NetSignal* netsignal = &mCurrentNetSegment->getNetSignal();
```

But actually, one more time, simply don't convert the reference to a pointer here, then it's obvious that it can't be `nullptr` :smiley: :

```suggestion
    NetSignal& netsignal = mCurrentNetSegment->getNetSignal();
```

Then use `&netsignal` where you really need a pointer.

libs/librepcb/projecteditor/boardeditor/fsm/bes_drawtrace.cpp
```diff
@@ -976,11 +978,14 @@ void BES_DrawTrace::layerComboBoxIndexChanged(int index) noexcept {
...
-    Point            startPos = mFixedStartAnchor->getPosition();
-    BI_FootprintPad* padAtStart =
-        findPad(board, startPos, nullptr, mCurrentNetSignal);
-    if (findVia(board, startPos, mCurrentNetSignal) ||
+    Board&     board     = mPositioningNetPoint1->getBoard();
+    Point      startPos  = mFixedStartAnchor->getPosition();
+    NetSignal* netsignal = &mCurrentNetSegment->getNetSignal();
+    // NOTE(5n8ke): netsignal must not be nullptr, since a connection should
+    // only be made to the current NetSignal
+    Q_ASSERT(netsignal);
```
ubruhin, line 986: Same here:

```suggestion
    NetSignal& netsignal = mCurrentNetSegment->getNetSignal();
```

ubruhin (commented):
Although it turned out that the "fix" is actually not needed (sorry...), I always like removing members since usually it reduces complexity and error-proneness :smiley:

libs/librepcb/projecteditor/boardeditor/fsm/bes_drawtrace.cpp
5n8ke: I chose to use a pointer here since all the following find* take a pointer as the netsignal parameter.
The check is not needed, but I added the asserts as a sanity check to make it explicit, since a null pointer was previously possible with mCurrentNetSignal.

ubruhin: Hmm okay - to be honest, I don't like it (IMO a reference would be much cleaner) but this time I'll turn a blind eye :wink:

PR #765, opened 2020-09-01
Library Editor: Duplicate all categories

### LINKED ISSUE #696: When duplicating symbols, only one category is copied

```
LibrePCB Version: 0.1.4
Git Revision:     ae04eef5a71b5ba66ae2cee6b631c1c933ace535
Build Date:       2020-05-02T18:50:04
CPU Architecture: x86_64
Operating System: Arch Linux
Platform Plugin:  xcb
```

##### SUMMARY

When duplicating a symbol with multiple categories, only one of them is copied.

##### STEPS TO REPRODUCE

1. Create symbol with a category
2. Add an additional category
3. Duplicate symbol by right clicking

(Note: Not sure if it's related, but it should be possible to assign multiple categories in the creation wizard. Right now you can choose only one.)

5n8ke:
Copy all categories to when duplicating a symbol/device/... in a library.

- NewElementWizardContext holds a list of all categories of the original item
- duplicate/new element wizard
  - add "and x more" label to signify more "hidden" categories. 
  - reset clears the complete list
  - change replaces *all* categories with the newly selected one
  - the wizard cannot assign multiple categories. This has to be done as usual in the editor
- fixes #696

### DIFF
--- a/libs/librepcb/libraryeditor/newelementwizard/newelementwizardcontext.h
+++ b/libs/librepcb/libraryeditor/newelementwizard/newelementwizardcontext.h
@@ -129,7 +129,7 @@ public:  // Data
   QString                   mElementKeywords;
   QString                   mElementAuthor;
   tl::optional<Version>     mElementVersion;
-  tl::optional<Uuid>        mElementCategoryUuid;
+  QSet<Uuid>                mElementCategoryUuids;
 
   // symbol
   SymbolPinList mSymbolPins;
--- a/libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.cpp
+++ b/libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.cpp
@@ -152,16 +152,24 @@ void NewElementWizardPage_EnterMetadata::btnChooseCategoryClicked() noexcept {
       return;
     }
   }
-  mContext.mElementCategoryUuid = categoryUuid;
+  mContext.mElementCategoryUuids.clear();
+  if (categoryUuid.has_value()) {
+    mContext.mElementCategoryUuids.insert(*categoryUuid);
+  }
   updateCategoryTreeLabel();
 }
 
 void NewElementWizardPage_EnterMetadata::btnResetCategoryClicked() noexcept {
-  mContext.mElementCategoryUuid = tl::nullopt;
+  mContext.mElementCategoryUuids.clear();
   updateCategoryTreeLabel();
 }
 
 void NewElementWizardPage_EnterMetadata::updateCategoryTreeLabel() noexcept {
+  tl::optional<Uuid> rootCategoryUuid = tl::nullopt;
+  if (mContext.mElementCategoryUuids.count()) {
+    rootCategoryUuid = mContext.mElementCategoryUuids.values().first();
+  }
+
   switch (mContext.mElementType) {
     case NewElementWizardContext::ElementType::ComponentCategory:
     case NewElementWizardContext::ElementType::Symbol:
@@ -173,7 +181,7 @@ void NewElementWizardPage_EnterMetadata::updateCategoryTreeLabel() noexcept {
           *mUi->lblCategoryTree);
       builder.setHighlightLastLine(true);
       builder.setOneLine(true);
-      builder.updateText(mContext.mElementCategoryUuid);
+      builder.updateText(rootCategoryUuid);
       break;
     }
     case NewElementWizardContext::ElementType::PackageCategory:
@@ -184,13 +192,19 @@ void NewElementWizardPage_EnterMetadata::updateCategoryTreeLabel() noexcept {
           *mUi->lblCategoryTree);
       builder.setHighlightLastLine(true);
       builder.setOneLine(true);
-      builder.updateText(mContext.mElementCategoryUuid);
+      builder.updateText(rootCategoryUuid);
       break;
     }
     default: {
       mUi->lblCategoryTree->setText(tr("Root category"));
       break;
     }
+  }
+  if (mContext.mElementCategoryUuids.count() > 1) {
+    mUi->lblMoreCategories->setText(
+        tr("... and %1 more.").arg(mContext.mElementCategoryUuids.count() - 1));
+  } else {
+    mUi->lblMoreCategories->setText(QString());
   }
 }
 
--- a/libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.ui
+++ b/libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.ui
@@ -122,6 +122,22 @@
       </widget>
      </item>
      <item>
+      <widget class="QLabel" name="lblMoreCategories">
+       <property name="sizePolicy">
+        <sizepolicy hsizetype="Maximum" vsizetype="Preferred">
+         <horstretch>0</horstretch>
+         <verstretch>0</verstretch>
+        </sizepolicy>
+       </property>
+       <property name="text">
+        <string notr="true">... and X more.</string>
+       </property>
+       <property name="alignment">
+        <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
+       </property>
+      </widget>
+     </item>
+     <item>
       <widget class="QToolButton" name="btnChooseCategory">
        <property name="minimumSize">
         <size>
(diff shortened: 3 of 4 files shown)


5n8ke:
This PR changes the behavior of copying a symbol/component/... by copying all categories and not just the first one.
@ubruhin This raises the question of what should happen, when the user changes or resets the category. Should all be changed or reset? The current implementation is, that only the displayed category is changed or reset, leading to an error with the functional test.

Or should I implement a new category tree that allows the display and addition/deletion of more than one category during the copy dialog.

ubruhin:
I suspect it would be quite difficult to properly handle multiple categories in the UI without making it cumbersome for users. So currently my opinion is to simply let the "choose category" button simply clear all categories and add only the one which was selected. It's not perfect since in the UI it's not visible when multiple categories are copied. But at least it covers the typical use case of copying a library element with all categories (if there is no need to click on "choose category without making the UI more complicated.

dbrgn:
> Or should I implement a new category tree that allows the display and addition/deletion of more than one category during the copy dialog.

I would prefer that, if possible in a nice way. I find it a bit strange that only one category can be assigned when creating the element, but that more can be added later on...

ubruhin:
@dbrgn Most wizards provide only a small subset of the actually available features, that's the purpose of a wizard - provide a simple, intuitive way to fill in complex data. When providing all available options in the wizard, it will get cluttered up and gets more cumbersome to use. If you want to have all options from the beginning on, IMO we should rather provide an option to create a new element without the wizard at all, which simply creates a completely empty library element...

Of course I'd also be happy with allowing to edit multiple categories, but I have no idea how this could look like without decreasing usability (e.g. a new wizard page would be an option to provide the whole category edit widget, but IMHO that would be too annoying for users).

5n8ke:
I thought about the wizard clutter today a bit more.

One possible solution might be to change the category display, to be condensed into a single line instead of a new one for each category layer. This would allow to display them in a list and add two simple add and delete buttons for managing the categories.

Otherwise I'd change the behavior of this PR to reset or replace all categories. In addition a label would be introduced, that shows "and x other categories" if more than one is copied.

dbrgn:
> One possible solution might be to change the category display, to be condensed into a single line instead of a new one for each category layer. This would allow to display them in a list and add two simple add and delete buttons for managing the categories.

That sounds like a nice idea!

5n8ke:
I added Quick and "dirty" a *... and X more* label.
Duplicate Category
I think there is enough space in the Wizard to add a list of all categories but I'll wait to implement something like this until a consensus is reached.

libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.ui
```diff
@@ -121,6 +121,22 @@
...
+     <item>
+      <widget class="QLabel" name="lblMoreCategories">
+       <property name="sizePolicy">
+        <sizepolicy hsizetype="Maximum" vsizetype="Preferred">
+         <horstretch>0</horstretch>
+         <verstretch>0</verstretch>
+        </sizepolicy>
+       </property>
+       <property name="text">
+        <string>... and X more.</string>
```
ubruhin, line 133: Mark this string as not translatable to avoid unnecessary effort for translators:

```suggestion
        <string notr="true">... and X more.</string>
```

libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.ui
```diff
@@ -173,6 +189,15 @@
...
+ <resources>
+  <include location="../../../../img/images.qrc"/>
+  <include location="../../../../img/images.qrc"/>
+  <include location="../../../../img/images.qrc"/>
+  <include location="../../../../img/images.qrc"/>
+  <include location="../../../../img/images.qrc"/>
+  <include location="../../../../img/images.qrc"/>
+  <include location="../../../../img/images.qrc"/>
+  <include location="../../../../img/images.qrc"/>
+ </resources>
```
ubruhin, line 201: I have no idea why QtCreator sometimes adds this crap, but I usually discard these changes to avoid noise in the commits:

```suggestion
 <resources/>
```

libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.ui
```diff
@@ -164,7 +180,7 @@
         <string notr="true"/>
        </property>
        <property name="icon">
-        <iconset>
+        <iconset resource="../../../../img/images.qrc">
```
ubruhin, line 183: Same here:

```suggestion
        <iconset>
```

libs/librepcb/libraryeditor/newelementwizard/newelementwizardcontext.h
```diff
@@ -129,7 +129,8 @@ class NewElementWizardContext final : public QObject {
   QString                   mElementKeywords;
   QString                   mElementAuthor;
   tl::optional<Version>     mElementVersion;
-  tl::optional<Uuid>        mElementCategoryUuid;
+  tl::optional<Uuid>        mElementRootCategoryUuid;
+  QSet<Uuid>                mElementCategoriesUuid;
```
ubruhin, line 133: Wouldn't it be simpler to have only the QSet member? I mean, the optional<Uuid> could simply be generated on demand where needed (take one item from the set if it is not empty, else nullopt - it's impossible anyway that the set contains multiple items at the time we need the optional<Uuid>).

And btw, I usually name a list of UUIDs like *ElementUuids* instead of *ElementsUuid* :wink:

```suggestion
  QSet<Uuid>                mElementCategoryUuids;
```

libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.cpp
```diff
@@ -152,12 +152,19 @@ void NewElementWizardPage_EnterMetadata::btnChooseCategoryClicked() noexcept {
...
+    mContext.mElementCategoriesUuid.insert(*categoryUuid);
+  }
   updateCategoryTreeLabel();
 }
 
 void NewElementWizardPage_EnterMetadata::btnResetCategoryClicked() noexcept {
-  mContext.mElementCategoryUuid = tl::nullopt;
+  if (mContext.mElementCategoriesUuid.count()) {
+    mContext.mElementCategoriesUuid.clear();
+  }
```
ubruhin, line 166: Can be simplified :)

```suggestion
  mContext.mElementCategoriesUuid.clear();
```

libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.cpp
```diff
@@ -184,14 +191,21 @@ void NewElementWizardPage_EnterMetadata::updateCategoryTreeLabel() noexcept {
...
     }
     default: {
       mUi->lblCategoryTree->setText(tr("Root category"));
       break;
     }
   }
+  if (mContext.mElementCategoriesUuid.count() > 1) {
+    mUi->lblMoreCategories->setText(
+        tr(" ... and %1 more.")
+            .arg(mContext.mElementCategoriesUuid.count() - 1));
```
ubruhin, line 205: As a translator I'd consider leading or trailing spaces in a string as a mistake, therefore I usually move them to outside the tr():

```suggestion
        " " % tr("... and %1 more.")
                  .arg(mContext.mElementCategoriesUuid.count() - 1));
```

ubruhin (commented):
Works fine, thanks! :+1:

However, I'd actually prefer to store the UUIDs in only one `QSet<Uuid>` instead of an `optional<Uuid>` plus `QSet<Uuid>` since this is just redundance. Could you maybe change that?

5n8ke:
I can on Monday.

libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.ui
5n8ke: Done

libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.ui
5n8ke: Ok, I'll change that. It's quite annoying with the additional resource crap, but it seems to be only added once in this quantity (as of now :( )

libs/librepcb/libraryeditor/newelementwizard/newelementwizardpage_entermetadata.cpp
5n8ke: This does seem more like an oversight on my part. The label is right-aligned and does not need any trailing or leading spaces.

ubruhin (approved):
Perfect, thanks! :tada:

PR #773, opened 2020-09-26
Allow to modify polygons graphically

ubruhin:
Within the package-, symbol- and board editors, polygons can now be edited graphically:
- Dragging vertices of selected polygons
- Removing vertices by context menu on vertices
- Inserting vertices by context menu of line segments

librepcb_polygons


dbrgn:
Wohoo, finally! :tada:

dbrgn:
By the way, in many programs a vertex can be added by double-clicking. Maybe that would make sense here as well?

ubruhin:
> By the way, in many programs a vertex can be added by double-clicking. Maybe that would make sense here as well?

A double click on the line currently opens the properties dialog of the polygon, which I consider more important than inserting new vertices. And when double-clicking on a vertex (instead of the line), we don't know whether the new vertex should be inserted *before* or *after* the clicked vertex.

Or how do "many programs" handle this exactly? :)

dbrgn:
Ah, makes sense. Let's keep it like that for consistency!

PR #779, opened 2020-10-04
Use built-in markdown support when using Qt >=5.14

### LINKED ISSUE #777: Replace hoedown with cmark
##### SUMMARY

The library we use for markdown conversion - [hoedown](https://github.com/hoedown/hoedown) - is unmaintained since 2015. Furthermore, it's not included in any major Linux distribution.

##### SOLUTION

As an alternative, we could use [libcmark](https://github.com/commonmark/cmark), the reference implementation of the CommonMark specification. In contrast to hoedown, it is actively developed and it's packaged for all major Linux and BSD distributions (Arch, Alpine, CentOS, Debian, Fedora, FreeBSD, Ubuntu, ...).

I already wrote the necessary code, integrating libcmark was really simple. If you agree, I'd provide a PR with the necessary changes. I'd need you to clone cmark to the LibrePCB organization though.

dbrgn:
As discussed in the development chat: Instead of replacing hoedown with cmark (see #777), we can use Qt's built-in markdown support in Qt >=5.14. The rendering is slightly different than with hoedown (using the CommonMark standard), but since the binary releases are using a recent Qt version and up-to date distros should also ship a recent Qt version, most users will already be using the new rendering logic.

With this change, there will be one less vendored library for distro package maintainers to package (see #385).

Closes #777.

### DIFF
--- a/ci/azure-jobs-linux.yml
+++ b/ci/azure-jobs-linux.yml
@@ -14,6 +14,7 @@ jobs:
       Ubuntu_1604_Qt_5_14_2_GCC:
         IMAGE: librepcb/librepcb-dev:ubuntu-16.04-qt5.14.2-1
         DEPLOY: true
+        NO_HOEDOWN: true
       Ubuntu_1804_Clang:
         IMAGE: librepcb/librepcb-dev:ubuntu-18.04-2
         CC: clang
@@ -40,6 +41,9 @@ jobs:
   - bash: rm -rf libs/polyclipping
     displayName: Remove bundled polyclipping lib
     condition: and(succeeded(), contains(variables['UNBUNDLE'], 'polyclipping'))
+  - bash: rm -rf libs/hoedown
+    displayName: Remove bundled hoedown lib
+    condition: and(succeeded(), eq(variables['NO_HOEDOWN'], 'true'))
   - bash: ./ci/build_librepcb.sh
     displayName: Build LibrePCB
   - bash: ./ci/build_linux_appimage.sh
--- a/apps/librepcb-cli/librepcb-cli.pro
+++ b/apps/librepcb-cli/librepcb-cli.pro
@@ -28,7 +28,6 @@ LIBS += \
     -llibrepcbproject \
     -llibrepcblibrary \
     -llibrepcbcommon \
-    -lhoedown \
     -lmuparser \
     -lsexpresso \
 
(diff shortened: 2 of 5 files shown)


ci/azure-jobs-linux.yml
```diff
@@ -40,6 +41,9 @@ jobs:
   - bash: rm -rf libs/polyclipping
     displayName: Remove bundled polyclipping lib
     condition: and(succeeded(), contains(variables['UNBUNDLE'], 'polyclipping'))
+  - bash: rm -rf libs/hoedown
+    displayName: Remove bundled hoedown lib
+    condition: and(succeeded(), eq(variables['NO_HOEDOWN'], 'true'))
```
dbrgn, line 46: This should ensure that building without hoedown actually works. (I already tested this on my development machine, but it's always better to check it in CI as well.)

ubruhin:
Actually the idea of `MarkdownConverter` is to have a library independent API to parse MD files :wink: Couldn't we adjust only `markdownconverter.cpp` to keep having a single place for this task (and avoiding conditional file compilation, which doesn't work for header files anyway)? Probably we can use `QTextDocument::setMarkdown()` + `QTextDocument::toHtml()` instead of directly accessing `QTextEdit`(?) It's not a perfect solution, but IMHO it would be cleaner and easier to maintain than spreading the knowledge about the markdown library across totally unrelated units (ControlPanel in this case).

Another thing: Are you really sure to disable the GitHub markdown extensions? When looking at [these flags](https://code.woboq.org/qt5/qtbase/src/gui/text/qtextmarkdownimporter_p.h.html#QTextMarkdownImporter), somehow it looks like the github dialect just adds some features to the CommonMark dialect. Usually I totally agree on using commonly accepted specifications instead of proprietary extensions, but in this case probably it wouldn't hurt, does it?

dbrgn:
Alright, I updated the code to use a temporary `QTextDocument` in order to convert the MD input to a string. You're right, this results in a nicer API!

Regarding GFM vs CommonMark: It seems that - at least nowadays - GFM is a superset of CommonMark: https://github.github.com/gfm/#introduction It includes the following extensions:

- Tables
- Task list items
- Strikethrough
- Autolinks
- Disallow raw HTML

I'm not fond of the autolinking, but the rest is nice. In general, I'd would have preferred CommonMark to ensure that the supported input format stays the same even when switching the rendering library in the future, but since Qt ships a GFM implementation, that should not be required. Additionally, a lot of people will host their projects on GitHub, so it's nice if the README looks the same.

I've added a commit that switches to GFM, feel free to squash-merge that, or revert it before merging.

ci/azure-jobs-linux.yml
ubruhin: Sounds reasonable :+1: :slightly_smiling_face:

ubruhin (approved):
Thanks! :tada:

Issue #741, opened 2020-08-02
Add a way to duplicate project from the software


miniwark:
##### SUMMARY

The software do not allow to duplicate projects, either by a "Duplicate" or "Save As..." features. 

- This could be useful in case of "projects as a template": for example for boards with the same form factor (like Arduino shields)
- "Manual forking" to keep v1/v2/v3 ect. version of a board in there own project

##### SOLUTION

It's already possible to work around this outside the software by: 
- copy/paste a project folder
- rename the folder
- rename the main `.lpp` file
- edit the main title of the Readme file
- edit the name in the metadata file (preferably from menu: Project > Promerties > Name)

A "save As..." or "Duplicate" wizard would just do this automatically instead of manually.

PS: Alternately this sort of "forking" may be done with Git instead (with use of tags or branch names)

ollelogdahl:
I'd like to contribute to the project. looking into this!

ubruhin:
I think a "Save As..." would be the most intuitive variant (as it is well known from other programs), but probably also the most complicated one. We'd have to be very careful to really update the filepaths of all objects to avoid accidentally accessing the original project files after saving a projects to a new location.

Btw, I suggest to also assign a new UUID to the project (see class `ProjectMetadata`). The UUID has no clear technical functionality, but might be useful some day to distinguish between different projects.

> I'd like to contribute to the project. looking into this!

That would be great! :)

ollelogdahl:
Should this button be an option next to 'Save' option on both the Board Editor and the Schematic editor?

miniwark:
> Should this button be an option next to 'Save' option on both the Board Editor and the Schematic editor?

I think the best place would be in the main Control Panel instead.

PR #795, opened 2020-10-20
Saveas feature

ollelogdahl:
My implementation of the feature suggested in #741 

Works for the most part, missing 2 features:
- By default, the save dialog window requires all files to end in .lpp. This does not make sense for a folder. How should the dialog be called instead?
- The .lpp file doesn't get renamed to the same name as the new project, as the project files get strictly copied.

### DIFF
--- a/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
+++ b/libs/librepcb/projecteditor/boardeditor/boardeditor.cpp
@@ -208,6 +208,8 @@ BoardEditor::BoardEditor(ProjectEditor& projectEditor, Project& project)
   // connect some actions which are created with the Qt Designer
   connect(mUi->actionProjectSave, &QAction::triggered, &mProjectEditor,
           &ProjectEditor::saveProject);
+    connect(mUi->actionProjectSaveAs, &QAction::triggered,
+          [this]() { mProjectEditor.saveProjectAs(this); });
   connect(mUi->actionQuit, &QAction::triggered, this, &BoardEditor::close);
   connect(mUi->actionOpenWebsite, &QAction::triggered,
           []() { QDesktopServices::openUrl(QUrl("https://librepcb.org")); });
--- a/libs/librepcb/projecteditor/boardeditor/boardeditor.ui
+++ b/libs/librepcb/projecteditor/boardeditor/boardeditor.ui
@@ -78,6 +78,7 @@
      <string>&amp;File</string>
     </property>
     <addaction name="actionProjectSave"/>
+    <addaction name="actionProjectSaveAs"/>
     <addaction name="actionExportLppz"/>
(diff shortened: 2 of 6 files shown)


ollelogdahl:
Unsure why mac is failing. Seems to be unrelated.

ubruhin:
Thanks for working on this!

Without reviewing the code in detail yet, my biggest concern is that this this "Save as" feature behaves very differently compared to basically all other applications. A "Save as" usually saves the opened project to a new location, and any further modification on the project will be saved to the new location too. That's currently not the case - any further modification will be stored to the _old_ project location, which is probably not what the user expects.

To be honest, I'm not sure how complicated it will be to change this behavior. Maybe it's a matter of a few lines of code, maybe it's  a matter of hundreds of lines of code...

Another problem I experienced is that a lock file will be created immediately in the new project directory, which makes it impossible to open the new project. One has to restart LibrePCB to invalidate the lock.

> By default, the save dialog window requires all files to end in .lpp. This does not make sense for a folder. How should the dialog be called instead?

Hmm, actually the dialog doesn't require the file to end in .lpp. I suspect the "file must end with .lpp" error comes from `Project::create()` because the second argument needs to be the lpp file name (not path), but you passed the directory path instead.

dbrgn:
> Without reviewing the code in detail yet, my biggest concern is that this "Save as" feature behaves very differently compared to basically all other applications. A "Save as" usually saves the opened project to a new location, and any further modification on the project will be saved to the new location too. That's currently not the case - any further modification will be stored to the _old_ project location, which is probably not what the user expects.

This is what applications usually call "Save copy as" or "Save a copy" (e.g. Inkscape).

ollelogdahl:
Thank you for the feedback. You are right, it is kind of unintuitive. I will take a look at copying the project and switching to it, although not sure if it would require complete reload of the editor. If the change doesn't make sense, it might be worth going the Inkscape route. 

As said, looking into it. The .lpp thing should be easily fixed. I will be taking a deeper look at the project loading, and asking here if i get any more questions. The intended way to save using the current project and transactionaldirectory/-filesystem design is quite difficult to comprehend (atleast for my part).

ubruhin:
> This is what applications usually call "Save copy as" or "Save a copy" (e.g. Inkscape).

Ah, interesting, so far I never saw such a feature :)

> I will take a look at copying the project and switching to it, although not sure if it would require complete reload of the editor. If the change doesn't make sense, it might be worth going the Inkscape route.

Great! :tada:  Maybe it would not even be a big issue to reload the whole editor. We already do that when updating a project's library. Of course it would be much nicer without such a "workaround", but basically it works and is better than nothing ;)

ollelogdahl:
Great, good pointer. Will look at that code if the reload workaround is neccesary!

Issue #792, opened 2020-10-17
Unbundling libraries lead to messy project files and doesn't work reliable


ubruhin:
##### SUMMARY

In the past, we always linked statically to our dependencies bundled as submodules. Then we started to optionally allow dynamic linking and unbundling libraries: #698, #699, #742, #779, https://github.com/fontobene/fontobene-qt5/pull/2, https://github.com/fontobene/fontobene-qt5/pull/5, https://github.com/LibrePCB/docker-librepcb-dev/pull/4, https://github.com/LibrePCB/docker-librepcb-dev/pull/8

Due to these changes, many new problems popped up and required bugfixes or workarounds: #729, #778, #788, #789, #790, #791, https://github.com/fontobene/fontobene-qt5/pull/8, https://github.com/LibrePCB/quazip/pull/2, https://github.com/LibrePCB/quazip/pull/3, https://github.com/LibrePCB/docker-librepcb-dev/pull/5

Slowly I really struggle with all the dynamic linking and unbundling stuff. Before starting with these things, everything worked fine, it was very easy to build, on all platforms. The only problem was that packagers do not like bundled libraries. Now the project files get more and more complicated (or even messy!) to fix or work around problems introduced by the unbundling stuff. And most of these issues can't even be detected on CI as this would require building with dozens of different configurations. So for me this looks like a never ending story, wasting a lot of time and nerves.

Therefore I think this can't be the way to go. How do other projects handle dependencies to other libraries? Would CMake and/or a package manager like Conan be able to handle all use cases (static linking, dynamic linking, bundled libraries, system libraries) in a reliable way? What other options do we have to get a more reliable build system?

Please don't get me wrong @dbrgn - I really appreciate your work to help packagers adding LibrePCB to package managers. But now I really wonder if this is the way to go...

dbrgn:
I think most projects just link dynamically. A lot of the complexity right now doesn't come from "bundled vs system", but from "dynamic vs static".

I think if we'd create dynamically linked builds by default, that would make things easier already. Switching to CMake should also make things easier, because many libraries provide CMake scripts, so we can just bundle them without having to adjust build scripts (we could even use a "download external libraries" script instead of submodules, which might make things easier). CMake also has tools for library discovery, so we don't have to hardcode paths.

Not bundling libs at all would also be an option, but it would make it a bit more complex for contributors, especially on Windows.

Here's an overview of the libs we're using:

[...]

PR #802, opened 2020-11-01
Replace sexpresso library with own implementation

### LINKED ISSUE #783: Get rid of "sexpresso" dependency
##### SUMMARY

Currently we use the [sexpresso](https://github.com/BitPuffin/sexpresso) library to parse our S-Expression files. We wrap the library with our own class [`SExpression`](https://github.com/LibrePCB/LibrePCB/blob/master/libs/librepcb/common/fileio/sexpression.h) where we even implemented our own serializer instead of using sexpresso (to be able to control every linebreak etc. of generated files).

There's nothing wrong with that library, it works like a charm. But since we already implemented our own serializer, we actually use sexpresso only for parsing files. And since parsing S-Expression files is quite trivial, I think it's not worth to include a third party library (which packagers of LibrePCB possibly want to unbundle) for this simple task.

##### SOLUTION

IMHO we should implement the S-Expression parser on our own, right within the `SExpression` class. Then we can get rid of the sexpresso dependency. This also helps packagers because sexpresso doesn't need to be unbundled.

But of course the parser needs to be implemented properly (escaping, error handling, ...), including some unit tests.

ubruhin:
Replacing `sexpresso::parse()` with our own parser in the `SExpression` class. See #783 for the motivation behind this change.

Because we also used `sexpresso::escape()` to escape characters in strings to be serialized, I also had to replace this by our own implementation. Then I saw that `sexpresso::escape()` even replaced `?` by `\?`, `'` by `\'` and the _audible bell_ by `\a` (wtf?), which doesn't make sense to me :wink: So I removed these replacements, but only for LibrePCB 0.2.x to keep the 0.1.x file format as-is.

Positive side effect of the own implementation: It seems to be much faster than `sexpresso::parse()` :smiley: Not an accurate measurement, but on my computer the background library scan now takes ~25% less time than before (tested only in debug mode).

I just hope there are no critical bugs :see_no_evil: I added some unit tests and tested manually by opening/saving whole projects, at least so far it seems to work properly...

- Fixes #783.
- Useful for #792 since sexpresso no longer needs to be unbundled

### DIFF
--- a/libs/librepcb/common/fileio/sexpression.cpp
+++ b/libs/librepcb/common/fileio/sexpression.cpp
@@ -22,7 +22,7 @@
  ******************************************************************************/
 #include "sexpression.h"
 
-#include <sexpresso/sexpresso.hpp>
+#include "../application.h"
 
 #include <QtCore>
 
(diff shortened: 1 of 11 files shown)


libs/librepcb/common/fileio/sexpression.cpp
```diff
@@ -289,19 +302,151 @@ SExpression SExpression::createLineBreak() {
...
+
+  // Note: Until LibrePCB 0.1.5 we used the sexpresso library for escaping
+  // strings. This library escaped more characters than we do now. To still
+  // support reading the file format 0.1, we have to keep support for the
+  // old escaping behavior.
+  static QHash<QChar, QChar> escapedChars = {
+      {'\'', '\''},  // Single quote
+      {'"', '"'},  // Double quote
+      {'?', '\?'},  // Question mark
+      {'\\', '\\'},  // Backslash
```
dbrgn, line 391: The quotes and the backslash are the same before/after escaping. Is that on purpose?

dbrgn (commented):
The parser would be a good candidate for fuzzing, e.g. using [afl++](https://aflplus.plus/). Can be done at a later point in time though.

libs/librepcb/common/fileio/sexpression.cpp
ubruhin: These characters are only replaced if they appear after a backslash, i.e. `\"` is replaced by `"`. It just looks a bit strange because characters like `\` need to be escaped within the literal, so it looks like `\\` is replaced by `\\`, but actually it replaces `\\` by `\` :)

PR #804, opened 2020-11-16
SystemInfo: Add support for OpenBSD

ubruhin:
See https://librepcb.discourse.group/t/building-on-openbsd/320.

Now it compiles successfully and all unittests are passing. Tested on OpenBSD 6.8.

image

### DIFF
--- a/libs/librepcb/common/systeminfo.cpp
+++ b/libs/librepcb/common/systeminfo.cpp
@@ -45,6 +45,9 @@
 #if defined(Q_OS_SOLARIS)
 #include <libproc.h>
 #endif
+#if defined(Q_OS_OPENBSD)
+#include <sys/sysctl.h>
+#endif
 #elif defined(Q_OS_WIN32) || defined(Q_OS_WIN64)  // Windows
 #ifdef WINVER
 #undef WINVER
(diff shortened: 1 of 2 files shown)


libs/librepcb/common/systeminfo.cpp
```diff
@@ -233,6 +236,26 @@ QString SystemInfo::getProcessNameByPid(qint64 pid) {
...
   buf[len] = 0;
   processName = QFileInfo(QFile::decodeName(buf)).fileName();
+#elif defined(Q_OS_OPENBSD)
+  // https://man.openbsd.org/sysctl.2
+  // NOTE: This will return only the first 16 bytes of the process name. If
+  // someone finds a way to get the full process name, feel free to improve it.
+  kinfo_proc proc;
+  size_t procSize = sizeof(proc);
+  int mib[6] = {CTL_KERN,     KERN_PROC, KERN_PROC_PID, static_cast<int>(pid),
+                sizeof(proc), 1};
```
dbrgn, line 246: Just wondering, is this weird indentation on purpose? :slightly_smiling_face:

dbrgn (commented):
Cool! That screenshot looks like from 1995 :smile:

libs/librepcb/common/systeminfo.cpp
ubruhin: It's the output of clang-format, probably there is still some kind of alignment enabled in its rules...

ubruhin:
For the record, merging this even though CI failed because the failure is not related to this PR. Azure Pipelines has some performance issues today which seems to make the CPU heavy SQLite unittest fail :sob:

PR #810, opened 2020-12-20
Add support for polygons in schematics

### LINKED ISSUE #805: Add support for polygons in schematics
##### SUMMARY

It's quite common to draw some rectangles around related parts in a schematic, add some lines or arrows to highlight something, draw tables, etc. That's currently not possible in LibrePCB since it doesn't support adding polygons to schematics.

##### SOLUTION

Allow adding polygons to schematics, just like in the symbol editor. Should be quite straight forward since symbols already support polygons.

ubruhin:
Allow adding polygons to schematics:

librepcb_schematic_polygons

Fixes #805


dbrgn:
Cool! Would it be easy to add support for dotted lines?

ubruhin:
> Would it be easy to add support for dotted lines?

Not really... I created #812 to keep this feature request in mind.

PR #816, opened 2020-12-27
librepcb-cli: Log fatal messages to stderr

dbrgn:
These messages should always be visible on stderr, not just in the
logfile (which may not have been initialized properly if a fatal error
occurs).

### DIFF
--- a/apps/librepcb-cli/main.cpp
+++ b/apps/librepcb-cli/main.cpp
@@ -41,7 +41,11 @@ int main(int argc, char* argv[]) {
   // done as early as possible.
   Debug::instance();
 
-  // Silence debug output, it's a command line tool
+  // Silence logging output, it's a command line tool where logging messages
+  // could lead to issues when parsing the CLI output. Real errors will be
+  // printed to stderr explicitly and logging output can optionally be enabled
+  // with the "--verbose" flag. But still print fatal errors since this is the
+  // only way to print any error to stderr before the application gets aborted.
   Debug::instance()->setDebugLevelStderr(Debug::DebugLevel_t::Fatal);
 
   // Create Application instance
--- a/apps/librepcb-cli/main.cpp
+++ b/apps/librepcb-cli/main.cpp
@@ -42,7 +42,7 @@ int main(int argc, char* argv[]) {
   Debug::instance();
 
   // Silence debug output, it's a command line tool
-  Debug::instance()->setDebugLevelStderr(Debug::DebugLevel_t::Nothing);
+  Debug::instance()->setDebugLevelStderr(Debug::DebugLevel_t::Fatal);
 
   // Create Application instance
   Application app(argc, argv);


ubruhin:
Hm, what's the exact reason for this change? Actually the CLI logs errors to stderr already and exits with a non-zero exit code. Generally I really don't like to output log messages by default since this can lead to unreadable output, and could cause issues when parsing the CLI output.

If something goes wrong and you need more verbose output, just run the CLI with `--verbose` :wink:

dbrgn:
In the cmake branch, my build of `librepcb-cli` always crashed with SIGABRT, even before command line argument parsing:

```
[cmake]danilo@x2000:LibrePCB$ build/apps/librepcb-cli/librepcb-cli --help
Aborted (core dumped)
[cmake]danilo@x2000:LibrePCB$ build/apps/librepcb-cli/librepcb-cli --verbose
Aborted (core dumped)
```

There's no way to get any diagnostic output out of that except running a debug build through something like gdb.

With the change, I get the following output:

```
[cmake]danilo@x2000:LibrePCB$ build-debug/apps/librepcb-cli/librepcb-cli
[CRITICAL ] Could not find resource files! Probably packaging went wrong?! (/home/danilo/Projects/librepcb/LibrePCB/libs/librepcb/common/application.cpp:128)
[CRITICAL ] Expected resources location: "/mnt/data/Projects/librepcb/LibrePCB/build-debug/apps/share/librepcb" (/home/danilo/Projects/librepcb/LibrePCB/libs/librepcb/common/application.cpp:130)
[CRITICAL ] Executable location:         "/mnt/data/Projects/librepcb/LibrePCB/build-debug/apps/librepcb-cli/librepcb-cli" (/home/danilo/Projects/librepcb/LibrePCB/libs/librepcb/common/application.cpp:131)
[CRITICAL ] Build output directory:      "" (/home/danilo/Projects/librepcb/LibrePCB/libs/librepcb/common/application.cpp:133)
[CRITICAL ] Is developer build:          false (/home/danilo/Projects/librepcb/LibrePCB/libs/librepcb/common/application.cpp:135)
[  FATAL  ] Default stroke font could not be loaded! (/home/danilo/Projects/librepcb/LibrePCB/libs/librepcb/common/application.cpp:201)
Aborted (core dumped)
```

Much more helpful.

Usually critical and fatal errors should not happen in regular use, so they will never clutter the output, right?

ubruhin:
OK, I see the Problem. But I still don't like the solution ;)

The real problem is actually that the error handling in the `Application` class is miserable. Logging some messages and then aborting the application is really bad. Consider opening the LibrePCB GUI on Windows where you won't see the logging output (not even in the console!) you have exactly the same problem (application crashes and you have no idea why).

I think about refactoring the `Application` class since a long time because it is so crappy (not only the error handling). Error handling should not be done with logging. The CLI should print errors to stderr, and the GUI should show a message box. And the application should be quit normally, not by aborting.

But - after thinking about this again - it might actually still make sense to set the CLI logging level to FATAL because `qFatal()` actually aborts the application after printing the log message. So neither the CLI nor the GUI have any chance to properly handler this error.

So if you change the log level to FATAL, I think I'd approve this :slightly_smiling_face:

dbrgn:
> But - after thinking about this again - it might actually still make sense to set the CLI logging level to FATAL because qFatal() actually aborts the application after printing the log message.

Yeah, that's actually why I wanted to change it, and then I thought that logging critical errors in addition to fatal errors would make sense.

I still think it would be good to log critical errors as well (as a solution until the application class is refactored), but I can also change the default log level to fatal.

apps/librepcb-cli/main.cpp
```diff
@@ -42,7 +42,7 @@ int main(int argc, char* argv[]) {
   Debug::instance();
 
   // Silence debug output, it's a command line tool
```
ubruhin, line 44: Adding an explanation *why* the logging level is set to `Fatal`:

```suggestion
  // Silence logging output, it's a command line tool where logging messages
  // could lead to issues when parsing the CLI output. Real errors will be
  // printed to stderr explicitly and logging output can optionally be enabled
  // with the "--verbose" flag). But still print fatal errors since this is the
  // only way to print any error to stderr before the application gets aborted.
```

ubruhin:
>> Consider opening the LibrePCB GUI on Windows where you won't see the logging output (not even in the console!) you have exactly the same problem (application crashes and you have no idea why).
>>
>Yes, but at least there's a way to find out what the problem is without creating a debug build and attaching a debugger...

Hmm which way? AFAIK on Windows you don't get any stdout/stderr when starting a GUI application from the command line...

> I still think it would be good to log critical errors as well (as a solution until the application class is refactored)

I still don't like it, for at least two reasons :wink:
- There is no clear definition what a critical error actually means. A developer might call `qCritical()` in situations where the program actually works fine. Example: "workspace settings not found, will load default settings" -> is this a critical error? A warning? A debug message? I don't know, but for sure I don't want to see such a message in the CLI output as long as the requested task succeeded.
- This could lead to situations where the user sees messages with the prefix `[FATAL]`, but the CLI exited with code 0. IMHO that would be too confusing (I'd even consider this as a bug). If the task succeeded, no errors should be printed. And if the task failed, the exit code shall be nonzero.

dbrgn:
> Hmm which way? AFAIK on Windows you don't get any stdout/stderr when starting a GUI application from the command line...

Are you sure? I'm pretty sure you should be able to see stdout/stderr when running the command from cmd/powershell.

Anyways, PR should now be ready.

ubruhin:
> Are you sure? I'm pretty sure you should be able to see stdout/stderr when running the command from cmd/powershell.

Pretty sure :wink: Windows distinguishes between CLI applications and GUI applications, while only CLI applications have stdout/stderr attached to a console.

> Anyways, PR should now be ready.

Thanks! :+1:

PR #817, opened 2020-12-27
Add workspace override environment variable

Tubbles:
Please give lots of feedback on coding style etc :)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -16,3 +16,4 @@ yourself by creating a pull request (see [CONTRIBUTING.md](CONTRIBUTING.md)).
 - [@chrisgwerder](https://github.com/chrisgwerder)
 - [@0xB767B](https://github.com/0xB767B)
 - Josua Schmid ([@schmijos](https://github.com/schmijos))
+- [@Tubbles](https://github.com/Tubbles)
--- a/apps/librepcb/main.cpp
+++ b/apps/librepcb/main.cpp
@@ -170,6 +170,15 @@ static int runApplication() noexcept {
   FilePath path = Workspace::getMostRecentlyUsedWorkspacePath();
   qDebug() << "Recently used workspace:" << path.toNative();
 
+  // If the workspace path is specified by environment variable, use this one.
+  const char* wsEnvVarName = "LIBREPCB_WORKSPACE";
+  QString wsEnvStr = qgetenv(wsEnvVarName);
+  if (!wsEnvStr.isEmpty()) {
+    qInfo() << "Workspace path overridden by" << wsEnvVarName
+            << "environment variable:" << wsEnvStr;
+    path.setPath(wsEnvStr);
+  }
+
   // If creating or opening a workspace failed, allow to choose another
   // workspace path until it succeeds or the user aborts.
   while (true) {
--- a/dev/doxygen/pages/developers.md
+++ b/dev/doxygen/pages/developers.md
@@ -16,15 +16,19 @@ version can't be opened again with any other (stable or unstable) version!!!".
 So it's highly recommended to use a dedicated workspace for testing purposes,
 then the files in your productive workspace are safe. When working with real
 projects or libraries, make a backup first or at least use a version control
-system so any modifications can be reverted afterwards.
+system so any modifications can be reverted afterwards. This can also be achieved
+by setting the environment variable `LIBREPCB_WORKSPACE` to the workspace
+you want to use. If you use QtCreator, this variable can be set in the run
+configuration, so that this workspace is only used when starting LibrePCB from
+within QtCreator.
 
 Unstable application versions show a warning message box on every startup to
-protect users from (accidentylly) beaking their files (for example if they are
+protect users from (accidentally) beaking their files (for example if they are
 trying out a nightly build). For developers this message box is very annoying,
 so you can set the environment variable `LIBREPCB_DISABLE_UNSTABLE_WARNING=1` to
-disable it. If you use QtCreator, this variable can be set in the run
-configuration, so the warning is only disabled when starting LibrePCB from
-within QtCreator. But be careful with the disabled warning! :)
+disable it. This variable can also be set in the run configuration, so the
+warning is only disabled when starting LibrePCB from within QtCreator, similar
+to the one above. But be careful with the disabled warning! :)
 
 
 # Git {#doc_developers_git}


dev/doxygen/pages/developers.md
```diff
@@ -16,15 +16,19 @@ version can't be opened again with any other (stable or unstable) version!!!".
...
 projects or libraries, make a backup first or at least use a version control
-system so any modifications can be reverted afterwards.
+system so any modifications can be reverted afterwards. This can also be achieved
+by setting the environment variable `LIBREPCB_FORCE_WORKSPACE` to the workspace
+you want to use. If you use QtCreator, this variable can be set in the run
+configuration, so that this workspace is only used when starting LibrePCB from
+within QtCreator.
 
 Unstable application versions show a warning message box on every startup to
 protect users from (accidentylly) beaking their files (for example if they are
```
dbrgn, line 26: I spotted an unrelated typo, maybe you can fix that as well? :slightly_smiling_face: accidentylly -> accidentally

apps/librepcb/main.cpp
```diff
@@ -205,7 +205,22 @@ static bool isFileFormatStableOrAcceptUnstable() noexcept {
  ******************************************************************************/
 
 static FilePath determineWorkspacePath() noexcept {
-  FilePath wsPath(Workspace::getMostRecentlyUsedWorkspacePath());
+  static const QString FORCE_VAR_NAME("LIBREPCB_FORCE_WORKSPACE");
```
dbrgn, line 208: I think I'd prefer the variable to be `LIBREPCB_WORKSPACE` (there's no "forcing" involved, it's just another configuration option that takes precedence), but I don't know what @ubruhin thinks about it.

Tubbles: Agreed, LIBREPCB_WORKSPACE sounds better :) Wanna resolve this or wait for @ubruhin ?

apps/librepcb/main.cpp
```diff
@@ -205,7 +205,22 @@ static bool isFileFormatStableOrAcceptUnstable() noexcept {
  ******************************************************************************/
 
 static FilePath determineWorkspacePath() noexcept {
-  FilePath wsPath(Workspace::getMostRecentlyUsedWorkspacePath());
+  static const QString FORCE_VAR_NAME("LIBREPCB_WORKSPACE");
```
ubruhin, line 208: Couldn't we use a plain C string here, to avoid the conversion `.toStdString().c_str()` below? And since the env var is no longer named "force", it would be nice to rename the variable as well ;-)

Something like this:

```suggestion
  const char* workspaceEnvVarName = "LIBREPCB_WORKSPACE";
```

I'm not sure if needed, but if `qWarning()` below really needs a `QString`, the conversion can be done easily with `QString(workspaceEnvVarName)`.

apps/librepcb/main.cpp
```diff
@@ -205,7 +205,22 @@ static bool isFileFormatStableOrAcceptUnstable() noexcept {
...
+  FilePath wsPath;
+  QString wsForceStr = qgetenv(FORCE_VAR_NAME.toStdString().c_str());
+
+  if (wsForceStr != "") {
+    wsPath = FilePath(wsForceStr);
+    if (!Workspace::isValidWorkspacePath(wsPath)) {
+      qWarning() << FORCE_VAR_NAME << "set to invalid workspace:" << wsForceStr;
+      wsPath = FilePath("");
```
ubruhin, line 216: Not really critical, but I'd recommend to use the default constructor to explicitly create an invalid FilePath object:

```suggestion
      wsPath = FilePath();
```

AUTHORS.md
```diff
@@ -16,3 +16,4 @@ yourself by creating a pull request (see [CONTRIBUTING.md](CONTRIBUTING.md)).
 - [@chrisgwerder](https://github.com/chrisgwerder)
 - [@0xB767B](https://github.com/0xB767B)
 - Josua Schmid ([@schmijos](https://github.com/schmijos))
+- ([@Tubbles](https://github.com/Tubbles))
```
ubruhin, line 19: The parentheses around the whole line are probably a copy&paste mistake(?) :see_no_evil: 

```suggestion
- [@Tubbles](https://github.com/Tubbles)
```

apps/librepcb/main.cpp
```diff
@@ -205,7 +205,22 @@ static bool isFileFormatStableOrAcceptUnstable() noexcept {
...
 static FilePath determineWorkspacePath() noexcept {
-  FilePath wsPath(Workspace::getMostRecentlyUsedWorkspacePath());
+  static const QString FORCE_VAR_NAME("LIBREPCB_WORKSPACE");
+  FilePath wsPath;
+  QString wsForceStr = qgetenv(FORCE_VAR_NAME.toStdString().c_str());
+
+  if (wsForceStr != "") {
+    wsPath = FilePath(wsForceStr);
+    if (!Workspace::isValidWorkspacePath(wsPath)) {
+      qWarning() << FORCE_VAR_NAME << "set to invalid workspace:" << wsForceStr;
```
ubruhin, line 215: I wonder if this should actually be replaced (or extended) with a message box. If you don't actively read the console output, you might accidentally work with the wrong workspace since the workspace from the config file will silently be used as a fallback. And even worse, you have no way to abort loading the wrong workspace if the env var contains an invalid path. With a message box (containing a "Cancel" button), you are able to close LibrePCB before your productive v0.1 workspace gets loaded with LibrePCB v0.2 (for example).

Or what do you think about this @Tubbles?

ubruhin (commented):
Great! :smiley: 

I added some comments to the code...

apps/librepcb/main.cpp
ubruhin: :+1: for LIBREPCB_WORKSPACE :)

AUTHORS.md
Tubbles: Haha uuh, yeah.. Thanks for catching that 😆

apps/librepcb/main.cpp
Tubbles: Yeah maybe just close LibrePCB? Similar to how opening an already opened workspace closes LibrePCB?

Tubbles: I updated the logic, so it now resides directly in the main function (it became so simple by doing that..!). The application now does not load any workspace if the variable is set incorrectly, prints the error dialog, and prints a critical message to stderr.

ubruhin: >  Yeah maybe just close LibrePCB? Similar to how opening an already opened workspace closes LibrePCB?

Sounds good to me! :+1:

> I updated the logic, so it now resides directly in the main function (it became so simple by doing that..!).

To be honest, I liked it more when it was in the `determineWorkspacePath()` function :wink: Now there is (a little bit of) duplicate code because `openWorkspace()` is called from two different locations. If the env var detection is located in  `determineWorkspacePath()`, `openWorkspace()` is called only from one location and the overall logic in `main()` stays very simple to understand. Note that to enforce closing LibrePCB, you could simply call `return FilePath();` within `determineWorkspacePath()`, then `main()` will not open any workspace.

Btw, I just had the idea that the error handling with the message box could even be simplified. Actually you could remove the call to `isValidWorkspacePath()` (i.e. not validating the workspace path from the env var), then you'll get a message box automatically since `openWorkspace()` will fail afterwards. Sure, the error message will then be missing the information that the env var caused to load that workspace, but as an alternative you could previously call something like `qInfo() << "Loading workspace from environment variable LIBREPCB_WORKSPACE";` so at least in the terminal we'd see *why* this invalid workspace was opened.

But that's my personal preference (I like to keep code simple :grin:), I'm totally fine to accept your current solution with explicitly opening a message box :slightly_smiling_face:

ubruhin: @Tubbles any update on this topic? :slightly_smiling_face:

Tubbles: Not much lately, been alot at work 😔

ubruhin: OK, no problem :) Just let me know if I can help somehow.

ubruhin: @Tubbles With the small refactoring done in #906, this feature has now become even easier to implement (but introduced merge conflicts) - thus I allowed myself to update this PR accordingly :see_no_evil: I hope this is fine for you.

The behavior is now as following: If an invalid workspace path is specified by `LIBREPCB_WORKSPACE`, an error message appears that the workspace could not be opened. After clicking this message away, the workspace chooser dialog appears so the user can immediately choose another workspace - or click "cancel" to close the application.

Tubbles: Yes of course 😄🥳 this seems to mesh nicely with #906!

ubruhin: Perfect, so let's merge this! :rocket:

Thanks for your contribution! :tada:

PR #824, opened 2021-01-01
Improve error messages when project library update fails

ubruhin:
If the pinout of a device or component was changed after adding such elements to a project, the project library updater might fail due to invalid connections when running the next time. This is expected behavior (thus a warning appears in the library editor), but now the user at least gets a much more expressive error message. This way it's easier to determine which library element caused the error.

Since these are still pretty low level error messages, I didn't made them translatable, as decided in https://librepcb.discourse.group/t/119.

Fixes https://librepcb.discourse.group/t/project-update-library-fail-error-source-identification/344.


dbrgn:
Nice! Better error messages are always useful :slightly_smiling_face:

PR #826, opened 2021-01-04
Unit Test: make unit tests pass without warnings on win7-64 with Qt 5.15

andyfrommel:


libs/librepcb/common/systeminfo.cpp
```diff
@@ -279,6 +284,13 @@ QString SystemInfo::getProcessNameByPid(qint64 pid) {
   if (processName.endsWith(" (deleted)"))
     processName.chop(strlen(" (deleted)"));
 #elif defined(Q_OS_WIN32) || defined(Q_OS_WIN64)  // Windows
+  if (!isProcessRunning(pid)) {
```
ubruhin, line 287: I understand that this might fix failed unit tests in some cases. But to be honest, somehow I don't like to call `isProcessRunning()` here since beside a unit test, `getProcessNameByPid()` is never called with a PID of a killed process. `DirectoryLock` is the only location where `getProcessNameByPid()` is called, and only with a PID previously checked with `isProcessRunning()`. Now with this change, the `isProcessRunning()` function is called two times every time we want to determine a process name.

IMHO we could simply change the documentation of `getProcessNameByPid()` that the returned value is not guaranteed to be correct if the PID does not represent a running process (i.e. the user has to call `isProcessRunning()` first). And then just remove the unit tests which call `getProcessNameByPid()` with a PID of a non-running process.

Or what do you think about this @andyfrommel?

tests/unittests/common/systeminfotest.cpp
```diff
@@ -103,11 +103,15 @@ TEST_F(SystemInfoTest, testIsProcessRunning) {
      QT_VERSION_CHECK(5, 3, 0))  // QProcess::processId() requires Qt>=5.3
   {
     QProcess process;
-    process.start(getTestProcessExePath().toStr());
+    QString sProgramName = getTestProcessExePath().toStr();
+    QStringList slArguments; // no args needed
+    process.start(sProgramName, slArguments);
```
ubruhin, line 108: May I ask what's the intention of this change (and the same change a few lines below)? Isn't it just more verbose code with exactly the same behavior as previously on a single line?

ubruhin (commented):
Thanks for this contribution! :tada:  I added some comments below with my thoughts. And btw, I like the comments added to `isProcessRunning()` :+1: :smiley:

libs/librepcb/common/systeminfo.cpp
andyfrommel: A failed unit test means either the production code or the unit test is faulty.
My first opinion was the production code needs a change - because the unit tests
seemed to work on all other users dev computer and the failure did only appear on
mine - but i was wrong. The unit test does not have a try-catch section, so it
does not consider all possible reactions of the tested method.

Removing a unit test that sometimes fails? better not, been there (even if the method
is not used anywhere else in the code).
A unit test should test the production code against the spec. It can happen that it
is not specified how a method should react under a certain condition and then its
result is not predictable for the unit test. In those cases all possible results should
be merged together as a combined result (or even better: the spec should be reworked)

I have reverted the code in systeminfo.cpp and changed the unit test accordingly.
The other two throw() cases in the tested method are still hard..impossible to
provoke in order to test them in a unit test.

tests/unittests/common/systeminfotest.cpp
andyfrommel: **primary intention:** to change it.
triggered by the warning i got in the qt ide:

grafik

grafik

this was discussed in:
https://forum.qt.io/topic/116527/deprecation-of-qprocess-start-and-execute-without-argument-list

the start method with only a string as a mandatory attribute will disappear, which, in turn means this LOC will
break the build when it comes to the move to qt6, it needs to be changed some time.
(p.s. and i have overseen the third usage of process.start() that has a deprecated warning)

**secondary intention:** the way to change it.
this is very much a matter of opinion.
the shortest change would be 
    process.start(getTestProcessExePath().toStr(), QStringList());

but with this line of code my fingers instantly started to itch.
it does not really express WHAT it is doing, start a path with an undisclosed empty string list?

following some clean code advices:
introducing variables gives the chance to emphasize the intention and when needed, gives a
place to add some WHY explanations.

    QString fullProgramName = getTestProcessExePath().toStr();
    QStringList callArguments; // no args needed for this process
    process.start(sProgramName, slArguments);

the names of the local variables have been chosen according the signature of the start(...)
method. (and the hungarian notation for locals slipped-in, sorry).

there are (of course) other opinions out there about this trial to improve readability;
it is called misuse or even abuse of local variables.
that's the point where the discussions end in a dispute and should be stopped as no
party will ever agree to the argumentation of the other...

a compromise to the two versions would be the change of the way the test data is retrieved:
    process.start(getTestProgramFullName(testNr), getTestProgramArguments(testNr));

this would also give a chance to work with different test data combinations out of a set e.g. testNr = 3;

ubruhin: Ah OK I see, thanks for the explanation! :+1:

Personally I'd have chosen the compact solution:

```cpp
 process.start(getTestProcessExePath().toStr(), QStringList());
```

But I'm also fine with your solution, so let's keep it :)

libs/librepcb/common/systeminfo.cpp
ubruhin: OK I agree it's better to adjust the test to also accept throwing exceptions in addition to returning an empty string :+1:

However, honestly I don't like some things in the current test implementation:

- It checks for the string "QueryFullProcessImageNameW() failed" - but this string is Windows specific so the test will be considered as failed if an exception is thrown on another operating system
- Since usually an empty string is returned (instead of throwing an exception), the "catch" part of the test is almost never executed. So if we ever change some implementation detail of the `getProcessNameByPid()` method which leads to a different exception message, the test will still be successful on most systems, but might fail on systems where an exception is thrown (like yours) - so we'll have the same problem with an unreliable unit test again :wink:
- The test is quite complex and verbose :)

What about a much simpler solution, which doesn't check the exception message (since it is pretty irrelevant)? Suggestion:

```cpp
try {
  processName = SystemInfo::getProcessNameByPid(pid);
} catch (const Exception&) {
  processName = QString();  // consider as process not running
}
EXPECT_EQ(QString(), processName) << qPrintable(processName);
```

Btw the coding style of the changes does not follow our coding style, see [here](https://developers.librepcb.org/df/d24/doc_code_style_guide.html) what coding style we have and how to format it automatically with clang-format. If you're not able to run the formatter, no problem, we could do it for you :slightly_smiling_face:

ubruhin:
Closing this due to inactivity - sorry. Feel free to reopen (and rebase) if there is still a need for this change.

PR #857, opened 2021-02-16
CI: Update to latest runners and Docker images

ubruhin:
To make sure LibrePCB compiles with the latest Qt version 5.15.2 (at least on Windows), and to have CMake available which we need for #798. On Ubuntu 16.04, Qt 5.15.2 unfortunately doesn't work for us since it is missing SSL support, thus we stick with Qt 5.14.2 for now.

### DIFF
--- a/ci/azure-jobs-doxygen.yml
+++ b/ci/azure-jobs-doxygen.yml
@@ -1,8 +1,8 @@
 jobs:
 - job: Doxygen
   pool:
-    vmImage: 'ubuntu-16.04'
-  container: librepcb/librepcb-dev:ubuntu-18.04-2
+    vmImage: 'ubuntu-20.04'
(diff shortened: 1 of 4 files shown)


dbrgn (approved):
Cool, now I have a reason to rebase #798 again 😉

rnestler:
> On Ubuntu 16.04, Qt 5.15.2 unfortunately doesn't work for us since it is missing SSL support, thus we stick with Qt 5.14.2 for now.

Why not deprecate Ubuntu 16.04, it is EOL in a few month anyways.

ubruhin:
> Why not deprecate Ubuntu 16.04, it is EOL in a few month anyways.

Because of [this](https://docs.appimage.org/reference/best-practices.html#binaries-compiled-on-old-enough-base-system).

PR #871, opened 2021-02-28
Allow opening locked workspace/project/library in some cases

### LINKED ISSUE #734: The workspace cannot be opened after library edit or crash

```
LibrePCB Version: 0.1.4 from Flathub
Git Revision:
Qt Version:
Operating System: Linux Manjaro (Arch)
...
```

##### SUMMARY

When starting the application i get the following message, even after a reboot:

> The workspace "/home/miniwark/Documents/Developpements/PCB/LibrePCB" cannot be opened: The workspace is already opened by another application instance or user!

The LibrePCB just quit. (may be related to #154)

##### STEPS TO REPRODUCE

It had happen randomly times already. I think it's related to crash and/or bad saves during local Library editing (but a also have the case without crashes).

I suspect than there is a temporary file somewhere (instead of a PID) who keep track of open library parts, but who is not correctly deleted.

##### EXPECTED RESULTS

LibrePCB must offer me a way to try to force my way after the warning instead of just quitting.

ubruhin:
If a workspace, project or library directory is locked by a `.lock` file but we are not 100% sure if the lock is still valid, the error message dialog now provides the ability to discard the lock and open the directory anyway:

image

This is especially useful when running LibrePCB as a Flatpak since due to PID namespacing we are not able to check if the application holding a lock is still running or not.

But if we are absolutely sure the lock is still valid, this option is not available since generally it could be dangerous to discard file locks.

Fixes #734

### DIFF
--- a/libs/librepcb/common/fileio/directorylock.cpp
+++ b/libs/librepcb/common/fileio/directorylock.cpp
@@ -68,7 +68,8 @@ void DirectoryLock::setDirToLock(const FilePath& dir) noexcept {
  *  Getters
  ******************************************************************************/
 
-DirectoryLock::LockStatus DirectoryLock::getStatus() const {
+DirectoryLock::LockStatus DirectoryLock::getStatus(
+    QString* lockedByUser) const {
   // check if the directory to lock does exist
   if (!mDirToLock.isExistingDir()) {
     throw RuntimeError(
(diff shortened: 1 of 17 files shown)


kilt-er:
Awesome!!  Thanks

libs/librepcb/common/fileio/directorylock.cpp
```diff
@@ -99,29 +100,44 @@ DirectoryLock::LockStatus DirectoryLock::getStatus() const {
...
-  // now check if that process is still running (if not, the lock is considered
-  // as stale)
-  if (SystemInfo::isProcessRunning(lockPid)) {
-    if (SystemInfo::getProcessNameByPid(lockPid) == lockAppName) {
-      return LockStatus::Locked;  // the application which holds the lock is
-                                  // still running
+  // Check if the lock was created by this application instance.
+  if (lockPid == thisPid) {
+    if (dirsLockedByThisAppInstance().contains(mDirToLock)) {
+      return LockStatus::LockedByThisApp;
```
dbrgn, line 120: Very nice trick! 🙂

PR #874, opened 2021-03-03
SQLiteDatabaseTest: Simplify flaky concurrency test case

ubruhin:
After several years of annoying issues with this unit test on Windows (because Windows cannot reliably remove files, run multiple threads at the same time, and because Windows is incredibly slow), I have no more nerves to maintain this unittest. So I heavily simplified it to use only a single thread in addition to the main thread. This is actually the main use-case for SQLiteDatabase, so I think we don't loose a lot of important test coverage.

I really hope this finally fixes the constantly failing CI pipelines. It is a huge pain to have so many CI failures.


ubruhin:
CI was still flaky due to the functional tests with Funq since the DLL injection does not work reliable. Now I set up pytest-rerunfailures to automatically rerun the failed tests. It's not worth to spend even more hours investigating these issues, it's just wasted time... Maybe now CI gets useful again.

PR #875, opened 2021-03-04
Component editor: Remove checkbox text "required/optional"

### LINKED ISSUE #475: Connection "required/optional" checkbox
##### SUMMARY
In my opinion, the 'connection required' column can be improved. 
I think it is not intuitive if you need to check a checkbox with the label 'optional' to change the status to 'required'.

##### SOLUTION

Rename the column into 'Connection required' and just leave the checkboxes.

required

ubruhin:
Removed the text of the component signal checkbox for choosing "required" or "optional" to avoid confusion what state the checkbox actually has. Instead, the column title "Connection" is now changed to "Required" and the tooltip is more descriptive.

image

It's a bit ugly now since the checkbox is not centered. I tried to center it but found no reliable, simple way to do so...

Fixes #475


dbrgn (approved):
Great!

PR #879, opened 2021-03-06
Support mirroring alignment of schematic net labels

bjaminn:
Allow the NetLabel(s) to be aligned based on the standard 9 position Alignment.
Works with the CmdMirrorSelectedSchematicItems command.

### DIFF
--- a/libs/librepcb/common/geometry/netlabel.cpp
+++ b/libs/librepcb/common/geometry/netlabel.cpp
@@ -46,15 +46,16 @@ NetLabel::NetLabel(const Uuid& uuid, const NetLabel& other) noexcept
 }
 
 NetLabel::NetLabel(const Uuid& uuid, const Point& position,
-                   const Angle& rotation) noexcept
-  : onEdited(*this), mUuid(uuid), mPosition(position), mRotation(rotation) {
+                   const Angle& rotation, const Alignment& alignment) noexcept
+  : onEdited(*this), mUuid(uuid), mPosition(position), mRotation(rotation), mAlignment(alignment) {
 }
 
 NetLabel::NetLabel(const SExpression& node, const Version& fileFormat)
   : onEdited(*this),
     mUuid(deserialize<Uuid>(node.getChild("@0"), fileFormat)),
     mPosition(node.getChild("position"), fileFormat),
-    mRotation(deserialize<Angle>(node.getChild("rotation/@0"), fileFormat)) {
+    mRotation(deserialize<Angle>(node.getChild("rotation/@0"), fileFormat)),
+    mAlignment(node.getChild("alignment"), fileFormat) {
 }
 
 NetLabel::~NetLabel() noexcept {
@@ -94,6 +95,16 @@ bool NetLabel::setRotation(const Angle& rotation) noexcept {
   return true;
 }
 
+bool NetLabel::setAlignment(const Alignment &alignment) noexcept {
+  if (alignment == mAlignment) {
+    return false;
+  }
+
+  mAlignment = alignment;
+  onEdited.notify(Event::AlignmentChanged);
+  return true;
+}
+
 /*******************************************************************************
  *  General Methods
  ******************************************************************************/
@@ -102,6 +113,7 @@ void NetLabel::serialize(SExpression& root) const {
   root.appendChild(mUuid);
   root.appendChild(mPosition.serializeToDomElement("position"), true);
   root.appendChild("rotation", mRotation, false);
+  root.appendChild(mAlignment.serializeToDomElement("alignment"), false);
 }
 
 /*******************************************************************************
--- a/libs/librepcb/common/geometry/netlabel.cpp
+++ b/libs/librepcb/common/geometry/netlabel.cpp
@@ -37,7 +37,8 @@ NetLabel::NetLabel(const NetLabel& other) noexcept
   : onEdited(*this),
     mUuid(other.mUuid),
     mPosition(other.mPosition),
-    mRotation(other.mRotation) {
+    mRotation(other.mRotation),
+    mAlignment(other.mAlignment) {
 }
 
 NetLabel::NetLabel(const Uuid& uuid, const NetLabel& other) noexcept
@@ -47,7 +48,11 @@ NetLabel::NetLabel(const Uuid& uuid, const NetLabel& other) noexcept
 
 NetLabel::NetLabel(const Uuid& uuid, const Point& position,
                    const Angle& rotation, const Alignment& alignment) noexcept
-  : onEdited(*this), mUuid(uuid), mPosition(position), mRotation(rotation), mAlignment(alignment) {
+  : onEdited(*this),
+    mUuid(uuid),
+    mPosition(position),
+    mRotation(rotation),
+    mAlignment(alignment) {
 }
 
 NetLabel::NetLabel(const SExpression& node, const Version& fileFormat)
@@ -55,7 +60,12 @@ NetLabel::NetLabel(const SExpression& node, const Version& fileFormat)
     mUuid(deserialize<Uuid>(node.getChild("@0"), fileFormat)),
     mPosition(node.getChild("position"), fileFormat),
     mRotation(deserialize<Angle>(node.getChild("rotation/@0"), fileFormat)),
-    mAlignment(node.getChild("alignment"), fileFormat) {
+    mAlignment() {
+  auto parsedAlignment = node.tryGetChild("alignment");
+  if (parsedAlignment) {
+    Alignment tempAligment(*parsedAlignment, fileFormat);
+    mAlignment = tempAligment;
+  }
 }
 
 NetLabel::~NetLabel() noexcept {
@@ -95,7 +105,7 @@ bool NetLabel::setRotation(const Angle& rotation) noexcept {
   return true;
 }
 
-bool NetLabel::setAlignment(const Alignment &alignment) noexcept {
+bool NetLabel::setAlignment(const Alignment& alignment) noexcept {
   if (alignment == mAlignment) {
     return false;
   }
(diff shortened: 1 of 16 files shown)


ubruhin:
Hi,

Thanks for the contribution! I have some comments:

- The Alignment class allows to specify 9 different positions. But you only check if it is right aligned or not (2 positions). So the Alignment class is wrong for this usecase. I'd say a boolean property called "mirror" would be sufficient (and consistent with the "mirror" property of symbols). I quickly checked Eagle, it does it exactly the same way.
- The file format version is not taken into account when loading S-Expression files, thus with this patch it is not possible anymore to open existing projects. The new property must not be read when loading from file format 0.1 (upgrade mechanism). This can be done the following way:
  - In the constructors initializer list, initialize the new member variable with the default value.
  - In the constructors body, initialize the new member variable inside this condition:
    ```cpp
    if (fileFormat >= Version::fromString("0.2")) {
      // Read new property from S-Expression node.
    }
    ```
- In `SGI_NetLabel::updateCacheAndRepaint()` you removed some handling of rotation. I didn't test it yet but I suspect now rotation is not working properly anymore :wink: Please check this.
- In the `NetLabel` class, the new property must be handled in the following methods as well:
  - Copy constructor
  - `operator==()`
  - `operator=()`
- If possible, run `clang-format` to format the code according our coding style. See [here](https://developers.librepcb.org/df/d24/doc_code_style_guide.html) how this can be done (if you have Docker on Linux, it's very easy). Otherwise, don't worry, I can push a new commit to this branch to fix the coding style violations.

bjaminn:
I really am excited about this project and have wanted to contribute. 

* My view on using the alignment instead of just a boolean is based on my experience with some other EDA tools that allow the 9 position alignment. I only implemented the Right alignment because currently that is the only value that seems reasonably doable with the given user interface. It seems to use something other than bottom it would require adding a vertical flip. And center would require something to address that. Being a new contributor I don't know the long term vision of how this should be implemented. 
* I have updated the S-Expression constructor with the constructor body updating the member. My C++ is a little rusty so I don't know if that is the most efficient way.
* Yes you were correct I botched this up. That function should now be fixed. 
* Again my C++ is rusty and I forgot about these on the first update. They should be good now.
* I am pretty sure I have this fixed now. I was able to run it in Ubuntu with WSL2. I did modify the script to make it work native without docker. I will add that as a new PR in the future once I make it more robust.

dbrgn:
> I am pretty sure I have this fixed now. I was able to run it in Ubuntu with WSL2. I did modify the script to make it work native without docker. I will add that as a new PR in the future once I make it more robust.

Note that clang-format is (very unfortunately) not compatible between different versions. Different versions may generate different output, and some versions may even fail to parse the configfile. For a non-Docker-based approach, you'd need to guarantee that the correct version is installed, and show a warning otherwise. (I'm not sure what @ubruhin would prefer here.)

ubruhin:
> My view on using the alignment instead of just a boolean is based on my experience with some other EDA tools that allow the 9 position alignment. I only implemented the Right alignment because currently that is the only value that seems reasonably doable with the given user interface. It seems to use something other than bottom it would require adding a vertical flip. And center would require something to address that. Being a new contributor I don't know the long term vision of how this should be implemented.

What other EDA tool do you mean? I checked KiCad 5 and Eagle 7.5, both do not provide such a complex alignment feature. And in my opinion, this is fine (keep it simple, do not implement a complex feature where a simple one is totally sufficient). And as we see in your changes, you implemented only left vs. right alignment, but not all 9 positions, probably because it is not needed :wink: In addition, our netlabels have an auto-rotate feature: If a label is rotated by 180°, the text is not rendered upside down, but the alignment has changed from lower left to upper right. So in fact, there *is* already some more advanced alignment functionality. When implementing a full 9-position alignment on top of it, we'd have multiple combinations for the same result (e.g. 0° rotation with upper right alignment is exactly the same as 180° rotation with lower left alignment). This would be a nightmare to maintain :wink:

So I really think the simplest option is the best: Just allow to choose between "mirrored" and "not mirrored", i.e. using a simple boolean value.

> I have updated the S-Expression constructor with the constructor body updating the member. My C++ is a little rusty so I don't know if that is the most efficient way.

As I wrote above, just wrap the S-Expression parsing into an `if` statement which checks the file format version. Please don't use `tryGetChild()` in this case since it silently fails if the deserialization logic is not compatible with the serialization (e.g. different node name used - we already had exactly the same issue at another place where `tryGetChild()` was used). With the `if` condition and then `getChild()`, this risk is eliminated.

> Yes you were correct I botched this up. That function should now be fixed.

OK I'll check if it works properly soon. But could you now please remove the commented out line of code?

> Again my C++ is rusty and I forgot about these on the first update. They should be good now.

Looks good! :+1:

> I am pretty sure I have this fixed now. I was able to run it in Ubuntu with WSL2. I did modify the script to make it work native without docker. I will add that as a new PR in the future once I make it more robust.

Hmm I'm not sure if clang-format always has the `-6.0` suffix. At least if you install clang-format on Windows (natively), I guess it does not have this suffix. Instead, on Linux I recommend to set the clang-format version to use with this command:

```bash
update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-6.0 50
```

Afterwards, `clang-format` is an alias for `clang-format-6.0` so the script works properly without modification. Maybe this works on WSL as well?

bjaminn:
> > My view on using the alignment instead of just a boolean is based on my experience with some other EDA tools that allow the 9 position alignment. I only implemented the Right alignment because currently that is the only value that seems reasonably doable with the given user interface. It seems to use something other than bottom it would require adding a vertical flip. And center would require something to address that. Being a new contributor I don't know the long term vision of how this should be implemented.
> 
> What other EDA tool do you mean? I checked KiCad 5 and Eagle 7.5, both do not provide such a complex alignment feature. And in my opinion, this is fine (keep it simple, do not implement a complex feature where a simple one is totally sufficient). And as we see in your changes, you implemented only left vs. right alignment, but not all 9 positions, probably because it is not needed 😉 In addition, our netlabels have an auto-rotate feature: If a label is rotated by 180°, the text is not rendered upside down, but the alignment has changed from lower left to upper right. So in fact, there _is_ already some more advanced alignment functionality. When implementing a full 9-position alignment on top of it, we'd have multiple combinations for the same result (e.g. 0° rotation with upper right alignment is exactly the same as 180° rotation with lower left alignment). This would be a nightmare to maintain 😉
> 
> So I really think the simplest option is the best: Just allow to choose between "mirrored" and "not mirrored", i.e. using a simple boolean value.

I have experience with DxDesigner though not saying that is a model to follow. I personally am fine with right or left bottom alignment for netlabels. I totally missed the fact that the 180 flip combined with the mirrored or not means that we can get a 4 position option.

Do you think it is worth doing center or middle alignment?

> > I have updated the S-Expression constructor with the constructor body updating the member. My C++ is a little rusty so I don't know if that is the most efficient way.
> 
> As I wrote above, just wrap the S-Expression parsing into an `if` statement which checks the file format version. Please don't use `tryGetChild()` in this case since it silently fails if the deserialization logic is not compatible with the serialization (e.g. different node name used - we already had exactly the same issue at another place where `tryGetChild()` was used). With the `if` condition and then `getChild()`, this risk is eliminated.

I follow the reasoning now. I was trying to avoid exceptions on poorly formatted input, but i see the value in raising an exception.

> > Yes you were correct I botched this up. That function should now be fixed.
> 
> OK I'll check if it works properly soon. But could you now please remove the commented out line of code?

Followed up a mistake with another one. Should be good now.

> 
> > Again my C++ is rusty and I forgot about these on the first update. They should be good now.
> 
> Looks good! 👍
> 
> > I am pretty sure I have this fixed now. I was able to run it in Ubuntu with WSL2. I did modify the script to make it work native without docker. I will add that as a new PR in the future once I make it more robust.
> 
> Hmm I'm not sure if clang-format always has the `-6.0` suffix. At least if you install clang-format on Windows (natively), I guess it does not have this suffix. Instead, I recommend to set the clang-format version to use with this command, if needed (e.g. on Linux):
> 
> ```shell
> update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-6.0 50
> ```
> 
> Afterwards, `clang-format` is an alias for `clang-format-6.0` so the script works properly without modification. Maybe this works on WSL as well?
> 
> This way the script will work even if you don't have exactly version 6.0 installed (although you should not use a version other than 6.0, it's better to use another version than not running the script at all 😁).

The clang-format mod I made was not supposed to be included in this set. I actually figured it should be included as a separate atomic commit. I also improved my version to use an environment variable. My update-alternatives wasn't working like i wanted it to so your command clarifies it.

bjaminn:
I ran into some issues with the figuring out the how to migrate the test database. I changed the file format version on my code so that it would skip the mirror attribute. Without using the tryGetChild I don't know how to have it default to a specific value. Any suggestions?

ubruhin:
I just tested the feature and found an issue. Mirroring only works correctly with netlabels which have a rotation of either 0° or 180°. Netlabels with different rotation will have a wrong rotation after mirroring. For example this:

image

gets mirrored to this:

image

I think the rotation needs to be inverted (* -1) when mirroring (our order of transformations is: mirror, then rotate, then translate).

In addition, somehow it would be nice to support mirroring directly when placing a new netlabel. Maybe it would be quite easy by implementing `processMirror()` in the `SchematicEditorState_AddNetLabel` class. Maybe you want to give this a try as well? However, it's not a must-have feature for me, I'd also accept this PR without that feature :)
`

ubruhin:
Thanks for the update, I'd say now it's good to merge! :tada: So this will be included in LibrePCB 0.2.0 (due to the file format change, we can't include it in 0.1.6).

Btw, nice catch with the bounding box fix :+1:

PR #880, opened 2021-03-12
Allow overriding clang-format with environment variable

bjaminn:
Allow the `CLANGFORMAT` environmental variable to override `clang-format`

### DIFF
--- a/dev/format_code.sh
+++ b/dev/format_code.sh
@@ -18,6 +18,7 @@ set -euo pipefail
 
 DOCKER=""
 DOCKER_CMD="docker"
+CLANGFORMAT=${CLANGFORMAT-clang-format}
 ALL=""
 for i in "$@"
 do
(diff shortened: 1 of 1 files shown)


dev/format_code.sh
```diff
@@ -18,6 +18,7 @@ set -euo pipefail
 
 DOCKER=""
 DOCKER_CMD="docker"
+CLANGFORMAT=${CLANGFORMAT-clang-format}
```
dbrgn, line 21: That syntax looks weird. Shouldn't this be `${CLANGFORMAT:clang-format}`?

bjaminn: I had found it after a google search. After more research it might be `${CLANGFORMAT:-clang-format}`
https://wiki.bash-hackers.org/syntax/pe#use_a_default_value
Will overwrite the unset and the variables set to null

dbrgn: Ah, yeah, that's what I meant to write above 🙂 Can you update your pull request?

ubruhin: @dbrgn now it should be fine, right?

dbrgn: I think so, yes!

dbrgn (approved):

ubruhin:
Nice, thanks @bjaminn and @dbrgn! :smiley:

Issue #830, opened 2021-01-16
Add Gerber X3 export incl. pick&place and netlist data


ubruhin:
##### SUMMARY

With Gerber X3 it's possible to generate standardized pick&place data files, which makes it easier to order assembled PCBs (if the assembly house supports it, of course). Hopefully the non-standardized pick&place file format EDA tools use today can be replaced by Gerber X3 some day.

In addition to pick&place data, Gerber also allows to export the netlist, which some PCB manufacturers support for the electrical test of PCBs. That's already possible with Geber X2 I guess, but LibrePCB does not yet support it.

##### SOLUTION

However we implement it, we still need to keep support for the legacy (before X2) Gerber format since some PCB manufacturers do not accept modern Gerber files...

PR #881, opened 2021-03-14
Refactor and improve gerber export for better compatibility

### LINKED ISSUE #872: Gerber Generation : Provide a way to generate RS274-X gerber files (X1?)
##### SUMMARY

It seems that big PCB manufacturers don't support Gerber X2 files :scream: 

##### SOLUTION

Add a way to generate RS274-X files.

It might be just removing `%TF` lines
(Ucamco reference Gerber viewer accept as Gerber X1 the files generated by LibrePCB where `%TF` lines have been removed manually).

### LINKED ISSUE #873: Gerber Generation : Octagons were not correctly handled by manufacturers

```
LibrePCB Version: 0.1.5
Git Revision:     a435c5095c6f94df4da130fe832ea97f42d17901
Build Date:       2020-10-03T19:48:36
Linking type:     static
Unbundled libs:   -
CPU Architecture: x86_64
Operating System: Ubuntu 20.10
Platform Plugin:  xcb
```

##### SUMMARY

Even if files seems to be correctly formatted (Ucamco reference gerber viewer says :+1: ),
2 big manufacturers (PCBWay or JLCPCB) have difficulties to handle Gerber files generated by LibrePCB.
It seems that octagons are not correctly handled.
It is with octagons with difference between width and height.
Others seems to be correctly handled.

##### STEPS TO REPRODUCE

1. create a component with top octagonal pad of 1.0 x 0.6 mm
2. insert it in your design
3. generate Gerber files
4. test it in [Ucamco reference Gerber viewer](https://gerber-viewer.ucamco.com/) => it's ok
5. send them to a manufacturer (here : PCBWay or JLCPCB) => files refused

##### EXPECTED RESULTS

image

##### ACTUAL RESULTS

image

EQ

ubruhin:
Since several Gerber related issues and improvement requests popped up the last months, I decided to refactor and improve the whole Gerber export implementation. Hopefully now the Gerber files are much more compatible with some old, crappy CAM software.

### FabricationOutputDialog: Always use *.drl suffix for drill files

So far we used the *.drl file extension for drill files when choosing the default settings, but *.txt when using Protel settings. However, it seems that other tools use the *.drl file extension for drills even when using Protel file extensions. And Ucamco highly recommends to use either *.nc or *.drl. Since *.drl seems to be more commonly used, let's use it too when choosing the Protel settings in the fabrication output dialog.

### Export circles as strokes instead of flashes

Ucamco's [PCB Fabrication Data](https://www.ucamco.com/files/downloads/file_en/130/pcb-fabrication-data-a-guide_en.pdf) guide clearly states:

    All pads must be flashes. All flashes must be pads.

Which we didn't follow yet for circles (on any layer). So let's draw circles as strokes and regions instead of flashes.

### Do not use zero-size aperture for regions

Actually G36/G37 regions do not have an aperture attached. But for compatibility reasons, it's better to still select an aperture as usual. We used an aperture of size 0, but this already caused some issues in the past (although not critical) and the Gerber specs recommends to not use zero-size apertures. So let's use an aperture size of 0.01mm (it has no impact on the rendered image anyway).

### Set multi quadrant arc mode (G75) as default

Instead of starting with single quadrant mode (G74) in the file header and then switching to multi quadrant mode (G75) as soon as we create the first arc, let's enable multi quadrant mode already in the header.

### Don't use complex and error prone apertures

Some of our exported Gerber apertures cause troubles with particular CAM software (e.g. CAM350 or Generis2000). This commit replaces such apertures with other, equivalent macros (leading in the same image) which cause less incompatibility issues:

- Don't use arithmetic expressions in aperture macros anymore. These caused issues because some CAM software uses wrong operator precedence.
- Don't use variables (e.g. "$1") in aperture macros anymore. Use only hardcoded macro parameters. Not absolutely sure if these causes issues with CAM software, but it heavily reduces complexity without heavily increasing the Gerber file size for typical PCBs.
- Circles in aperture macros no longer specify the optional rotation parameter since some CAM tools don't render such circles or even fail to parse such files completely. The Gerber specs also recommends to omit this parameter.
- Don't use the "Regular Polygon (P)" aperture primitive anymore. We used it for octagon pads. But it was implemented wrong (specified inner diameter instead of outer diameter) and the Gerber specs also mention exactly this bug existing in some tools. Replaced by outline macros now.

### Generate X1 comments instead of X2 commands

Gerber X2 attributes are not supported by all PCB fabricators, so users have issues with ordering PCBs at such fabricators. By using G04 comments (which exist since Gerber X1) for the X2 attributes, this problem is avoided. This is the way recommended by Ucamco and also implemented by several other EDA tools.

This is how an X2 command looks like:

    %TF.GenerationSoftware,LibrePCB,LibrePCB,0.1.5*%

And this is an X1 comment containing the same information:

    G04 #@! TF.GenerationSoftware,LibrePCB,LibrePCB,0.1.5*

In future, we might allow the user to choose between the X1 compatibility mode and real X2 attributes. But X1 is the safer default since every fabricator can read Gerber X1 files. So let's output X1 by default. Probably most up-to-date Gerber readers can interpret such standardized X1 comments anyway, so maybe we don't even need to provide real X2 output in the near future.

### Set many more X2 attributes

Set some more attributes which seem to be quite useful for fabricators or for reviewing the Gerber files in a viewer which support these attributes:

- ".FileFunction" to indicate what functionality the file has. This improves layer auto-detection (silkscreen, copper, ...) in some Gerber viewers and probably also in some CAM software.
- ".FilePolarity" to indicate what polarity the file has. Just to make clear whether to add or remove material.
- ".SameCoordinates" to indicate that all files use the same coordinates. Avoids issues with offset mismatch etc.
- ".AperFunction" to clearly state whether a flashed object is a THT pad, SMT pad or a via. Helps fabricators to understand the function of flashed objects.
- ".N" to specify the net name of each copper object. Helps to automate electrical test of PCBs (e.g. testing for short circuits etc.) and to check for design rule violations (e.g. clearances).
- ".P" to specify pin numbers of flashed pads. Especially useful to know which of the pads is pad 1 (e.g. for assembly).
- ".C" to specify the component designator (e.g. "R5") of elements. Helps to understand which objects are related to each other (e.g. all silkscreen lines of a particular component).

Maybe we should make the latter three attributes optional (user configurable) some day since one might have concerns about intellectual property of a design. But the electrical connections are (obviously) visible in the copper layers anyway, so IMHO it's not really a big issue if net names are included too. Or are there other opinions? :)

### Add lots of new unit tests

To avoid regressions once we modify the Gerber generator classes in future. In addition to tests checking the correctness of the generated Gerber output, there are many tests verifying we do not use Gerber features which caused issues in the past.

Fixes #872.
Fixes #873.
Implements parts of #830.


dbrgn:
I did not look at the code, but the description sounds great!

> Maybe we should make the latter three attributes optional (user configurable) some day since one might have concerns about intellectual property of a design. But the electrical connections are (obviously) visible in the copper layers anyway, so IMHO it's not really a big issue if net names are included too. Or are there other opinions? :)

As long as nobody complains, I'd keep it that way.

ubruhin:
@atommann could you maybe check if the following Gerber files work correctly in your CAM software? Just to be sure these changes do not introduce new problems :wink:

- Exported with default configuration: [default.zip](https://github.com/LibrePCB/LibrePCB/files/6217332/default.zip)
- Exported with Protel configuration: [protel.zip](https://github.com/LibrePCB/LibrePCB/files/6217335/protel.zip)

atommann:
> 
> 
> @atommann could you maybe check if the following Gerber files work correctly in your CAM software? Just to be sure these changes do not introduce new problems wink
> 
>     * Exported with default configuration: [default.zip](https://github.com/LibrePCB/LibrePCB/files/6217332/default.zip)
> 
>     * Exported with Protel configuration: [protel.zip](https://github.com/LibrePCB/LibrePCB/files/6217335/protel.zip)

Both work correctly in the CAM software.

But I have a quick question, file gerbertest_COPPER-BOTTOM.gbr, start from line 335, it defines the shield-shaped object. My doubt is: Why it's not in a G36/G37 pair and Gerbv still show it correctly?

```
D34*
X130175000Y34925000D02*
G02*
X130175000Y29845000I-2540000J-2540000D01*
G01*
X135255000Y29845000D01*
G03*
X135255000Y34925000I0J2540000D01*
G01*
X130175000Y34925000D01*
```

ubruhin:
> Both work correctly in the CAM software.

Perfect :+1:

> But I have a quick question, file gerbertest_COPPER-BOTTOM.gbr, start from line 335, it defines the shield-shaped object. My doubt is: Why it's not in a G36/G37 pair and Gerbv still show it correctly?

These lines are not the region to be filled, but only an outline of zero size. Each polygon consists of an outline, and optionally by a fill. The outline is always exported as strokes, and the fill is only exported as regions if it needs to be filled. Since the outline of this polygon has a width of 0mm, it is basically invisible in the Gerber viewer :wink:

Actually this is a good point - should we maybe better not exporting polygon outlines at all if its width is zero and its area is filled? :thinking: Probably it doesn't really matter...

ubruhin:
One more change, suggested by Ucamco:

### ExcellonGenerator: Set Gerber file/apterure attributes

Although Excellon files are not Gerber files, there is a standard [1]
describing how Gerber X2 attributes can also be used in Excellon files
to add metadata in a standardized way. This might help some fabricators
to automatically detect whether drills are plated or not, and between
which layers the drills need to be created.

[1] https://www.ucamco.com/files/downloads/file_en/305/the-xnc-file-format-specification-technical-manual_en.pdf

PR #899, opened 2021-06-20
Library editor: Ignore delete command if no items selected

ubruhin:
Currently, the "delete" action has shown a message box to confirm deleting the selected items, even though there were no items selected (i.e. no library elements will be deleted). Now this message box is only shown if there are really some items selected.

In addition, I adjusted the corresponding functional test to explicitly select the items to delete. This is the annoying flaky test on CI, let's hope this change makes it more reliable...


ubruhin:
> In addition, I adjusted the corresponding functional test to explicitly select the items to delete. This is the annoying flaky test on CI, let's hope this change makes it more reliable...

Didn't fix the issue :sob: I have now marked the flaky test with xfail on Windows since I don't think there is anything wrong with LibrePCB but this flaky test already caused lots of frustration. It's time to finally get rid of this #*!@!?$ CI failure.

PR #907, opened 2021-08-28
DRC: Support partial checks & quick re-run from dock widget

### LINKED ISSUE #869: Board Editor: Allow partial DRC checking
##### SUMMARY

DRC checking is a bit long as it checks a lot of things.
(on my current board, more than 40s with a recent laptop)
It will be great to check only what the user needs at the current moment to speed up the process.

##### SOLUTION

On the left of each line of rules, add a checkbox to validate/invalidate this rule checking.
Add also "activate all rules" and "deactivate all rules" buttons.

ubruhin:
# DRC Dialog

### Save/restore window position/size

Restore the last dialog window position and size instead of using defaults every time opening the dialog.

### Simplify/cleanup user interface

- Remove status list widget on the left side of the dialog
- Display status messages as tooltip of progress bar (mouse hover)
- Display last status message as text of progress bar
- Hide progress bar until the DRC is started
- Move run button into button box (next to the close button)
- Remove yellow background of hint text, use italic font instead

### Add checkboxes to select checks to execute

By default, all checks are enabled. But now the user can disable specific checks to focus on specific things. For example, it might be useful to disable the check for missing connections if the layout is not finished yet. Or the rebuild of planes can be disabled since this can consume quite some time, even though the user might know they are already up to date.

|   |   |
| --- | --- |
| **Old:** | librepcb-drc-dialog-old |
| **New:** | librepcb-drc-dialog-new |

# DRC Dock Widget

### Add buttons to DRC dock widget to open settings or quickly re-run DRC

When fixing DRC messages, currently one had to open the DRC dialog, run the DRC, and close the dialog every time one wanted to know if the messages are fixed. Now this is much easier: Just one click on the new button in the dock widget is needed to run the DRC again (with the same settings as used in the previous run). In addition, there's also a button to open the DRC dialog (e.g. if settings need to be changed).

librepcb-drc-dock

Fixes #869

### DIFF
--- a/libs/librepcb/project/boards/drc/boarddesignrulecheck.cpp
+++ b/libs/librepcb/project/boards/drc/boarddesignrulecheck.cpp
@@ -61,7 +61,11 @@ namespace project {
 
 BoardDesignRuleCheck::BoardDesignRuleCheck(Board& board, const Options& options,
                                            QObject* parent) noexcept
-  : QObject(parent), mBoard(board), mOptions(options), mMessages() {
+  : QObject(parent),
+    mBoard(board),
+    mOptions(options),
+    mProgressStatus(),
+    mMessages() {
 }
 
 BoardDesignRuleCheck::~BoardDesignRuleCheck() noexcept {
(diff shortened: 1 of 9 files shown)


dbrgn:
Very cool!

PR #914, opened 2021-09-16
Make format_code.sh also format CMake and *.qrc files

### LINKED ISSUE #908: Re-add CMake compatible sort_qmake_file_entries.py script
As mentioned in https://github.com/LibrePCB/LibrePCB/pull/798#pullrequestreview-684494480, once the CMake PR (#798) branch is merged, we should re-add the `sort_qmake_file_entries.py` script.

Comment from @ubruhin:

> Btw, I just realized that with removing `sort_qmake_file_entries.py` the *.qrc files are now no longer sorted as well. So it would be really nice to re-add it - I gave it a try to even sort filenames in `CMakeLists.txt` as well, feel free to copy it from [here](https://gist.github.com/ubruhin/109312152d2298b8adeb7fec7f1ea3d5) :wink: But I could also add it later with a separate PR if you like. My script is for sure not very elegant, but so far it seems to work fine :slightly_smiling_face: 

Blocked by #798.

ubruhin:
Using the [`cmakelang`](https://cmake-format.readthedocs.io/en/latest/index.html) tool to automatically format all CMake files. It's not perfect, but IMHO it's good enough to ensure consistent formatting of all our CMake files. The most important feature is that it sorts all sources passed to `add_executable()` and `add_library()` - this reduces the risk of merge conflicts and makes it easier to add new sources without worrying about formatting and sorting (especially since qtcreator is *not* able to add the sources automatically).

I also removed the "Headers", "Sources" and "Forms" sections since these are actually not needed anymore (previously with qmake, they were needed). So this:

```cmake
add_library(
  ...
  # Sources
  widgets/centeredcheckbox.cpp
  widgets/doublespinbox.cpp

  # Headers
  widgets/centeredcheckbox.h
  widgets/doublespinbox.h

  # Forms
  widgets/centeredcheckbox.ui
  widgets/doublespinbox.ui
)
```

is now converted to this:

```cmake
add_library(
  ...
  widgets/centeredcheckbox.cpp
  widgets/centeredcheckbox.h
  widgets/centeredcheckbox.ui
  widgets/doublespinbox.cpp
  widgets/doublespinbox.h
  widgets/doublespinbox.ui
)
```

which even leads to cleaner diffs when adding new files, since all related files are next to each other.

Any opinion about these changes @dbrgn? Would be great if you could review this PR :slightly_smiling_face: 

I also re-added the ability to format *.qrc files, but this time using `xmlsort` directly within `format_code.sh` instead of a dedicated Python script.

Closes #908.


.cmake-format.py
```diff
@@ -0,0 +1,257 @@
...
+with section("format"):
+
+  # Disable formatting entirely, making cmake-format a no-op
+  disable = False
+
+  # How wide to allow formatted cmake files
+  line_width = 80
+
+  # How many spaces to tab for indent
+  tab_size = 2
```
dbrgn, line 44: If you want to change the indentation, you should update [.editorconfig](https://github.com/LibrePCB/LibrePCB/blob/dc0e9c0a6efb668ce36430fd03a9fd824f68da8d/.editorconfig#L13) as well.

.cmake-format.py
```diff
@@ -0,0 +1,257 @@
...
+# -----------------------------
+# Options affecting formatting.
+# -----------------------------
+with section("format"):
+
+  # Disable formatting entirely, making cmake-format a no-op
+  disable = False
+
+  # How wide to allow formatted cmake files
+  line_width = 80
```
dbrgn, line 41: In my opinion, in general a line width of 100 looks better in many cases (especially because wrapping heuristics need to be used less often), but I don't have a strong opinion.

CMakeLists.txt
```diff
@@ -137,36 +163,43 @@ add_subdirectory(apps/EagleImport)
...
             REGEX "/\.(git|gitattributes|gitignore)$" EXCLUDE
-        )
-    endif()
+    )
+  endif()
 endforeach()
-install(DIRECTORY ${CMAKE_BINARY_DIR}/i18n DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/librepcb)
+install(DIRECTORY ${CMAKE_BINARY_DIR}/i18n DESTINATION
+                  ${CMAKE_INSTALL_DATAROOTDIR}/librepcb
+)
```
dbrgn, line 205: Hm, this one's weird. Shouldn't `DESTINATION` be on the second line as well? Maybe the formatting tool doesn't understand the structure of the install command.

apps/librepcb-cli/CMakeLists.txt
```diff
@@ -4,59 +4,64 @@ set(CMAKE_AUTOUIC OFF)
 set(CMAKE_AUTORCC OFF)
 
 # Executable
-add_executable(librepcb_cli
-    # When building on macOS, create a bundle
```
dbrgn, line 8: This comment got removed. But I assume it should be self-explanatory.

dbrgn (commented):
In quite some cases I think the CMake code now looks worse than before, but that's always the case with autoformatters 🤷🏻‍♂️ I left a few comments.

.cmake-format.py
ubruhin: Generally I agree, but since we already use a width of 80 in all the hundrets of other files, I like to use the same width also for the project files.

.cmake-format.py
ubruhin: Good point, thanks! -> Updated `.editorconfig`.

CMakeLists.txt
ubruhin: Yeah the install command indeed doesn't work very good, thus I have overridden the install keyword with a custom configuration in `.cmake-format.py` :see_no_evil: I tweaked the config a bit, now it looks better.

apps/librepcb-cli/CMakeLists.txt
ubruhin: Re-added the comment (it's now on the same line as `MACOSX_BUNDLE` and the target name, but I'd say that's OK).

ubruhin (commented):
Thanks for the review! Updated the PR accordingly.

> In quite some cases I think the CMake code now looks worse than before, but that's always the case with autoformatters 🤷🏻‍♂️

Yes, I totally agree :slightly_smiling_face: However, IMO the results of `cmake-format` are good enough that the advantages outweigh the disadvantages :wink:

.cmake-format.py
dbrgn: I don't think our source code files need to match the build system files, but it's your choice.

CMakeLists.txt
dbrgn: Much better 🙂

CMakeLists.txt
```diff
@@ -137,36 +163,45 @@ add_subdirectory(apps/EagleImport)
...
+      DIRECTORY ${d}
+      DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}
+      REGEX "/\.(git|gitattributes|gitignore)$" EXCLUDE
+    )
+  endif()
 endforeach()
-install(DIRECTORY ${CMAKE_BINARY_DIR}/i18n DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/librepcb)
+install(DIRECTORY ${CMAKE_BINARY_DIR}/i18n
+        DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/librepcb
+)
```
dbrgn, line 207: One last thing: Why is the format of this `install` different than the one above (inside the `if`)? (In my opinion the version above looks better, but I don't have a strong opinion on that.)

dbrgn (approved):

CMakeLists.txt
ubruhin: Hmm I don't know exactly - somehow this can be controlled with the configuration parameters `min_prefix_chars` and `max_prefix_chars` but I still don't understand what they do exactly so I left them at their default value :see_no_evil: 

However, we can enforce wrapping the `install` command by adding it to the `always_wrap` configuration :slightly_smiling_face:  This fixes the inconsistency here and does not change any other file. I'll update the branch accordingly.

ubruhin:
@dbrgn I updated the branch to fix the inconsistent wrapping of `install` commands. Only one `install` command was affected, everything else is not touched by this update.

I'll merge as soon as CI succeeds.

PR #917, opened 2021-09-22
Add DXF import to package-, symbol- and board editors

### LINKED ISSUE #891: Board Editor: Add DXF import
Really Great software!

I though when checking it...the most important is here! But then I could not find any way of importing a PCB shape with reference holes via a DXF file etc. Maybe some other way? Through a picture, PDF, or anything that has the exact sizes to begin the board outline and mounting holes etc.

Nowadays, a design begins at the 3D level after hand and mind sketches, designs are no longer based around the PCB. And PCB sizes needs to be imported from that 3D design somehow. The days of square PCB's are long gone.

Jabba

ubruhin:
## Summary

This adds a DXF import to the package editor, symbol editor and board editor. This is especially useful to import board outlines or package dimensions from a mechanical CAD drawing, to avoid the need to manually draw these polygons.

(Although the symbol editor is not really something mechanical, it might still be useful to import a logo or so.)

Closes #891

## Dialog

The dialog allows to configure how to import the DXF:

- Layer of the imported lines
- Optionally import circles as drills instead of polylines (e.g. for board mounting holes)
- The line width to be used for the imported objects. The layer thickness as defined in the DXF is ignored (I'm not sure if this would even be accessible, but usually the MCAD line width is not directly usable for EDA anyway)
- Additional scale factor, in case the scaling does not work as expected. The importer tries to scale the objects according the measure unit as defined in the DXF, but DXF files can be unitless - for such DXF files, LibrePCB doesn't know how to scale the contained numbers (it will assume millimeters in that case).
- Positioning mode: Either interactive (position defined with mouse click) or by manually specifying coordinates.

image

## Workflow

1. Choose file
2. Choose options
3. Interactive placement (optional)

librepcb-dxf-import

## Limitations

The imported data is not postprocessed in any way - if the DXF is crap, the imported objects will be crap too. For example, if the imported board outline consists of multiple line segments instead of a single, closed polyline, LibrePCB will not convert it to a closed polyline so it might lead to issues when exporting such a file to Gerber. Currently such things should be polished in a MCAD first, before importing the DXF into LibrePCB.

## Implementation

I added the third party library [dxflib](https://www.qcad.org/en/90-dxflib) for parsing the DXF files. This library seems to be the only reasonable choice. It is even available in the package manager of various Linux distributions.

There is also another library [libdxfrw](https://github.com/codelibs/libdxfrw) which looks more feature rich (also supports reading DWG files), but also much more complex. In addition, it does not seem to be available in package managers. So for now, I decided to go with dxflib. AFAIK KiCad and Horizon both use dxflib as well.

The dxflib is added as a Git submodule from the repository here: https://github.com/LibrePCB/dxflib. Patches for LibrePCB are added in https://github.com/LibrePCB/dxflib/pull/1.

I added dxflib to our CMake project files, but it can be unbundled with `UNBUNDLE_DXFLIB`. CI even verifies that unbundling works. The required Docker image patch is added in https://github.com/LibrePCB/docker-librepcb-dev/pull/13.

@dbrgn It would be great if you could take a look at the CMake files, since you are now our CMake expert :smiley: I added the CMake config only with my copy&paste skills, not with CMake skills :see_no_evil:


dbrgn:
Awesome! 🎉

The CMake config looks good to me. I tested the DXF import with an Arduino shield file and that worked fine.

I wonder if it would be possible to detect polylines, and import them as such?

ubruhin:
> The CMake config looks good to me. I tested the DXF import with an Arduino shield file and that worked fine.

Perfect, thanks for reviewing and testing! :tada:

> I wonder if it would be possible to detect polylines, and import them as such?

Probably that would be possible, but I prefer to postpone that improvement to some day later to get the 0.1.6 release done sooner :see_no_evil: If you have a DXF with multiple lines instead of a single polyline, one can already fix that in a MCAD like LibreCAD as a workaround. If the DXF then contains a proper polyline, LibrePCB will import is as a single polygon.

I mean, basically it makes more sense anyway to fix ugly DXF files in a MCAD. An "auto-polishing" feature in LibrePCB would be just "nice to have" to avoid needing to manually fix such DXF files.

dbrgn:
> Probably that would be possible, but I prefer to postpone that improvement to some day later to get the 0.1.6 release done sooner 🙈 

Yes, that makes sense.

> If you have a DXF with multiple lines instead of a single polyline, one can already fix that in a MCAD like LibreCAD as a workaround. If the DXF then contains a proper polyline, LibrePCB will import is as a single polygon.

Ah, I see!

dbrgn (approved):

Issue #852, opened 2021-02-11
OpenGL rendering changes Schematic Editor Background to black


deanusjack:
##### VERSION / OS / ENVIRONMENT
LibrePCB Version: 0.1.5
Git Revision:     a435c5095c6f94df4da130fe832ea97f42d17901
Build Date:       2020-10-03T20:06:07
Linking type:     static
Unbundled libs:   -
Qt Version:       5.15.0 (built against 5.15.0)
CPU Architecture: x86_64
Operating System: Windows 10 Version 1909
Platform Plugin:  windows

##### SUMMARY
Enabling OpenGL Hardware Acceleration causes the Schematic Editor background to change from White to Black, making it hard to read.

##### STEPS TO REPRODUCE
Workspace Settings -> Appearance -> Enable OpenGL Hardware Acceleration -->Open project->  Go to Schematic -> Background is now black.
Close Project -> Workspace Settings -> Appearance -> Disable OpenGL Hardware Acceleration -->Open project->  Go to Schematic -> Background  is white as normal

##### EXPECTED RESULTS
Background of Schematic editor not to change colour.

##### ACTUAL RESULTS
Background of Schematic editor changes colour making it hard to see/read due to contrast between red/black,

##### ADDITIONAL INFORMATION
Fresh Installation of LibrePCB.

PR #921, opened 2021-10-04
issue #852 - OpenGL rendering changes Schematic Editor Background to black

masondb:
backgroundBrush() would return "no brush" when switching to opengl for rendering for graphicsview. For boardeditor  setbackgroundbrush is used in the constructor for BoardEditor class but not for SchematicEditor class. Simply added white brush to SchematicEditor constructor to set background

### DIFF
--- a/libs/librepcb/projecteditor/schematiceditor/schematiceditor.cpp
+++ b/libs/librepcb/projecteditor/schematiceditor/schematiceditor.cpp
@@ -109,6 +109,7 @@ SchematicEditor::SchematicEditor(ProjectEditor& projectEditor, Project& project)
   mGraphicsView = new GraphicsView(nullptr, this);
   mGraphicsView->setUseOpenGl(
       mProjectEditor.getWorkspace().getSettings().useOpenGl.get());
+  mGraphicsView->setBackgroundBrush(Qt::white);
   setCentralWidget(mGraphicsView);
 
   // Add actions to toggle visibility of dock widgets


ubruhin:
Thanks for the contribution! :tada: 

My thoughts:
- I think it makes more sense to move the change into the `GraphicsView` constructor. Then the issue will automatically be fixed for *all* places where `GraphicsView` is used (`SymbolEditorWidget`, and several dialogs) and we won't have the same issue again if `GraphicsView` is added to more windows in future.
- Since the board editor sets the *foreground* brush as well, I'd prefer to also set it explicitly to `Qt::black` by default, just next to the background brush.
- The commit message does not follow the [conventions](https://chris.beams.io/posts/git-commit/) we specify in [CONTRIBUTING.md](https://github.com/LibrePCB/LibrePCB/blob/master/CONTRIBUTING.md). However, it's not a big deal, I can specify a different commit message when squash&merge this PR :wink: Btw, the issue number does not need to be part of the commit message, we usually only specify it in the PR description.

Could you maybe update the PR accordingly, at least with the first two suggestions? Or what do you think about these suggestions?

And FYI, if you push any update to this branch, [we prefer squashed&force-pushed commits](https://developers.librepcb.org/df/d30/doc_developers.html#doc_developers_pullrequests) instead of adding additional commits. But in case you aren't familiar with this workflow, just do whatever is the easiest for you :slightly_smiling_face:

PR #936, opened 2021-11-24
Integrate EAGLE library import into library editor

### LINKED ISSUE #224: Eagle import: UUID list file sensible defaults
The eagle importer needs a (temporary? - please clarify) UUID list file.
If no such file is given, library creation does seem to succeed, however the referencing UUIDs are generated new for every Symbol/Package/Device step, which causes problems later on.
I created a new empty file for this, and it seemed to work. I am not sure however if this is the intended way to use it (please clarify).
I suggest to do one of the following to make it more clear to users:

- Disable the convert buttons until a valid UUID list file is selected
- use a default filename somewhere safe
- use a default hidden file if none is given

Also, there should be some sort of tooltip which explains what should be selected here (either a specific file or a random new file which can be deleted afterwards).

ubruhin:
Actually I just wanted to do some code reorganization/refactoring, but then struggled with the EAGLE import code which was a hacky standalone tool until now - either I had to refactor it as well (which makes no sense since the tool was not used anymore) or I had to delete it (which I don't like since it might still be useful some day). So I decided to refactor it and then integrate it into the library editor to make it more useful again :see_no_evil: 

Unfortunately it was *much* more effort needed to get it working really well, with an intuitive user interface. But now I think it is ready to make it available for end users. Maybe it is helpful for users considering the migration from EAGLE to LibrePCB since now they can easily import their libraries, thus reducing the migration effort.

It is for sure not perfect - there are several little things which do not look 100% identical to EAGLE. And the quality of the imported libraries will never be as high as native LibrePCB libraries since the library concept of EAGLE is different to LibrePCB. But I think even lower quality libraries are quite useful, especially if they can be created with almost no effort.

Supported are EAGLE 6 or later *.lbr files (XML file format).

Here's how it works:
librepcb-eagle-library-import

The standalone EAGLE library import application is now removed from the project since it makes no sense anymore.

Closes #224.


dbrgn:
Nice, some users will be really happy with this 🙂

PR #937, opened 2021-12-03
CI: Update macOS to 10.15 and fix+test unbundling muparser

ubruhin:
- Update macOS job to 10.15 since 10.14 [is deprecated and will be removed soon](https://devblogs.microsoft.com/devops/hosted-pipelines-image-deprecation/).
- Update Ubuntu 20.04 docker images to new version which contains libmuparser
- Fix wrong include path to `muParser.h`
- Test unbundling muParser on CI

@dbrgn If I remember correctly, the wrong include path to `muParser.h` lead to issues with the AUR package, right? Then I suspect this change will fix it for the next release...


dbrgn:
> @dbrgn If I remember correctly, the wrong include path to `muParser.h` lead to issues with the AUR package, right? Then I suspect this change will fix it for the next release...

Correct, previously we [manually patched the include path](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=librepcb#n53).

dbrgn:
Will this work when not unbundling as well?

dbrgn (approved):

ubruhin:
> Correct, previously we manually patched the include path.

:+1:

> Will this work when not unbundling as well? Seems like it, considering the CI passes...

Yes, the directory `libs/muparser/include` is then passed as include path to the compiler.

PR #939, opened 2021-12-06
Refactor organization of sources, static libraries and namespaces

ubruhin:
## Motivation

This is a major refactoring of the directory- and namespace structure of all our code. The main motivation is to start separating logic from GUI stuff since currently these things are mixed together (see #67). One negative side effect of this architecture issue is that our CLI depends on the existence of a window manager (i.e. does not run without an x-server), but that's only one of the issues we have with the current architecture.

The architecture issue is not entirely fixed with this PR, but it is an important step into that direction.

## Old Architecture

The current static library organization looks like that:

- `libs/librepcb/common`: `namespace librepcb`
- `libs/librepcb/library`: `namespace librepcb::library`
- `libs/librepcb/project`: `namespace librepcb::project`
- `libs/librepcb/workspace`: `namespace librepcb::workspace`
- `libs/librepcb/libraryeditor`: `namespace librepcb::library::editor`
- `libs/librepcb/librarymanager`: `namespace librepcb::library::manager`
- `libs/librepcb/projecteditor`: `namespace librepcb::project::editor`

image

Where *any* of these libraries contain a lot of GUI stuff (graphics items, widgets, undo commands, ...).

## New Architecture

The static libraries are now organized completely different, basically only 2 relevant libraries and corresponding namespaces exist now:

- `libs/librepcb/core`: `namespace librepcb`
- `libs/librepcb/editor`: `namespace librepcb::editor`

The former now contains *all* the logic which was previously spread across `libs/librepcb/common`, `libs/librepcb/library`, `libs/librepcb/project` and `libs/librepcb/workspace`. The namespaces `librepcb::library`, `librepcb::project` and `librepcb::workspace` do not exist anymore, all the related classes are now in the namespace `librepcb`.

The latter one is basically the result of merging the libraries `libs/librepcb/libraryeditor`, `libs/librepcb/librarymanager` and `libs/librepcb/projecteditor`. In addition, all GUI related classes from `libs/librepcb/common`, `libs/librepcb/library`, `libs/librepcb/project` and `libs/librepcb/workspace` are moved into that new library. All the namespaces `librepcb::library::editor`, `librepcb::project::editor` and `librepcb::workspace::editor` are now merged into a single namespace `librepcb::editor`.

image

## Advantages

- Once the `librepcb/core` library is completely free of GUI stuff, the CLI will no longer require a window manager to work (not possible yet, since more code refactoring is needed to completely get rid of the GUI dependency).
- Once we provide official Python bindings for LibrePCB, it will be based on `librepcb/core` as well and thus will be completely free of GUI stuff as well.
- Less (nested) namespaces -> less complexity in general. Previously, GUI related classes were spread accross various namespaces (`librepcb::library|project|workspace::editor|manager`) although some of them are tightly related together.
- Undo commands were not implemented where they were actually needed. For example, `CmdPolygonEdit` was implemented in `librepcb/common` because `Polygon` was there too. But in that library, we actually don't know which undo commands will be needed in the GUI editors. For this reason, there were even undo commands implemented which were not used at all. Now all undo commands are in `librepcb/editor`, so they are located right where they are actually needed.

## Concerns

Although I'm pretty sure this goes into the right direction, I still have some concerns if everything makes totally sense.

- Generally it feels very wrong to *merge* several small static libraries and namespaces into larger libraries and more cluttered namespaces (-> higher risk of name clashes etc.) - As the project is getting larger and larger, I'd expect a refactoring should *split* large libraries into smaller ones :see_no_evil: 
- Due to a single core library, dependency violations between the `common`, `library`, `project` and `workspace` classes are no longer rejected. The dependencies between these classes should only be in one direction (`workspace -> project -> library -> common`). Previously you got a compiler error if you violated this architecture (because these were separate static libraries). Now all these classes are in the `core` library, so no compile error anymore if the architecture is violated. However, maybe it's not *that* a big issue even if some small architecture violation occurs...
- It seems that having larger static libraries leads to longer linking time than several smaller libraries, although the difference is not very big. Previously a source file modification took ~16s until all libraries and applications were linked together (decent computer, but debug build). Now it takes ~20s. It's a bit cumbersome for development, but I think ~4s shouldn't be a reason to decide against an architectural refactoring...

## Feedback

A second opinion would be very welcome, maybe from @rnestler or @dbrgn? :slightly_smiling_face: Unfortunately probably it's not easy to understand this refactoring in detail. Maybe the file listing and namespace listing of Doxygen could help a bit to get a better picture what has actually changed:
- Directory structure: [old](https://developers.librepcb.org/files.html) - [new](https://developers.librepcb.org/_branches/refactor-source-library-organization/files.html)
- Namespace structure: [old](https://developers.librepcb.org/annotated.html) - [new](https://developers.librepcb.org/_branches/refactor-source-library-organization/annotated.html)


ubruhin:
Moved some more files and improved the architecture overview diagram to represent the layers `library`, `project` and `workspace` and their dependency order within the `core` library.

So far I'm quite happy with this new architecture. And no feedback = no objections, so I think I'll merge this soon :wink:

dbrgn:
The proposal looks sensible! It makes sense to move all core business logic into a library that's completely independent of the UI. This way, functionality like library management could be exposed to other tools (e.g. the library generator).

The only concern that came to mind for me was this one:

> Due to a single core library, dependency violations between the `common`, `library`, `project` and `workspace` classes are no longer rejected. The dependencies between these classes should only be in one direction (`workspace -> project -> library -> common`). Previously you got a compiler error if you violated this architecture (because these were separate static libraries). Now all these classes are in the `core` library, so no compile error anymore if the architecture is violated. However, maybe it's not _that_ a big issue even if some small architecture violation occurs...

In Java, there's a tool called [ArchUnit](https://www.archunit.org/), which allows testing certain architecture invariants, e.g. that files in one directory are only accessed from files in another directory, but not the other way around. Maybe there's something similar in C++?

ubruhin:
Thanks for the feedback! :tada: 

> In Java, there's a tool called ArchUnit, which allows testing certain architecture invariants, e.g. that files in one directory are only accessed from files in another directory, but not the other way around.

Ah, that's an interesting tool, probably that would indeed be quite useful. But unfortunately so far I didn't find something similar for C++ :disappointed:

PR #947, opened 2022-01-13
Implement new, more powerful print / image export feature

ubruhin:
## Problem

Our current print and PDF-export features rely on `QGraphicsView` and `QGraphicsScene`, which basically just sends the screen content (with different resolution) to the printer or a PDF file. This make the whole thing very simple (trivial to implement), but has some serious issues:

- The PDF export should be part of the `core` (non-GUI) library since it is also needed by the CLI. But `QGraphicsView` must be in the `editor` (GUI) library since it is a `QWidget` and requires a display. This is a fundamental problem, see #67 and #939 for details.
- We have very little control about what content is actually sent to the printer or PDF, basically it's just the screen content. This might be suitable for very simple tasks, but are not enough for more advanced tasks.

## New Concept

This PR completely decouples the print and other graphics export features from the rendering on screen. No `QGraphicsView` or `QGraphicsScene` is needed anymore, everything is drawn directly to the output device/file using a `QPainter`. This gives us the flexibility to make the output configurable as we need it, and it works without a display (-> useful for CLI).

The now completely new print- and graphics export feature provides a lot of new features:

- Provide new configuration options for printing, PDF and SVG export:
  - Page size (was hardcoded to A4 until now)
  - Page orientation (was hardcoded to landscape until now)
  - Page margins
  - Scale factor (to allow exact 1:1 output instead of "fit to page")
  - Rotation
  - Mirroring (mainly useful for boards)
  - Minimum line width (needed to keep all lines visible, even very thin ones)
  - Optionally use only black/white instead of layer colors
  - Optionally use black background (not allowed for printing, but makes board PDF/SVG looking better with default layer colors)
  - For boards, the exported layers are now configurable instead of using the layers currently visible in the board editor (i.e. the export is now independent of the board editor state)
  - All layer colors are now configurable as well, and by default the board layers are optimized for white background
- Support several pixmap output formats (PNG, BMP, ...) in addition to PDF and SVG, with configurable DPI
- Support copying images to clipboard instead of writing to file
- Support exporting multiple different board views into one PDF (e.g. assembly top + assembly bottom) :muscle: 
- Board stroke texts in PDF files are overlayed with transparent TTF text, so these texts can now be found with the search function of PDF viewers
- Completely new dialog window for all export variants: Printing, PDF export, image export, copy to clipboard
- The library editor now also supports exporting/printing of symbols and footprints.
- Live preview of the exported pages, updated in a worker thread (GUI remains responsive) when modifying configuration! :rocket: 
- Progress bar during the (asynchronous) export, instead of blocking the UI thread
- Configuration is memorized in the user settings (`~/.config/LibrePCB`) to avoid needing to re-enter them each time
- CLI feature `--export-schematics` now supports SVG and several pixmap file formats in addition to PDF
- CLI feature `--export-schematics` now creates PDF files with a custom page size which corresponds to the schematic size plus 10mm margins. Previously it used A4 landscape (hardcoded). I think it shouldn't be a problem if the PDF does not use a standardized paper format, PDF viewers usually provide options how to scale/fit the pages into a standard paper size when sending it to a printer.

### Print

image

image

image

image

## Image Export

image

### Board Layer Selection

image

### Stroke Text In Exported PDF

image

### Workflow (slightly outdated)

librepcb-graphics-export-dialog

## TODO

- [x] Integrate this new dialog into the library editor as well to make it possible to print or export symbols and footprints
- [x] Finalize concept of layer selection in board export - maybe make colors configurable?
- [x] Code cleanup, unit tests, manual tests, bugfixing, ...


dbrgn:
Awesome! I especially like this one:

> Board stroke texts in PDF files are overlayed with transparent TTF text, so these texts can now be found with the search function of PDF viewers

ubruhin:
Finally all TODOs should be resolved now :rocket: Layer colors and board output pages are now fully configurable as well, the library editor now also provides this dialog to print/export symbols and footprints, and many tests have been implemented. Updated the description and screenshots in the post above.

This was a *lot* of work... But now we have the world's best print/export dialog! :muscle:

PR #964, opened 2022-03-26
Save only modified workspace settings to file, keeping unknown settings

ubruhin:
### Problem

Until now, the workspace settings file (`v0.1/settings.lp`) has been considered as stable, i.e. identical for all v0.1.x application versions - just like we keep the file format stable for library- and project files. But unfortunately this doesn't allow us to implement any additional application settings without introducing a new file format version - although probably most users don't put this file under version control (in contrast to libraries and projects).

### Idea

So my idea was to allow file format changes for that `settings.lp` file, as long as it is compatible between all applications of the same major version. However, currently we always build the whole file content from scratch every time it is saved - this means that unknown file entries will be lost (e.g. after downgrading to an older application version, which has less settings). Probably it's not a big deal, but somehow I still don't like the fact that settings could be lost in this case.

### Workspace Settings Modifications

Thus I refactored the workspace settings load/save mechanism to overwrite only *modified* settings. This way, unknown settings file entries are not lost when saving the settings to disk. In addition, this means that the application will always work with the latest default values unless the user explicitly modified the settings - or in other words, most users will now automatically profit from improved default values with new LibrePCB releases. Currently this is not the case, default values are loaded only when opening the application the first time, and then all these values are saved and then restored the next time, so even when we change default values, users will continue using the *old* default values which were valid at the time they started LibrePCB the first time.

One drawback is that with keeping unknown entries, it's not so clear anymore which order the file entries shall have (the order is not relevant for function, just for having a well defined, stable file format). Thus I decided to make the file simply sorted alphabetically, which will lead to a reformatting of the settings file when upgrading from LibrePCB 0.1.6 to 0.1.7. Hopefully nobody cares about this :wink:

### S-Expression File Format Changes

To support keeping unknown file entries, I had to adjust the S-Expression parser to preserve linebreaks. Until now they were skipped (i.e. the formatting was lost) because we didn't need the formatting information.

During this adjustment, I realized that currently linebreaks are sometimes not optimal for version control. Consider this file content, which represents a list of items:

```scheme
 (library_norm_order
  (norm "IEC 60617")
 )
```

If the list is empty, currently the linebreak is omitted:

```scheme
 (library_norm_order)
```

It looks nice, but it leads to less clean diffs when adding/removing entries. In addition, some logic is needed to conditionally add the trailing linebreak or not. IMHO it would be cleaner to always add a linebreak for such elements, i.e.:

```scheme
 (library_norm_order
 )
```

So for file format v0.2 I implemented it this way now (no change for v0.1.x to keep the file format stable).

Another ugly thing is that some files currently contain even empty lines. For example `circuit.lp` looks like that:

```scheme
(librepcb_circuit

 (netclass ada6400e-cb5c-41b6-9a4e-e2f39ffa3383 (name "default"))

 (net 0654411b-a090-4025-a026-4c61e686662e (auto false) (name "LED_TX")
  (netclass ada6400e-cb5c-41b6-9a4e-e2f39ffa3383)
 )

 (component 04187e1b-d8f2-490f-8731-96198fa4fa7e
  (lib_component b91cf23a-4f07-4b99-8f52-0b42304aef20)
  (lib_variant 0eebde17-6d45-4f73-95fe-3078a0f68ab6)
  (lib_device none)
  (name "LOGO1") (value "Made With\nLibrePCB")
 )

)
```

Especially the last empty line is ugly IMHO. Thus for file format v0.2 I also completely removed the possibility to create empty lines. I think that's much cleaner and simpler:

```scheme
(librepcb_circuit
 (netclass ada6400e-cb5c-41b6-9a4e-e2f39ffa3383 (name "default"))
 (net 0654411b-a090-4025-a026-4c61e686662e (auto false) (name "LED_TX")
  (netclass ada6400e-cb5c-41b6-9a4e-e2f39ffa3383)
 )
 (component 04187e1b-d8f2-490f-8731-96198fa4fa7e
  (lib_component b91cf23a-4f07-4b99-8f52-0b42304aef20)
  (lib_variant 0eebde17-6d45-4f73-95fe-3078a0f68ab6)
  (lib_device none)
  (name "LOGO1") (value "Made With\nLibrePCB")
 )
)
```

Any thoughts welcome :slightly_smiling_face:

### DIFF
--- a/libs/librepcb/core/attribute/attribute.cpp
+++ b/libs/librepcb/core/attribute/attribute.cpp
@@ -113,13 +113,13 @@ bool Attribute::setTypeValueUnit(const AttributeType& type,
 void Attribute::serialize(SExpression& root) const {
   if (!checkAttributesValidity()) throw LogicError(__FILE__, __LINE__);
   root.appendChild(mKey);
-  root.appendChild("type", *mType, false);
+  root.appendChild("type", *mType);
   if (mUnit) {
-    root.appendChild("unit", *mUnit, false);
+    root.appendChild("unit", *mUnit);
   } else {
-    root.appendChild("unit", SExpression::createToken("none"), false);
+    root.appendChild("unit", SExpression::createToken("none"));
   }
-  root.appendChild("value", mValue, false);
+  root.appendChild("value", mValue);
 }
 
 /*******************************************************************************
--- a/libs/librepcb/core/fileio/transactionalfilesystem.cpp
+++ b/libs/librepcb/core/fileio/transactionalfilesystem.cpp
@@ -446,19 +446,25 @@ void TransactionalFileSystem::saveDiff(const QString& type) const {
   }
 
   SExpression root = SExpression::createList("librepcb_" % type);
-  root.appendChild("created", dt, true);
-  root.appendChild("modified_files_directory", filesDir.getFilename(), true);
+  root.ensureLineBreak();
+  root.appendChild("created", dt);
+  root.ensureLineBreak();
+  root.appendChild("modified_files_directory", filesDir.getFilename());
   foreach (const QString& filepath, Toolbox::sorted(mModifiedFiles.keys())) {
-    root.appendChild("modified_file", filepath, true);
+    root.ensureLineBreak();
+    root.appendChild("modified_file", filepath);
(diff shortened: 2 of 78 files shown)


dbrgn:
Yeah, I think that makes sense!

PR #970, opened 2022-04-18
CMake: Add alias for muparser target

dbrgn:
Building LibrePCB with unbundled MuParser on Arch Linux currently fails,
because the target `MuParser::MuParser` is not defined. Because this
used to work before, I assume the target was renamed from
`MuParser::MuParser` to `muparser::muparser` some time ago.

To fix this, add an alias if the CMake package was found, but the pascal
case alias isn't defined yet.

### DIFF
--- a/cmake/FindMuParser.cmake
+++ b/cmake/FindMuParser.cmake
@@ -37,6 +37,11 @@ find_package(muparser 2.0 QUIET)
 if(muparser_FOUND)
   message(STATUS "Using system MuParser")
 
+  # Add uppercase alias if only the lowercase target is defined
+  if(NOT TARGET MuParser::MuParser)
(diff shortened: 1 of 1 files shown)


ubruhin (approved):
Maybe actually it never worked before, but in your previous tests the library was found with pkgconfig? Anyway, the `if` condition isn't a big deal so let's keep it :+1:

dbrgn:
> Maybe actually it never worked before, but in your previous tests the library was found with pkgconfig?

Unlikely, because the pkgconfig code path comes after the cmake `find_library` approach.

I already applied this patch to the AUR package (https://aur.archlinux.org/packages/librepcb) and it seems to build fine now.

i9e1:
I also came across this rename when building the package about 2 weeks ago. 

Building 0.1.6-3 works fine.

PR #973, opened 2022-04-24
GraphicsExportDialog: Fix slow UI due to recursion

ubruhin:
Some operations of the board editor graphics export dialog were very
slow due to unintended recurions. Fixed by blocking the corresponding
Qt signals.


dbrgn:
Typo in the commit message: recurions -> recursion

PR #984, opened 2022-05-29
Implement "Find" and "Find Next/Previous" in editors

ubruhin:
The "search" toolbar was already added to all editors some time ago. However, so far you had to manually move the cursor into that toolbar to start searching, go to the next result (pressing Enter again) or to clear the search keyword. This PR adds several improvements to this feature:

- Add "Find" entry to the menus of all editors, with the keyboard shortcut Ctrl+F. So in any editor you can now press Ctrl+F to move the focus to the search field, then just start typing the term to search for.
- Add "Find Next" and "Find Previous" entries to the menu of the schematic- and board editor, with keyboard shortcuts F3 resp. Shift+F3. These allow to navigate to the next/previous search result even if the focus is not in the search toolbar anymore.
- Fix non-numeric order of search results (e.g. R10 appeared before R2).
- Library editor: Clear the applied filter with the Escape key. No more need to manually clear the search term in the toolbar.


dbrgn:
I didn't test it, but the described fixes sound very nice 🙂

PR #986, opened 2022-06-06
Refactor & highly improve keyboard navigation

### LINKED ISSUE #48: Make Ctrl + Q close the whole Application not just the window

### LINKED ISSUE #49: Use Ctrl + W to close a Project

### LINKED ISSUE #544: Schematic Editor : Add Keyboard Shortcuts for main functionalities
##### SUMMARY
In Schematic Editor, "Select", "Draw Wire", "Add NetLabel" and "Add Component" do not have keyboard shortcuts (or it is not showed in the menu).

##### SOLUTION
Add keyboard shortcuts for main functionalities (not only the four I mentioned) and show it in the menu.

For information, in Kicad, if you type `'?'` you'll get all the shortcuts you can use in the current context.
Think the idea is good, but the implementation is bad, as it is never up to date.

ubruhin:
This is the first step needed to make keyboard shortcuts configurable. I'll open a separate PR later to make them configurable, but the heavy work is already contained in this PR.

Check out [this forum post](https://librepcb.discourse.group/t/hotkeys-anyone/229/22) to see which shortcuts will be added with this PR. There are still some open points, thus only a draft PR so far.

### Add class `EditorCommandSet` containing all commands

- Ensure consistent icons and texts for menu/toolbar actions across the whole application
- Get rid of all duplicated text strings for menu/toolbar actions to reduce the number of strings to translate (reducing effort)
- Assign many more keyboard shortcuts to the actions of all editors
- Allows to make keyboard shortcuts configurable (implemented in a later commit)

### Add helper class `MenuBuilder` for building `QMenu`

- Get rid of duplicate menu translation strings across all editors

### Refactor menus, toolbars and shortcuts in all editors

- Create all `QAction` objects through `EditorCommandSet`, not using the UI designer anymore for this
- Create all `QMenu` objects through `MenuBuilder`, not using the UI designer anymore for this
- Create all toolbars manually by code, not using the UI designer anymore for this
- Add many more actions (incl. shortcuts) to menus and some more to toolbars
- Add many additional shortcuts in library-, schematic- and board editor tools
- In the library editor, dynamically disable all actions in menus and toolbars when not available in the current editor state
- Fix several minor UI issues

### Disable inserting tabulator in all text input fields

To allow keyboard navigation with "tab", even if the focus is in a text input field.

### Use tabulator key to move focus into command toolbar

Instead of (partially) moving the focus automatically into the command toolbar (e.g. when adding a component to the schematic), the focus now always stays in the graphics view. But the tabulator key now moves the focus into the command toolbar in any tool.

Reason: Having the focus in the toolbar sometimes breaks keyboard shortcuts (e.g. 'R' for rotate) because some toolbar widgets contain text input fields which consume these shortcuts - therefore this shouldn't happen automatically.

Downside is that in some tools the user now has to press 'tab' manually to start editing for example the value of a text label, but at least this concept is now consistent across all tools.

### Related issues

- Closes #48 
- Closes #49
- Closes #544


dns2utf8:
Very cool! Thank you! :partying_face:

PR #995, opened 2022-09-03
Implement distance measure tool in all editors

### LINKED ISSUE #861: Add distance measurement tool
##### SUMMARY

A simple tool to measure distances in all editors.

##### SOLUTION

Maybe as a new item in the "Tools" toolbar. The first mouse click sets the start point of the measurement, the second click sets the end point and somehow the distance should be displayed (without a message box would be nice).

Would be nice to snap to underlying graphics elements (e.g. center of pads) by default, and disable snapping by pressing the SHIFT modifier (like the trace drawing tool).

ubruhin:
### Description

This adds a measure tool to the symbol-, package-, schematic- and board editors:

- The tool can be started by its toolbutton, or with Ctrl+M as default shortcut.
- When the tool is active, the whole background is grayed out and a ruler & measurement results are drawn as an overlay.
- The ruler contains ticks & labels for both metric and imperial units, and the tick interval dynamically scales with the zoom level.
- The ruler contains a circle at half the measured distance to allow verifying that something is symmetric.
- The results overlay shows various information in the unit configured in the grid settings (start/end coordinates, dX, dY, distance, angle).
- The first click specifies the start point, the second click the end point of the measurement.
- The ruler snaps on the grid and on many objects (e.g. pad edges, polygon vertices, hole circles, ...).
- Snapping on objects can temporarily be disabled by pressing SHIFT.
- The measured distance can directly be copied into the clipboard by pressing Ctrl+C (useful for further calculations outside of LibrePCB).
- The status bar shows hints about how to use the tool.

### Limitations

To keep the implementation complexity low, currently the tool doesn't respect the layer visibility to determine the snap coordinates, thus the ruler snaps even on hidden objects. During my testing this was not a problem at all. If it is indeed a problem (maybe with much more complex boards) it could be fixed some time later.

### Preview

librepcb-measure-2-small

Closes #861

### DIFF
--- a/libs/librepcb/core/graphics/graphicspainter.cpp
+++ b/libs/librepcb/core/graphics/graphicspainter.cpp
@@ -109,19 +109,24 @@ void GraphicsPainter::drawCircle(const Point& center, const Length& diameter,
 void GraphicsPainter::drawText(const Point& position, const Angle& rotation,
                                const Length& height, const Alignment& alignment,
                                const QString& text, QFont font,
-                               const QColor& color,
-                               bool mirrorInPlace) noexcept {
+                               const QColor& color, bool autoRotate,
+                               bool mirrorInPlace, int fontPixelSize) noexcept {
   if (text.trimmed().isEmpty() || (!color.isValid())) {
     return;  // Nothing to draw.
   }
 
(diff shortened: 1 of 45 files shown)


dbrgn:
That looks awesome! 🎉

PR #1010, opened 2022-09-10
Library editor: Fix possible crash when closing tabs

ubruhin:
If a command was active while closing a library editor tab, in some cases it could lead to a crash due to unintended recursion during object destruction (at least in the symbol editor). Fixed by a more explicit cleanup procedure.

### DIFF
--- a/libs/librepcb/editor/library/pkg/packageeditorwidget.cpp
+++ b/libs/librepcb/editor/library/pkg/packageeditorwidget.cpp
@@ -119,14 +119,14 @@ PackageEditorWidget::PackageEditorWidget(const Context& context,
   updateMetadata();
 
   // Setup footprint list editor widget.
-  mUi->footprintEditorWidget->setReferences(mPackage->getFootprints(),
-                                            *mUndoStack);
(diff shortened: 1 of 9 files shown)


libs/librepcb/editor/library/pkg/packageeditorwidget.cpp
```diff
@@ -171,9 +171,15 @@ PackageEditorWidget::PackageEditorWidget(const Context& context,
 }
 
 PackageEditorWidget::~PackageEditorWidget() noexcept {
+  // Clean up the state machine nicely to avoid unexpected behavior.
+  mFsm->processAbortCommand();
+  mFsm->processAbortCommand();
+  mFsm->processAbortCommand();
```
dbrgn, line 177: I just skimmed over the changes, but: Why 3x the same abort call? Is 3 a specific number that is needed? Maybe a loop with an explaining loop count variable name might be a better solution?

ubruhin: This abort call does the same as the 'Esc' key, but in some "nested" editor states you need to press 'Esc' several times to completely exit the whole tool. Added a comment to clarify this :)

PR #1013, opened 2022-09-11
CMake: Fix & improve unbundling of GTest/GMock

### LINKED ISSUE #933: CMake: Infinite recursion when finding gtest and gmock
I get this error when building. I'm using the compile flag -DUNBUNDLE_GTEST=ON.

I don't understand how cmake works but my guess is line 24 in cmake/FindGTest.cmake where it tried to find_package(Gtest) is actually running the FindGTest.cmake file again. 

```
CMake Error at cmake/FindGTest.cmake:1 (set):
  Maximum recursion depth of 1000 exceeded
Call Stack (most recent call first):
  cmake/FindGTest.cmake:24 (find_package)
  cmake/FindGTest.cmake:24 (find_package)
  cmake/FindGTest.cmake:24 (find_package)
  cmake/FindGTest.cmake:24 (find_package)
  ...
  (lots more of the same here)
  ...
  cmake/FindGTest.cmake:24 (find_package)
  CMakeLists.txt:141 (find_package)
```

ubruhin:
Adding `CONFIG` to the `find_package()` seems to fix the endless recursion which was reported in #933. I don't understand what `CONFIG` means exactly, but at least [in this](https://chromium.googlesource.com/external/github.com/pwnall/googletest/+/HEAD/googletest/README.md) documentation they also do it this way. And local testing shows that this seems to fix the problem.

In addition, I adjusted `FindGTest.cmake` to proceed even if `find_package(GMock)` failed - `find_package(GTest)` should be sufficient since GTest usually also contains GMock, so no separate package is needed for GMock.

I also extended CI to test unbundling of GTest/GMock. However, since Ubuntu doesn't provide a cmake config, I extended `FindGTest.cmake` to also search with `pkg-config` which is then used on Ubuntu. I hope it works (waiting for the pipeline...).

See also the Docker image update: https://github.com/LibrePCB/docker-librepcb-dev/pull/20

Fixes #933

### DIFF
--- a/ci/azure-jobs-linux.yml
+++ b/ci/azure-jobs-linux.yml
@@ -19,9 +19,9 @@ jobs:
         CC: clang
         CXX: clang++
       Ubuntu_2004_GCC:
-        IMAGE: librepcb/librepcb-dev:ubuntu-20.04-6
+        IMAGE: librepcb/librepcb-dev:ubuntu-20.04-7
       Ubuntu_2004_Unbundled:
-        IMAGE: librepcb/librepcb-dev:ubuntu-20.04-6
+        IMAGE: librepcb/librepcb-dev:ubuntu-20.04-7
         UNBUNDLE_DXFLIB: true
         UNBUNDLE_MUPARSER: true
         #UNBUNDLE_QUAZIP: true  # Note: Does not ship quazip 1.x yet
--- a/ci/azure-jobs-linux.yml
+++ b/ci/azure-jobs-linux.yml
@@ -23,6 +23,7 @@ jobs:
       Ubuntu_2004_Unbundled:
         IMAGE: librepcb/librepcb-dev:ubuntu-20.04-7
         UNBUNDLE_DXFLIB: true
+        UNBUNDLE_GTEST: true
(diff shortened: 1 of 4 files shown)


dbrgn:
Regarding the modes, see https://cmake.org/cmake/help/latest/command/find_package.html#id4

ubruhin:
> Regarding the modes, see https://cmake.org/cmake/help/latest/command/find_package.html#id4

Of course I saw that documentation, but for me this does not make clear whether it is correct to use `CONFIG` or not. So far I just *assume* it is fine since the GTest documentation uses it as well (and because it seems to work).

dbrgn:
Yes, from what I understand, the `*-config.cmake` (or `*Config.cmake`) file is provided by the package itself, and the `Find*.cmake` file is provided by the distro. At least in the Arch package, a `GTestConfig.cmake` file is provided (I assume it's the same file as in Ubuntu). So `CONFIG` is the correct mode in this case.

dbrgn (approved):

ubruhin:
OK nice, thanks! :+1:

PR #1021, opened 2022-09-17
Make polygon/plane vertex handle size independent of zoom

### LINKED ISSUE #846: Library Editor : Change size of vertex control handle according to the zoom
##### SUMMARY
Vertex control is a great functionality.
Unfortunately in some case it is impossible to do what we want, especially when polygon is really small (0.4x0.2mm for example).
Vertex control handles use all the space and we can click on the polygon to move it.

image

The circles of control have fixed size. So even at maximum zoom, it is not possible.

image

##### SOLUTION

Modify the size of handles according to the zoom (it can be non linear).

### LINKED ISSUE #910: Board Editor : Vertex focus remains even if shape not selected

```
LibrePCB Version: 0.1.5
Git Revision:     20f2c30ac403cc69fa0db4cbb85acfb43681abc0
Build Date:       2021-07-11T12:00:30
Linking type:     static
Unbundled libs:   -
CPU Architecture: x86_64
Operating System: Ubuntu 20.10
Platform Plugin:  xcb
```

##### STEPS TO REPRODUCE

1. select a shape
2. select another shape

##### EXPECTED RESULTS

Only one shape selected

##### ACTUAL RESULTS
image

ubruhin:
- Keep polygon/plane vertex handle size constant in terms of screen size instead of scene size, i.e. they are no longer scaled when zooming.
- Reduce size of vertex handled which are close to each other, without affecting the (invisible) area where the vertices can be grabbed. This avoids overlapping handles (which look ugly) while still allow easy dragging.
- When clicking in between two vertices which are very close to each other, do no longer grab *both* vertices (which would merge them together), but only grab the vertex which is closer to the cursor.
- Since the visual handles do no longer always correspond to the actual grab areas, the handles are now drawn blurry. This makes it clear to the user that the grab area is not always just a simple circle, but rather suggests "the nearer, the better".

I'm not fully happy with the look (didn't have a better idea), but the usability is clearly **way** better than before :slightly_smiling_face: 

librepcb-vertex-handles

Fixes #846
Fixes #910

### DIFF
--- a/libs/librepcb/core/graphics/polygongraphicsitem.cpp
+++ b/libs/librepcb/core/graphics/polygongraphicsitem.cpp
@@ -23,6 +23,7 @@
 #include "polygongraphicsitem.h"
 
 #include "../graphics/graphicslayer.h"
+#include "../utils/toolbox.h"
 
 #include <QtCore>
 #include <QtWidgets>
@@ -43,11 +44,12 @@ PolygonGraphicsItem::PolygonGraphicsItem(Polygon& polygon,
     mPolygon(polygon),
     mLayerProvider(lp),
     mEditable(false),
+    mVertexHandleRadiusPx(0),
+    mVertexHandles(),
     mOnEditedSlot(*this, &PolygonGraphicsItem::polygonEdited) {
-  setPath(mPolygon.getPath().toQPainterPathPx());
   setLineWidth(mPolygon.getLineWidth());
   setLineLayer(mLayerProvider.getLayer(*mPolygon.getLayerName()));
-  updateVertexGraphicsItems();
+  updatePath();
   updateFillLayer();
   setFlag(QGraphicsItem::ItemIsSelectable, true);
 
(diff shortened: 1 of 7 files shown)


dbrgn:
Hmm, I'm not too fond of the fuzzy circles either...

The grab area and circles are only different when two handles are close together, right? If that's the case, then I wouldn't see an issue with drawing circles. If two are close together, then it should be obvious that the closer one wins when trying to select one of them.

ubruhin:
> The grab area and circles are only different when two handles are close together, right?

That's one case, but "close together" is relative since it depends on the zoom level (when zoomed out, the handled will be shrunk even for large polygons). And there's also another special case: When the grab area is *smaller* than the line width, the circles wouldn't be visible at all since they would be drawn just on top of the line (and using other, automatically determined color is not trivial). With the blurry circles, I can make them just a bit larger than the line width to still see the visual handles without suggesting the user that he can drag on these areas.

> If two are close together, then it should be obvious that the closer one wins when trying to select one of them.

Yes that should be obvious, but it is very strange that in this case the actual grab area can be much larger than the circle - one thinks he can click outside these circles to clear the selection, but this doesn't work and feels like a bug. Same again, with the blurry circles we don't make a wrong suggestion about the actual grab area.

It's really hard to explain all this with text. I played around hours with many different solutions. The behavior depends a lot of the combination of line width, zoom level and vertex distance.

I think the only reasonable alternative would be to use circles which represent the actual grab area (i.e. don't shrink or enlarge them depending on vertex distance and line width). I guess the main drawback of that is that it leads to overlapping circles which doesn't look nice. If there are serious doubts about the current solution, I could give it a try again... :wink:

PR #1037, opened 2022-09-27
Show waiting spinner in tree/list views during library scan

ubruhin:
In all tree views and list views which show content from the workspace library database (e.g. categories, symbols, components, ...), show a spinner during the scan to indicate that the content might be outdated and will be updated soon.

This should help to avoid confusion when users add libraries or library elements, but then can't find those things in the tree/list views. Especially after installing a new library, the whole library seems to be empty for quite some time - with the spinner, it should be clear that some update is in progress:

librepcb-waiting-spinner

In addition, another similar improvement: If the workspace library background scan is in progress while a new window is opened, the progress bar in the status bar did not appear immediately but only on the next progress update event (could be several seconds later). Now the progress bar is shown immediately when opening a window while the scan is in progress.

### DIFF
--- a/libs/librepcb/core/workspace/workspacelibrarydb.cpp
+++ b/libs/librepcb/core/workspace/workspacelibrarydb.cpp
@@ -99,6 +99,10 @@ WorkspaceLibraryDb::~WorkspaceLibraryDb() noexcept {
  *  Getters
  ******************************************************************************/
 
+int WorkspaceLibraryDb::getScanProgressPercent() const noexcept {
+  return mLibraryScanner->getProgressPercent();
+}
+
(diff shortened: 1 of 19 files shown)


dbrgn:
Nice! Would it look better with horizontally centered spinners?

ubruhin:
> Nice! Would it look better with horizontally centered spinners?

It might on empty lists, but not on populated lists. I placed them on the right since it reduces overlapping with the content:

image

And yes, it's intended to also show the spinners on populated lists if the scan is in progress, to indicate that the lists might be outdated (but they are fully functional even if the spinner is visible).

dbrgn:
Ah, that makes sense!

PR #1062, opened 2022-11-02
Add .git-blame-ignore-revs file

dbrgn:
I learned a new git trick today: https://www.michaelheap.com/git-ignore-rev/ (Inspired by https://github.com/microsoft/TypeScript/pull/51387) By using the `.git-blame-ignore-revs` file, you won't have reformatting commits polluting your blame output.

To use it, invoke `git blame` like this:

    git blame --ignore-revs-file .git-blame-ignore-revs

Alternatively, make this the default:

    git config [--global] blame.ignoreRevsFile .git-blame-ignore-revs

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,11 @@
+# 2018-09-28: Initial clang-format reformatting
+05a59de4341d53a5d4d942992eafb77bf3fbf006
+
+# 2018-12-09: Reformat with clang-format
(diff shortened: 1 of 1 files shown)


ubruhin (approved):
Hm okay, I almost never use git blame, but shouldn't hurt to add this file :see_no_evil: :slightly_smiling_face:

dbrgn:
I use it often to see the context, in which a line of code or a function was originally added 🙂

PR #1063, opened 2022-11-18
CLI: Use -v for --verbose instead of --version.

### LINKED ISSUE #1008: CLI: Use "-v" for "--verbose" instead of "--version"
##### SUMMARY

It's just a detail and I'm not sure if it's even worth to change it since technically this would be a breaking change. But somehow I think it would be more intuitive to use `-v` for `--verbose` instead of `--version` which is currently the case. Especially if we ever want to add several verbosity levels, it's very common to do so by `-vv` and `-vvv`. The short flag for `--version` could be changed to uppercase `-V` since this is also very common.

##### SOLUTION

- Use `-v` for `--verbose`
- Use `-V` for `--version`

However, since it's not really important and actually a breaking change (although probably nobody depends on `-v`), I'd postpone this to v0.2 to comply with semver.

Any opinions welcome :)

ammoniak:
Use -v for --verbose instead of --version.
Added -V for --version.

This is my first pull request for this project. Let me know if I missed something :)

Fixes #1008

### DIFF
--- a/apps/librepcb-cli/commandlineinterface.cpp
+++ b/apps/librepcb-cli/commandlineinterface.cpp
@@ -90,7 +90,7 @@ int CommandLineInterface::execute() noexcept {
   // option which we don't need, and the OS-dependent option "-?".
   const QCommandLineOption helpOption({"h", "help"}, tr("Print this message."));
   parser.addOption(helpOption);
-  const QCommandLineOption versionOption({"V","version"},tr("Show version."));
+  const QCommandLineOption versionOption({"V","version"},tr("Displays version information."));
   parser.addOption(versionOption);
   QCommandLineOption verboseOption({"v","verbose"}, tr("Verbose output."));
   parser.addOption(verboseOption);
--- a/apps/librepcb-cli/commandlineinterface.cpp
+++ b/apps/librepcb-cli/commandlineinterface.cpp
@@ -90,9 +90,10 @@ int CommandLineInterface::execute() noexcept {
   // option which we don't need, and the OS-dependent option "-?".
   const QCommandLineOption helpOption({"h", "help"}, tr("Print this message."));
   parser.addOption(helpOption);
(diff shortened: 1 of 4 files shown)


apps/librepcb-cli/commandlineinterface.cpp
```diff
@@ -90,8 +90,9 @@ int CommandLineInterface::execute() noexcept {
   // option which we don't need, and the OS-dependent option "-?".
   const QCommandLineOption helpOption({"h", "help"}, tr("Print this message."));
   parser.addOption(helpOption);
-  const QCommandLineOption versionOption = parser.addVersionOption();
-  QCommandLineOption verboseOption("verbose", tr("Verbose output."));
+  const QCommandLineOption versionOption({"V","version"},tr("Show Version.")); // = parser.addVersionOption();
```
ubruhin, line 93: I'd make the "Version" lowercase for consistency with the rest of the help text. And let's remove the comment to keep it clean :see_no_evil: 

```suggestion
  const QCommandLineOption versionOption({"V","version"},tr("Show version."));
```

ubruhin (changes requested):
Nice, thanks for contributing! :rocket: 

Beside the comment below:
- There are automated tests for the CLI which need to be adjusted as well (I assume CI will currently fail). See directory https://github.com/LibrePCB/LibrePCB/tree/master/tests/cli. Probably only some strings need to be updated, but no test logic.
- Would be great if you could format the code according our coding style. At least on Linux (with Docker) it's dead easy, see instructions [here](https://developers.librepcb.org/df/d24/doc_code_style_guide.html).

Let me know if you need any help. Or if you can't format the code easily, I could push a commit to do so.

ammoniak:
I will fix the test cases

ubruhin (approved):
Great, thanks! I added a commit to auto-format the code. Will merge it once CI succeeds :rocket:

ammoniak:
Thank you. I'll add the auto format to my IDE for the next one.

PR #1071, opened 2022-12-10
Add support for NPTH slots

### LINKED ISSUE #505: Add support for slotted NPTH

ubruhin:
### Summary

Replaces #510. But instead of specifying slots with just a width and height property (as originally intended by a `DrillSize` class), I decided to make slots even more flexible. Since the Excellon file format supports creating arbitrary slots by movement commands (G00..G03), let's reflect that flexibility in LibrePCB as well.

### File Format

So far, a hole object was stored this way:

```lisp
(hole 4ab2be01-222a-4f59-b239-e1de803ba96d (diameter 3.0) (position 2.54 2.54))
```

Now a hole is specified by vertices (exactly the same way as for polygons):

```lisp
(hole 4ab2be01-222a-4f59-b239-e1de803ba96d (diameter 3.0)
 (vertex (position 2.54 2.54) (angle 0.0))
)
```

Holes with only one vertex are circular. Holes with multiple vertices are slots. They can even be curved (`angle` != 0):

```lisp
(hole 4ab2be01-222a-4f59-b239-e1de803ba96d (diameter 1.6)
 (vertex (position 20.691974 1.641974) (angle 90.0))
 (vertex (position 22.0 3.438026) (angle 0.0))
 (vertex (position 23.0 5.0) (angle -90.0))
 (vertex (position 24.0 8.0) (angle 0.0))
)
```

### User Interface

The hole properties dialog allows to specify holes either as circular drills, simple linear slots, or arbitrary slots:

librepcb-npth-slots

The editors do not yet contain a tool to draw slots. Currently you can just add circular holes and and then convert them to slots by modifying their properties. A slot tool could be added later without needing any file format change.

### Excellon Export

By default, slotted holes are exported with commands G00..G03 in the Excellon drill file. This seems to be the recommended way (KiCad doesn't recommend G85 and the [XNC Specs](https://www.ucamco.com/files/downloads/file_en/452/xnc-format-specification-revision-2021-11_en.pdf?2ccd0fdd6dd78e87ed52185bbd06038b) from Ucamco doesn't even mention G85; see also https://github.com/LibrePCB/LibrePCB/pull/510#issuecomment-528586891).

However, there might be manufacturers which require slots exported as G85 slot commands. Even Gerbv does only render G85 correctly, but not G00..G03 which is quite a shame. Therefore I made it configurable which command to use:

image

Since G85 cannot draw arcs, it can be used only for boards which do not contain curved slots. If there are curved slots, an error occurs during the export.

### DRC

The support of curved slots unfortunately varies between different PCB manufacturers. Multi-segment slots maybe too, I'm not sure. And whether G85 shall be used or not, also depends on the manufacturer. To avoid surprises during production, I added a new DRC rule which allows to warn if slots of a particular style are used in a board:

image

Fixes #505.

### DIFF
--- a/libs/librepcb/core/export/excellongenerator.cpp
+++ b/libs/librepcb/core/export/excellongenerator.cpp
@@ -23,6 +23,8 @@
 #include "excellongenerator.h"
 
 #include "../fileio/fileutils.h"
+#include "../types/point.h"
+#include "../utils/toolbox.h"
 
 #include <QtCore>
 
@@ -41,7 +43,7 @@ ExcellonGenerator::ExcellonGenerator(const QDateTime& creationDate,
                                      const QString& projRevision,
                                      Plating plating, int fromLayer,
                                      int toLayer) noexcept
-  : mPlating(plating), mFileAttributes(), mOutput() {
+  : mPlating(plating), mFileAttributes(), mUseG85Slots(false), mOutput() {
   mFileAttributes.append(GerberAttribute::fileGenerationSoftware(
       "LibrePCB", "LibrePCB", qApp->applicationVersion()));
   mFileAttributes.append(GerberAttribute::fileCreationDate(creationDate));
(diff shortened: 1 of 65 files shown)


dbrgn:
Awesome, very thought-out PR! I like all the design decisions.

ubruhin:
Added one more thing: Since manufacturers usually cannot create slots as small as drills, I have now changed the design rule parameter "Minimum NPTH Drill Diameter" to check only circular drills, and added a new rule "Minimum NPTH Slot Width" which checks only slots. Now one can configure different minimums for drills and slots, as specified by the PCB manufacturers.

PR #1076, opened 2023-01-03
Add support for slotted pads

### LINKED ISSUE #418: Add support for slotted pad holes

ubruhin:
### Summary

Replaces #423 with a different approach, see details in #1071 which did the same but for non-plated holes.

### File Format

So far, a footprint pad object was stored this way:

```lisp
(pad 474059df-640b-4b17-98b2-148e340a7152 (side tht) (shape round)
 (position 2.54 0.0) (rotation 90.0) (size 4.064 2.032) (drill 1.016)
 (package_pad 474059df-640b-4b17-98b2-148e340a7152)
)
```

Now a pad acts as a kind of container which can contain 0..n hole objects as introduced in #1071:

```lisp
(pad 474059df-640b-4b17-98b2-148e340a7152 (side top) (shape round)
 (position 2.54 0.0) (rotation 90.0) (size 4.064 2.032)
 (package_pad 474059df-640b-4b17-98b2-148e340a7152)
 (hole 474059df-640b-4b17-98b2-148e340a7152 (diameter 1.016)
  (vertex (position 0.0 0.0) (angle 0.0))
 )
)
```

A pad without holes is automatically considered as SMT. If there's at least one hole, it's THT. One might wonder why there can be multiple holes in a pad - the implementation of 1 vs. n holes is almost the same and I think there might indeed be some cases where this is useful, so I decided to allow more than 1 hole. The idea to consider a pad as a container of objects might be extended to polygons some day, e.g. to attach a custom solder mask polygon to a pad.

Note that the `(side tht)` is automatically converted to `(side top)` during the file format upgrade. Although this property doesn't make that much sense for a THT pad, I don't like omitting it since this would lead to a different file format for SMT and THT pads. In the GUI I called the property "Component Side", I hope this way it's not too confusing for users (even a THT pad is assembled only from one side).

### Example

Now you can draw plated smileys :joy:

image

### User Interface

The same hole editor widget as introduced in #1071 is reused, and embedded into the footprint pad editor. If there's only one, round hole it can be directly edited as usual:

image

For more complex configurations, the new tab needs to be used. If there are multiple holes, you need to iterate through them:

image

The interface is for sure not the most convenient, but for these *very* rare complex cases it's not worth spending too much time on it. Simple cases (probably >99% of pads are simple) are still very easy to create & modify. 

### Excellon Export

Identical to #1071 (same export option reused).

### DRC

Two new DRC checks added, exactly the same as in #1071 but for plated holes.

### Package Check

Since due to the much higher flexibility of pad holes, it's now easier to create pads which cannot be manufactured. Therefore I implemented a new check in the package editor which now warns when creating any pad with a restring smaller than 150μm (also applies to overlapping holes or holes completely outside the pad).

Fixes #418.

### DIFF
--- a/libs/librepcb/core/CMakeLists.txt
+++ b/libs/librepcb/core/CMakeLists.txt
@@ -256,6 +256,8 @@ add_library(
   library/pkg/msg/msgpadclearanceviolation.h
   library/pkg/msg/msgpadoverlapswithplacement.cpp
   library/pkg/msg/msgpadoverlapswithplacement.h
+  library/pkg/msg/msgpadrestringviolation.cpp
+  library/pkg/msg/msgpadrestringviolation.h
   library/pkg/msg/msgwrongfootprinttextlayer.cpp
   library/pkg/msg/msgwrongfootprinttextlayer.h
   library/pkg/package.cpp
--- a/libs/librepcb/core/library/pkg/footprintpad.cpp
+++ b/libs/librepcb/core/library/pkg/footprintpad.cpp
@@ -51,36 +51,6 @@ SExpression serialize(const FootprintPad::Shape& obj) {
 }
 
 template <>
-SExpression serialize(const FootprintPad::BoardSide& obj) {
-  switch (obj) {
-    case FootprintPad::BoardSide::TOP:
-      return SExpression::createToken("top");
-    case FootprintPad::BoardSide::BOTTOM:
-      return SExpression::createToken("bottom");
-    case FootprintPad::BoardSide::THT:
-      return SExpression::createToken("tht");
-    default:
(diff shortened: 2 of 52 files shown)


dbrgn:
Awesome, thanks! Finally proper support for USB connectors 😄

pinoaffe:
Whoa, great to see this!

PR #1080, opened 2023-01-10
Use rounded corners for all expanded pad/via shapes

### LINKED ISSUE #597: Board editor: Make plane internal angles rounded as external angles
##### SUMMARY
when making a plane, external angles are rounded but not internal angles.

image

##### SOLUTION
make all angles rounded

image

ubruhin:
Instead of expanding pad/via shapes by keeping the same shape (i.e. sharp corners), this will expand them by maintaining a constant clearance around the whole shape (i.e. make corners rounded). It affects vias and pads with either a rectangular or octagon shape, in particular:

- Cutouts of vias/pads in copper planes (see #597).
- Stop mask / cream mask via/pad shapes if their expansion is > 0.
- Both of these things will have an effect on the Gerber output.

It's not a big thing, but it is slightly more space efficient (useful for dense boards) and ensures the plane area is as large as possible. A drawback is that the generated Gerber apertures are quite some more complicated now (there are no builtin apertures for rounded rects/octagons) but I think that should not be an issue.

image

Closes #597.

### DIFF
--- a/libs/librepcb/core/export/gerberaperturelist.cpp
+++ b/libs/librepcb/core/export/gerberaperturelist.cpp
@@ -125,41 +125,74 @@ int GerberApertureList::addObround(const PositiveLength& w,
 }
 
 int GerberApertureList::addRect(const PositiveLength& w,
-                                const PositiveLength& h, const Angle& rot,
+                                const PositiveLength& h,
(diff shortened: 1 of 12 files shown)


dbrgn:
I like it, looks much better 🙂

PR #1082, opened 2023-01-13
Fix possibly unexpected center of rotate/mirror operations

### LINKED ISSUE #728: Board Editor: Move and rotate unaligns devices

```
LibrePCB Version: 0.1.4
Git Revision:     ae04eef5a71b5ba66ae2cee6b631c1c933ace535
Build Date:       2020-05-02T18:50:43
Qt Version:       5.11.3 (built against 5.11.3)
CPU Architecture: x86_64
Operating System: Windows 10 (10.0)
Platform Plugin:  windows
```

##### SUMMARY
When a device, aligned to the snap grid, is selected, moved with left hold, and then rotated with right-click, the alignment is destroyed.

##### STEPS TO REPRODUCE
Open a board with a device aligned to the snap grid, select with left click, move with left hold, and rotate 1-3 times with right-click.

##### EXPECTED RESULTS
An aligned device is still aligned to the snap grid.

##### ACTUAL RESULTS
The device is no longer aligned.

##### ADDITIONAL INFORMATION
- It seems, that the rotation origin does not match the displayed device origin.
- The offset rotation origin seems to change with the relation of the actual mouse pointer in regard to the device origin. A bigger distance also results in a bigger offset of the origin.
- After four rotations, the device is again aligned with the snap grid.

ubruhin:
When rotating/mirroring items in any of the editors, so far the center point of the operation was always mapped to the grid interval. However, this is not expected when only a single item is selected since it moves the item's position if it was not aligned to the grid. Thus now using the item's exact center position (not mapped to grid) if only one is selected. For example rotating a single, unaligned pad now keeps its exact position.

If more than one item is selected, the behavior is not affected by this change.

This also seems to fix a bug (#728) of possibly unaligning a previously aligned device.

Closes #728.

### DIFF
--- a/libs/librepcb/editor/library/cmd/cmddragselectedfootprintitems.cpp
+++ b/libs/librepcb/editor/library/cmd/cmddragselectedfootprintitems.cpp
@@ -131,8 +131,11 @@ CmdDragSelectedFootprintItems::CmdDragSelectedFootprintItems(
     ++count;
   }
 
-  mCenterPos /= qMax(count, 1);
-  mCenterPos.mapToGrid(grid);
+  // Note: If only 1 item is selected, use its exact position as center.
+  if (count > 1) {
+    mCenterPos /= count;
+    mCenterPos.mapToGrid(grid);
+  }
 }
 
 CmdDragSelectedFootprintItems::~CmdDragSelectedFootprintItems() noexcept {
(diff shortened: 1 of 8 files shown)


vicencb:
On LibrePCB 2.1.1 it looks like "If more than one item is selected, the behavior **is** affected by this change."

That is, if i select multiple _aligned_ items on the board editor and mirror them as a group, they end up off grid. The issue seems to only affect mirroring, rotation is fine.

Also, if i flip twice the same selection, it reports: `Trace on layer "Bottom Copper" cannot be connected to the pad "2" of device "D15" (SMBJ-CA) since it is on layer "Top Copper".`
But if i flip it once, deselect, select again and flip again, it works as expected and all items recover their original alignment.

PR #1086, opened 2023-01-25
Make graphics appearance configurable in workspace settings

### LINKED ISSUE #496: Feature Request: Workspace Settings: Custom Color settings
##### SUMMARY
Custom color settings for each sections: library editor, sch. editor, pcb editor and their each graphics object (Pad, Via, Hole, Top Signal layer, Wire, Junction....). 

Each graphics object could be drawing by custom color. These colors presets, could be save to workspace settings file.

Why? Custom color system is better for designers. Each designer prefer own color scheme.
Again, Why? Try imagine, in your job (8 hour per day) you use commercial XY EDA system. You have fixed color for some object. And at home you want use LibrePCB for small DYI project. And you will lost in another color scheme...

##### SOLUTION
Custom setting can be accesible through workspace settings (for all sections separately: Library editor, sch. editor and pcb editor), and also in each section in separate menu Ex.: Edit->Preferences ->Colors.

Something like this: 
screen2

After clicking to colored rectangle, will open color picker for selecting another color.
Like this:
screen3

### LINKED ISSUE #654: Make background color configurable
Allow configuring the background color similar to the other board editor colors.

This requires a file format change.

@ubruhin probably an easy fix, right?

ubruhin:
Extending workspace settings to allow configuring **background/grid colors, layer colors, grid style** used in all graphics views. Multiple configurations, called *themes* can be created while one of them is used as the active theme. The idea is to support e.g. dark/light themes in future (currently only one builtin theme exists).

image

If you like, a schematic could now look like this :smiley: 

image

Closes #496, closes #654


dbrgn:
Very cool! Is there some kind of efficient way to develop such a theme? (E.g. editing a file in a text editor, and seeing the UI update in real time, or after pressing a key...)

ubruhin:
> Is there some kind of efficient way to develop such a theme? (E.g. editing a file in a text editor, and seeing the UI update in real time, or after pressing a key...)

No, currently you need to close & reopen a project as the settings are applied when opening a project. We can improve this in a later minor release, currently I'm focusing on file format changes to get 0.2.0 finished as soon as possible...

PR #1087, opened 2023-01-25
Build and show list of project file format upgrade messages

ubruhin:
The file format upgrade procedure can affect projects in several ways (e.g. modifying plane calculations, conversions due to obsolete features, ...) so it would be important to inform the user about such changes.

Thus the file format upgrade mechanism is now able to build upgrade notes for each particular project and the schematic editor shows a banner to inform the user about the upgrade. A hyperlink then opens the detailed list of messages:

image

The banner automatically disappears when saving the project since the upgrade is non-reversible then (before saving, the user could decide to close the project, effectively keeping the old file format).

In addition, the CLI now also prints these messages to stdout.


dbrgn:
👍🏻 nice solution!

PR #1089, opened 2023-01-30
Support partially upgrading from the previous unstable file format

ubruhin:
Allow partially upgrading the previous unstable file format if the `LIBREPCB_UPGRADE_UNSTABLE` environment variable is set to `1`. This allows to upgrade our test data files and thus heavily simplifies developing file format changes.

### DIFF
--- a/libs/librepcb/core/CMakeLists.txt
+++ b/libs/librepcb/core/CMakeLists.txt
@@ -430,6 +430,8 @@ add_library(
   project/schematic/schematicselectionquery.h
   serialization/fileformatmigration.cpp
   serialization/fileformatmigration.h
+  serialization/fileformatmigrationunstable.cpp
+  serialization/fileformatmigrationunstable.h
   serialization/fileformatmigrationv01.cpp
   serialization/fileformatmigrationv01.h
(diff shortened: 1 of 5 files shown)


Angelsen:
Having previously manually changed my settings I would like to try this feature. Should it be as simple as running 

`librepcb LIBREPCB_UPGRADE_UNSTABLE=1`

or is it added to a config file?

ubruhin:
Since it is an environment variable, it's actually the other way around (at least on Linux, not sure if it works the same way on other operating systems): 

    LIBREPCB_UPGRADE_UNSTABLE=1 librepcb

**However:** You are aware that using unstable LibrePCB builds WILL break your libraries and projects? :wink: This env var is intended only for development, not for productive use at all (it's not that useful since it only allows to upgrade only the previous file format revision, skipping revisions won't work).

Angelsen:
Thanks for the heads up on the risks of using the unstable release of LibrePCB. While I understand that it's not intended for productive use, I'm eager to test out the latest features and contribute to their development. However, I only use the unstable release for simpler projects. I appreciate your cautionary advice.

PR #1103, opened 2023-02-16
Add support for footprint pads with custom corner radius

### LINKED ISSUE #1027: Support rounded rectangular pads with configurable radius
##### SUMMARY

In other tools, rectangular pads with rounded edges are common pad shapes, with configurable radius. We should provide this feature as well.

##### SOLUTION

Maybe just add a parameter `radius` to rectangular pads (i.e. do not introduce a new pad shape). But let's also check how other tools handle that. I wonder if we should even remove the `round` pad shape then since it would be redundant :thinking: 

See also #540.

ubruhin:
### Summary

Adding a new property `radius` to footprint pads which allows to specify the corner radius in a range of 0..100%. The radius is applied to rectangular pads and octagonal pads.

Since a rectangular pad with 100% radius is the same as a rounded pad, we have a redundancy in the file format (two different representations for the same shape). Therefore I decided to remove the shape `round` from the file format - now a rounded pad is set to `rect` with a corner radius of 100%. Not perfectly intuitive unfortunately, but I think it's better than the redundancy... *EDIT: Renamed to `roundrect`.*

### File Format

Added a new node `radius`:

```scheme
  (pad 474059df-640b-4b17-98b2-148e340a7152 (side top) (shape roundrect)
   (position 2.54 0.0) (rotation 90.0) (size 4.064 2.032) (radius 1.0)
   (package_pad 474059df-640b-4b17-98b2-148e340a7152)
   (hole 474059df-640b-4b17-98b2-148e340a7152 (diameter 1.016)
    (vertex (position 0.0 0.0) (angle 0.0))
   )
  )
```

During the 0.1->0.2 file format migration, `(shape round)` is converted to `(shape roundrect) (radius 1.0)` and  `(shape rect)` is converted to `(shape roundrect) (radius 0.0)`. For other shapes, just a `(radius 0.0)` is added.

### User Interface

*Outdated - see update below.*

For convenience, the round shape is still available in the UI since it is used very often. It is then automatically "converted" to rectangular with 100% corner radius. The corner radius can only be specified manually if either rectangular or octagonal shape is selected:

image

The two input fields are connected - editing one of them automatically updates the other one.

Closes #1027


dbrgn:
Nice! Slightly rounded corners look great on PCBs.

A bit OT, but: I wonder what this means for our library guidelines and for the editor. Should there be a slight corner radius by default?

Also: I assume this the percentage refers to the shorter of the two lengths, correct?

ubruhin:
> A bit OT, but: I wonder what this means for our library guidelines and for the editor. Should there be a slight corner radius by default?

Good question, let me do some research if there are any recommendations about that. Maybe it would indeed make sense to use rounded corners by default.

> Also: I assume this the percentage refers to the shorter of the two lengths, correct?

Correct.

Issue #1007, opened 2022-09-09
CLI: Support running the library element check


ubruhin:
##### SUMMARY

Just like the ERC, we also need the ability to run library element checks from the CLI since this would be very useful for CI of libraries. However, #1006 needs to be implemented first.

##### SOLUTION

Probably similar to the ERC. However, maybe it should somehow be configurable *which* checks need to be executed?

Issue #1006, opened 2022-09-09
Support dismissing library element check messages


ubruhin:
##### SUMMARY

Currently there's no way to dismiss a false-positive library element check message, thus the check is not usable for CI of libraries yet as there will always be cases where some messages are expected. So we need to be able to suppress specific message, and this state needs to be saved in the library element (just like the ERC).

##### SOLUTION

Details need to be clarified, but probably somehow similar to the ERC.

ubruhin:
Implemented in #1107

PR #1106, opened 2023-02-20
Support dismissing library check messages

ubruhin:
## Summary

Support dismissing library check messages, which is then stored in the corresponding files to make it persistent (see #1006). I called it "approvals". This feature is a requirement to allow running the library checks from the CLI (#1007).

## User Interface

image

Approved messages are currently not removed from the list to allow reviewing them easily and usually the list is not flooded by lots of messages anyway.

## File Format

Within each library elements S-Expression file, new nodes are added for each approved message:

```scheme
 (approval "PadOriginOutsideCopper"
  (footprint 36215090-ab6d-45fe-9ffd-693fe09d4aa2)
  (pad d4907411-cbd5-4ce2-838a-82de885d718c)
 )
```

The child nodes depend on the kind of message (to make them unique), some kinds do not have children at all.

## Open Questions

Not 100% sure about how to format approved messages in the file format. Currently I implemented the identifier as a PascalCase string:

```scheme
 (approval "PadOriginOutsideCopper")
```

But I wonder if one of the following formatting would look nicer:

```scheme
 (approval "pad origin outside copper")
 (approval "pad-origin-outside-copper")
 (approval pad_origin_outside_copper)
 (approval PadOriginOutsideCopper)
```

Any opinions welcome :slightly_smiling_face:

### DIFF
--- a/apps/librepcb-cli/commandlineinterface.cpp
+++ b/apps/librepcb-cli/commandlineinterface.cpp
@@ -185,6 +185,10 @@ int CommandLineInterface::execute() noexcept {
       "all",
       tr("Perform the selected action(s) on all elements contained in "
          "the opened library."));
+  QCommandLineOption libCheckOption(
+      "check",
+      tr("Run the library element check, print all non-approved messages and "
+         "report failure (exit code = 1) if there are non-approved messages."));
   QCommandLineOption libSaveOption(
       "save",
       tr("Save library (and contained elements if '--all' is given) "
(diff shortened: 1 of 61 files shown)


ubruhin:
Whoops, PR created on the wrong branch :sob: :joy:

dbrgn:
With GitLab you can edit the target branch after creating the MR. Isn't that possible on GitHub? 🙂

ubruhin:
> With GitLab you can edit the target branch after creating the MR. Isn't that possible on GitHub? slightly_smiling_face

Actually I don't know, I realized my bad only *after* deleting the old branch since I was surprised that the PR got closed :sob: Didn't look for such an option then, just opened a new one...

PR #1107, opened 2023-02-20
Support dismissing library check messages

ubruhin:
## Summary

Support dismissing library check messages, which is then stored in the corresponding files to make it persistent (see #1006). I called it "approvals". This feature is a requirement to allow running the library checks from the CLI (#1007).

## User Interface

image

Approved messages are currently not removed from the list to allow reviewing them easily and usually the list is not flooded by lots of messages anyway.

## File Format

*Updated to the final implementation.*

Within each library elements S-Expression file, new nodes are added for each approved message:

```scheme
 (approved pad_origin_outside_copper
  (footprint 36215090-ab6d-45fe-9ffd-693fe09d4aa2)
  (pad d4907411-cbd5-4ce2-838a-82de885d718c)
 )
```

The child nodes depend on the kind of message (to make them unique), some kinds do not have children at all.

## Open Questions

Not 100% sure about how to format approved messages in the file format. Currently I implemented the identifier as a PascalCase string:

```scheme
 (approval "PadOriginOutsideCopper")
```

But I wonder if one of the following formatting would look nicer:

```scheme
 (approval "pad origin outside copper")
 (approval "pad-origin-outside-copper")
 (approval pad_origin_outside_copper)
 (approval PadOriginOutsideCopper)
```

Any opinions welcome :slightly_smiling_face:

### DIFF
--- a/libs/librepcb/core/library/cat/librarycategory.cpp
+++ b/libs/librepcb/core/library/cat/librarycategory.cpp
@@ -64,6 +64,8 @@ void LibraryCategory::serialize(SExpression& root) const {
   root.ensureLineBreak();
   root.appendChild("parent", mParentUuid);
   root.ensureLineBreak();
+  serializeMessageApprovals(root);
+  root.ensureLineBreak();
 }
 
 /*******************************************************************************
--- a/libs/librepcb/core/library/cmp/component.cpp
+++ b/libs/librepcb/core/library/cmp/component.cpp
@@ -162,6 +162,8 @@ void Component::serialize(SExpression& root) const {
   root.ensureLineBreak();
   mSymbolVariants.serialize(root);
   root.ensureLineBreak();
+  serializeMessageApprovals(root);
+  root.ensureLineBreak();
 }
 
 /*******************************************************************************
--- a/libs/librepcb/core/library/cmp/msg/msgduplicatesignalname.cpp
+++ b/libs/librepcb/core/library/cmp/msg/msgduplicatesignalname.cpp
@@ -43,7 +43,9 @@ MsgDuplicateSignalName::MsgDuplicateSignalName(
            "has several pins which are electrically exactly equal (e.g. "
            "multiple GND pins), you should add only one of these pins as a "
            "component signal. The assignment to multiple pins should be done "
-           "in the device editor instead.")) {
+           "in the device editor instead."),
+        "duplicate_signal_name") {
+  mApproval.appendChild("name", *signal.getName());
 }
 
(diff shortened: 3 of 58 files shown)


dbrgn:
Some quick bikeshedding thoughts (which I'll voice now since it also affects the file format, and not just the UI): I think I'd prefer the terminology "ignore" or "allow" over "approve", since that's what seems to be used by almost every linting tool. Approval sounds like part of a normal workflow ("there are some issues that need to be approved") as opposed to potential problems that need to be verified/ignored ("there are some warnings that may be ignored", or "there are some rule violations that may be allowed").

Unfortunately in the UI there's no good and simple icon for ignoring. But similar to "Fix" the word "Ignore" could be used. (Or just keep the checkmark and add an appropriate title text.)

ubruhin:
Thanks for your thoughts. Actually I was thinking about the term "dismiss", what do you think about that? Somehow I don't like "allow" and even "ignore", but I agree "approve" might not be ideal either.

The main reason why I decided against "dismiss" is that for me it doesn't feel consistent to the rest of the file format as it is not a noun:

```scheme
 (footprint ...)
 (pad ...)
 (dismiss "PadOriginOutsideCopper")
 (dismissed "PadOriginOutsideCopper")
```

OK it seems that "dismissal" seems to be the corresponding noun but somehow I don't feel comfortable with that either :thinking:

Same question would be for "ignore" and "allow" - how to actually name it in the file format? In the UI I don't care too much about the terminology, I think any of these terms would be clear for the user.

ubruhin:
> I think I'd prefer the terminology "ignore" or "allow" over "approve", since that's what seems to be used by almost every linting tool. 

After thinking some more about this, I think a linting tool is a quite different thing. For a linter, you indeed do ignore (or suppress) some messages, i.e. these messages are not generated and the user will never see them.

However, an ERC/DRC/library check is different. The messages are produced anyway, you just mark them as "okay" afterwards. And even then, they don't completely disappear, they are still accessible in the UI.

Given this difference, IMHO "approve" is actually not too wrong in this case as it is indeed some kind of approval process :see_no_evil:

ubruhin:
OK thanks for your feedback :+1: I don't have a strong opinion either but so far I think my favorite is "approval" with snake_case identifier. Or maybe "approved" since this term is already used in the file format of the ERC :thinking: 

```scheme
(approved pad_origin_outside_copper)
```

> By the way, another option for the file format would be to group these approvals or dismissed warnings:

Would be possible, but usually I don't do grouping in the file format because it adds unnecessary "noise" to the files (especially if there are no approvals, there would still be an empty group then).

PR #1111, opened 2023-02-24
Add stop mask opening to non-plated holes

ubruhin:
To avoid solder resist flowing into non-plated holes, this PR adds a stop mask opening to each hole (on both board sides). The expansion value is calculated from the board design rules, the same way as already done for vias and THT pads.

Eagle (at least some older version) does it exactly the same way. KiCad seems to add the stop mask by default as well, but they can optionally be disabled. In addition, as far as I can see the default expansion offset is zero, i.e. no clearance between the drill and the stop mask opening. Making it configurable would be nice for LibrePCB as well (maybe I'll implement that later) but I think a default expansion >0 is better that no expansion...

Actually this does not affect the file format, but I still consider it a breaking change since it has quite some impact on the generated Gerber files. Thus a file format upgrade message is shown to inform the user about this change.


ubruhin:
Update: Made the stop mask even configurable now in the hole properties dialog.

image

PR #1113, opened 2023-02-26
Board editor: Merge all board settings into one dialog

### LINKED ISSUE #1095: Board editor: no option to rename an existing board
LibrePCB Version: 0.1.7
Git Revision:     ba553c6fb
Build Date:       2022-10-03T08:32:18
Qt Version:       5.15.6 (built against 5.15.6)
CPU Architecture: x86_64
Operating System: macOS 12.6
Platform Plugin:  cocoa
TLS Library:      Secure Transport, macOS 12.6

##### SUMMARY
While you can rename a schematic sheet after creation there is no option to rename an existing board.

ubruhin:
### Description 

Merging two existing board editor dialogs into a new dialog called "Board Setup":
- Board Layer Stack Setup
- Board Design Rules

In future, DRC settings will be added to this dialog as well, and probably many more settings. With a single dialog, the UI concept scales much better than having a separate dialog for each kind of settings.

Also clarified some things in the design rules tab to help understanding these settings better.

Fixes #1095 since the board name can now be edited in this dialog.

### Screenshots

image

image


hooperbloob:
Looking at how the design rules are represented in the code.  Have you considered making them first class objects instead of discrete fields managed individually? You could have scalar rules, ratio rules, etc. that have intrinsic limits, labels, tooltips, equality checks, precedence preferences, etc.  UIs could then work with collections of them as a group, laying out their preferred widgets in the preferred order.  

Looks like Qt uses an XML file for widget placement. If you specify a parent container widget maybe your code could then insert the design rule widgets dynamically within it. If the rules had classifications they could be parked under headings of that name. Clicking on that heading could then hide/show them.  Adding/revising the rules then gets a lot easier.

Just my $0.02 worth...

hooperbloob:
It also then provides a natural entry point for allowing user-specified rules in the future.

ubruhin:
> Have you considered making them first class objects instead of discrete fields managed individually? You could have scalar rules, ratio rules, etc. that have intrinsic limits, labels, tooltips, equality checks, precedence preferences, etc. UIs could then work with collections of them as a group, laying out their preferred widgets in the preferred order.

I used such a concept at other places (e.g. workspace settings) in the past. But it turned out that it makes things way more complicated than necessary, and the generated UI is never as nice as when created by hand. So in the end I refactored everything to get rid of this complicated design and so far I'm still very happy with this decision.

For that reason, I prefer to keep the current, simple concept for the design rules as long as a generic concept is not necessary. But thanks anyway for your input! :slightly_smiling_face: Maybe some day a more generic concept is necessary indeed, but we can still refactor it later when really needed.

PR #1114, opened 2023-02-27
Update/extend information shown in about dialog

ubruhin:
### Summary

It turned out that for many bug reports it's important to know how LibrePCB was installed (AppImage? Snap? Flatpak?) since various bugs are related to packaging. So I included this information in the "About LibrePCB" details page, together with some general UI improvements.

### Support compiling author information into the executable

Allow to pass `-DLIBREPCB_BUILD_AUTHOR="..."` to `cmake` to embed information about the **build**. Our CI now sets this variable to "LibrePCB CI" so whenever this is shown in the about dialog, we know it was a binary built by ourselves (in contrast to distribution packages). If the variable is not set, it won't be displayed in the about dialog.

### Detect and display runtime environment

In addition to the build environment, the **runtime environment** is also of interest. For example our official installer, AppImage and *.tar.gz binaries are all containing the same binaries (so `LIBREPCB_BUILD_AUTHOR` is identical for all of them), but the AppImage is run in a quite different, sandboxed environment.

The runtime environment is detected by the following environment variables and shown in the about dialog if not empty:

- `LIBREPCB_RUNTIME`: Allows to manually override the runtime detection with a custom value. Useful e.g. when LibrePCB is started by a wrapper shell script from some packaging/sandboxing system.
- `SNAP`: If set, runtime is detected as "Snap".
- `FLATPAK_ID`: If set, runtime is detected as "Flatpak".
- `APPIMAGE`: If set, runtime is detected as "AppImage".

### Major UI rework to fix issues and update content

- Open hyperlinks in the web browser configured in the workspace settings.
- Fix too large dialog window size by merging many wordwrapping QLabels into a single, large QLabel.
- Make layout of dialog header more compact (logo beside text).
- Generally update/improve information shown in the about dialog.

### Screenshots

| Old | New |
|---|---|
| image | image |
| image | image |
| image | image |

### DIFF
--- a/ci/build_librepcb.sh
+++ b/ci/build_librepcb.sh
@@ -27,7 +27,8 @@ echo "Using CC=$CC"
 mkdir -p build && pushd build
 cmake .. \
   -DCMAKE_INSTALL_PREFIX=$(pwd)/install/opt \
-  -DBUILD_DISALLOW_WARNINGS=1
+  -DBUILD_DISALLOW_WARNINGS=1 \
+  -DLIBREPCB_BUILD_AUTHOR="LibrePCB CI"
 VERBOSE=1 $MAKE -j8
 $MAKE install
 popd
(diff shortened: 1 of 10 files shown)


dbrgn:
👌🏻

PR #1146, opened 2023-04-08
Linux: Support registering *.lpp file extension at runtime

### LINKED ISSUE #1058: Register *.lpp files from within the application
##### SUMMARY

Currently the `*.lpp` file extension is registered with the LibrePCB application only when using the installer on Windows or Linux. But with the portable packages (e.g. the AppImage) this is not the case. That's the reason why the installer is promoted as the recommended installation method, although this is quite unusual on Linux.

##### SOLUTION

It would be nice if the application itself can (optionally) register the file extension `*.lpp`. Then the portable packages would be much more useful.

However, there should be many things to consider carefully:
* The application might run with user rights, possibly not able to register a file extension (?)
* If the installation method already registered the file extension (maybe still done in the installer? maybe done by a Linux package management like Snap/Flatpak?) the option to register the file extension from within the application should be disabled.
* Registering the file extension shall be optional. But where to put the option?
  * In the wizard which appears when starting LibrePCB the first time?
  * In the workspace settings dialog?
  * As a (dismissable) popup in the control panel?
 * Somewhere also provide an option to unregister the extension.
 * Can we somehow also provide this feature on macOS?

Once this is implemented (and working with the AppImage), probably we should not promote the installer for Linux anymore.

ubruhin:
This feature allows to (optionally) integrate portable Linux packages (mainly the AppImage) into the desktop environment by installing the mime types & *.desktop file. This has always been a big drawback of our AppImage, which is finally fixed.

The option is available in the control panel (if applicable) and in the workspace settings dialog:

image

To avoid feeling it like "black magic", a dialog lists all files which will be created or deleted by this feature.

Note that currently this feature is only available in our official Linux binaries. For Windows it's not implemented (yet) and for MacOS it's not needed. Even for Linux it's only enabled in our official binaries - for any other distribution method (including Snap and Flatpak) this feature is disabled since such packages usually already take care of the registration.

Closes #1058


dbrgn:
Nice!

Issue #648, opened 2020-01-04
Do app updates from within the app


spookydd:
Perhaps in the preferences popup in an extra tab, or could be an option in a pull down menu (help maybe): A checkbox to allow automatic search for app updates, and offer updating now if there is one available.

spookydd:
I just did an update of libreoffice, which should be similar in many ways to librepcb, at least for its UI behaviors and a few other things.
They do the automatic app updates from within the app and perhaps there is a way to speed things up by looking at how they did it there. Looking into their code. Copying things from one open source project to an other should be totally ok right?

ubruhin:
Yeah I think that would be a nice feature, but of course only if LibrePCB was installed with the installer. For other deployment methods (bundle, AppImage, ZIP, AUR, ...) we can't do an automatic update and thus the update feature needs to be disabled.

For the record, the Qt installer framework already supports this use case: LibrePCB just needs to execute `librepcb-maintenance --checkupdates` to check if updates are available. If there are updates, executing `librepcb-maintenance --updater` opens the updater wizard.

But adding the configuration option to workspace settings require to change the settings file format, thus this needs to wait for the 0.2.x release.

PR #1147, opened 2023-04-08
Stop providing installers for Linux and MacOS

### LINKED ISSUE #894: macOS Installer: Wrong application bundle format
With the current installer (both 0.1.5 stable and the version built by the `cmake` branch), the installer does not use the proper application bundle format for macOS.

This is how it looks:

image

I'm not really familiar with macOS packaging, but this should not be a directory containing shared files and a binary, but instead it should be a self-contained application bundle that can be double-clicked directly. It works, but it's not nice.

The drag&drop dmg app bundle on the other hand works properly and shows a "librepcb"-entry with icon in the applications directory.

ubruhin:
This is not an easy decision, but given the facts that:

- Installers are quite unusual on Linux and MacOS
- The installer on MacOS creates quite a mess in the launcher (#894)
- The AppImage is now able to install mime types and *.desktop files (#1146)
- There are plenty of other installation methods on Linux (Snap, Flatpak, distribution packages, ...)

I think it makes no longer sense to maintain the installers for Linux and MacOS. The main drawback is that with the portable packages we don't have an update mechanism (while the installers *do*) but I think that's not a big issue and for long term we can fix this by an own update mechanism (#648).

For people currently using these installers, it means they should uninstall LibrePCB and then switch to a different installation method (will mention this in the release notes).

If somebody doesn't agree, please let me know ;)

Closes #894


dbrgn:
I absolutely do agree 🙂

PR #1156, opened 2023-04-29
Implement 3D viewer & STEP export

ubruhin:
# Summary

Finally this PR adds initial 3D support to LibrePCB! :rocket: See also the related [forum thread](https://librepcb.discourse.group/t/3d-models-of-packages-and-boards/).

## Dependencies

We now need additional dependencies to compile & run LibrePCB:

- [OpenCascade](https://www.opencascade.com/) for reading and writing STEP files. Both the official OCCT and the open-source edition OCE are compatible.
- [OpenGL Utility Library](https://en.wikipedia.org/wiki/OpenGL_Utility_Library) (GLU) for tesselating surfaces in the 3D viewer.
- And OpenGL for rendering.

These libraries cannot easily be bundled with the LibrePCB sources, thus for compiling they need to be installed on the system (fortunately they are available on most platforms). Due to dynamic linking they also need to be installed at runtime, but in our official binaries they are bundled (except OpenGL, but this is an "optional linking" anyway, managed by Qt).

However, since there *might* be troubles getting these libraries on some platforms, LibrePCB can optionally be compiled without them - the 3D features will simply not be available then. In addition, all these libraries are not accessed at all unless the user explicitly requests some 3D functionality. This should keep the rest of LibrePCB working fine in case these libraries cause any problem at runtime (e.g. due to an incompatible system library or so).

## 3D Package Models

The 3D representation of packages is implemented with **STEP files** stored in the packages directory. Each package may contain multiple such 3D models. In addition, each footprint specifies a list of 3D models to be usable for the corresponding footprint. Example: A TO220 package might be mounted vertically or horizontally and thus has these two footprints, while each of the footprint specifies a different 3D model to be compatible with.

In addition, each footprint specifies placement information for the STEP model (translation/rotation) in case the STEP file has a different origin & orientation than the footprint.

The package editor now is switchable between 2D footprint editing and 3D model placement:

librepcb-3d-package

## Package Check

The package editor shows an info message if there are footprints with no 3D model assigned, to remind the user adding one.

## 3D Board Viewer

The board editor now also allows to switch between 2D and 3D mode, or a split view with both. In addition, it is possible to switch between the different 3D models if a package provides multiple. The board properties as implemented in #1154 are also taken into account for rendering. The 3D preview is automatically rebuilt in a separate thread after each modification so it is kind of "pseudo-live".

librepcb-3d-board

## Board STEP Export

The board editor also allows to export the PCB including all devices as a STEP file. Only basic functionality for now, no configuration options yet but I think this is sufficient for the typical use-case.

librepcb-step-export

## CLI

The size of STEP files will sooner or later bloat our libraries heavily, thus I implemented a STEP file "minifier" according [this idea](https://gitlab.com/sethhillbrand/stepreduce). The GUI performs that operation automatically when loading a STEP file, but for the `librepcb-parts-generator` I think it will be useful to automate this process with the CLI. Therefore I implemented a new flag `--minify-step` which minifies all package STEP files within a library. 

## Notes

Note that of course the value of this feature depends on the availability of 3D models in the packages. Unless we extend our official libraries with 3D models, the 3D viewer will just display the PCB without devices. So far I do not intend to implement a "workaround" for users to allow adding 3D models to projects directly, i.e. without waiting for 3D models in our official libraries, or to use alternate 3D models. Probably this would be useful in many cases, but it should be doable without a file format change so I'd postpone this to a minor release.

Also be aware that in total this is a quite complex feature (especially the interaction with OpenCascade and OpenGL), thus I expect there are still various bugs or incompatibilities (but of course I already tested a lot).


xotho:
About 3D view, I suggest to add possibly a corner line in black that show the "face" presence of a 3D object and not a FLAT SOLID color. I think this can be good.

PR #1159, opened 2023-05-04
Board editor: Auto-rebuild planes asynchronously

ubruhin:
## Summary

So far, planes had to be rebuilt manually by triggering the corresponding tool button. And then the rebuild blocked the UI thread, which could be several seconds for a large board. Both was quite annoying.

Therefore this PR improves the plane rebuild in various ways:

- The rebuild is done in a separate thread to keep the UI responsive. A "waiting spinner" in the graphics view provides visual feedback about the activity in the worker thread.
- Whenever possible, only relevant planes are rebuilt. Planes on layers which are hidden or not modified since the last rebuild are simply skipped to speed up the rebuild.
- If all the following conditions are met, the rebuild is started automatically:
  - There is no command/tool active OR there was no input event (e.g. cursor movement) for at least 700ms.
  - The last plane rebuild was at least 1000ms in the past (i.e. a second pause between rebuilds).
  - The board editor window is in the foreground OR is receiving any user activity.

These conditions ensure that the rebuild is not triggered too often as it is quite CPU intensive (although asynchronous, too much CPU load is still annoying for example due to laptop battery usage or fan noise). Even with these restrictions, the rebuild runs often enough that a manual rebuild should not be needed anymore so I removed the corresponding button from the toolbar (it is still in the main menu to enfore rebuilding *all* planes, for example in case the partial rebuild skipped planes by mistake).

In addition to these usability improvements, the planes now also create cut-outs for arbitrary polygons and stroke texts on copper layers (previously these objects were ignored and thus possibly flooded/overlapped by planes).

## UI

The 700ms idle-time detection makes it very intuitive to draw traces within a plane - just stop mouse movement for a moment to get feedback about the plane clearance:

librepcb-planes

## Notes

I'm not sure if it would make sense to allow disabling these auto-rebuilds by some configuration option (for example to reduce battery usage on a laptop). At the moment I think it's not needed (I don't like unnecessary configuration options as they make things unnecessary complex), especially because the auto-rebuild can be disabled implicitly by hiding all planes.


dbrgn:
> I'm not sure if it would make sense to allow disabling these auto-rebuilds by some configuration option (for example to reduce battery usage on a laptop).

I don't think a _specific_ option is needed, but a generic "power save mode" might be interesting for some people (not for me). I think IntelliJ also has something like that. It would need to show a rebuild button though, when enabled.

However, I'd wait with such a feature until you get feature requests for it, from people who have a real need for the feature.

ubruhin:
> I don't think a specific option is needed, but a generic "power save mode" might be interesting for some people (not for me). I think IntelliJ also has something like that. It would need to show a rebuild button though, when enabled.

Sounds reasonable, there are more and more such asynchronous task which could all made dependend on such a power save mode :+1:

> However, I'd wait with such a feature until you get feature requests for it, from people who have a real need for the feature.

I agree, would be easy to provide such a feature in a later minor release.

PR #1166, opened 2023-05-30
Rename "top/bot placement" layers to "top/bot legend"

ubruhin:
**EDIT**: Renamed to *legend*.

---

Not sure yet if this really makes sense (any opinions welcome!):

So far the layers "Top Placement" and "Bottom Placement" were intended for any placement guide drawings, like pin-1 markings or package alignment lines. Typically these drawings are then used for the silkscreen, but actually it is configurable in the Gerber export what layers shall be put on silkscreen. Due to this conceptual separation, I did not name that layer "silkscreen". I wanted to name the layers according their *content* (or *function*), not according their physical usage.

However, I think there was already a lot of confusion between the placement- and documentation layers. In addition, it's not intuitive what layer should be used to put e.g. a logo or a custom text on the silkscreen since these do not feel to be related to the "placement" layer (there's nothing to place for a logo).

I wonder if it would be more intuitive to rename the placement layers simply to "Top/Bottom Silkscreen", so I opened this PR to do this renaming. Even though this way it does not indicate anymore what *content* to place on that layer, at least it's more clear what *happens* with the content on that layer...


ubruhin:
An alternative would be to call these layers "Top Legend" and "Bottom Legend" since IPC-7351C seems to use this terminology :thinking: Might be better than "placement" while still clearly differentiate from the physical layers...

dbrgn:
I think a dedicated silkscreen layer would be simpler conceptually, than multiple silkscreen layers. The concept of having placement and documentation layers is nice, but I think in practice it only adds unnecessary complexity...

But maybe that depends on the type of workflow. Professional users might have different requirements. (But those users could still create additional silkscreen layers that can be toggled on/off.)

Can we already assign roles to arbitrary layers? Maybe that might be good: A default single silkscreen layer, but the possibility to add other layers with the "silkscreen" role.

ubruhin:
> I think a dedicated silkscreen layer would be simpler conceptually, than multiple silkscreen layers. The concept of having placement and documentation layers is nice, but I think in practice it only adds unnecessary complexity...

Hm, I don't understand your point. Basically we have only one silkscreen layer, and I don't want to change that. The documentation layer is not intended for silkscreen and thus cannot be put on silkscreen. To be exact, we do have two more layers *names* and *values* which can be put on silkscreen, but these are intended only for texts, and that's needed to allow each of the typical use-cases:

- Either no silkscreen at all (to reduce costs)
- Only placement polygons (no space for texts)
- Placement polygons + component names (e.g. "R1") for manual assembly
- Placement polygons + component names + component values (e.g. "100nF") for complete assembly information

The desired use-case can currently be selected by enabling the corresponding layers for silkscreen:

image

> But maybe that depends on the type of workflow. Professional users might have different requirements. (But those users could still create additional silkscreen layers that can be toggled on/off.)

Hm, that's not possible currently and I don't intend to change that. I think it should be good enough to have these three layers (only one of them for polygons), so every user has to use exactly these layers for silkscreen.

> Can we already assign roles to arbitrary layers? Maybe that might be good: A default single silkscreen layer, but the possibility to add other layers with the "silkscreen" role.

I'm not sure if that makes sense. For the library we need a very strict concept of layers (without any customization) anyway, otherwise the whole idea of sharing libraries is broken. And just for boards customization is not needed since each board uses only a single workflow, so a single layer for silkscreen should be enough.

By the way, the reason why I'm not sure about naming the layer "silkscreen" is because this layer might *not* actually appear on silkscreen, since this can be toggled on/off per board. Also it looks very strange to enable silkscreen for silkscreen:

image

dbrgn:
Ah, yeah, I mixed up layers and name/value labels.

Legend is probably better than Placement. I don't have a strong opinion on this topic at the moment though.

ubruhin:
Alright, I renamed it to "legend" now. I also think it's a bit better than "placement" and it doesn't introduce the confusion of "silkscreen" so it might be a good compromise...

Issue #965, opened 2022-04-07
Allow selecting devices for BOM export (assembly variants)


ubruhin:
##### SUMMARY

Currently the BOM contains all devices on the board. But sometimes there are devices which do not represent a part to order (example: The LibrePCB Logo device) so these should not be contained in the BOM.

##### SOLUTION

The library component and/or device should have an attribute for this property, which the BOM export then should respect.

Should be considered together with the implementation of an MPN attribute and assembly variants, since these topics are related to each other.

ubruhin:
For the record: As suggested in https://github.com/LibrePCB/LibrePCB/issues/1001#issuecomment-1464995220, it should also be possible to make assembly printouts for each assembly variant, i.e. containing only the devices to be assembled.

PR #1177, opened 2023-07-11
Library: Support adding part numbers to devices

### LINKED ISSUE #653: Add manufacturer name field to library elements
### SUMMARY

As discussed in https://librepcb.discourse.group/t/naming-of-packages/167/, the manufacturer field could be included in the library file format for the following entities:

- Symbol
- Component
- Package
- Device

##### SOLUTION

- The field should be optional
- This requires a file format change, so it should probably be part of the v0.2 release
- The field can be used for filtering (e.g. in the "add component" dialog), but it should probably not be shown in the schematics or board editor by default

### LINKED ISSUE #782: Library structure not easy to be used in a real industry
This is difficult to build an individual "unchangable" (released) component, e.g. 
R=10k_T=5%_P=0.25W_Pk=R-THT-0207_Raster=7.62mm

including all production information (manufacturer, partnumber, manufacturer datasheet link, PCBA (internal) datasheet, PCBA THT bending type number) and so one. 

For the industry the "component" (component released inside the industry system) is something completly unchangable (no any other raster, no any other bending type, no any other notheing is allowed)

Currently (current LibrePCB library concept) the only one way is to put such kind of information into "component" element:

https://i.imgur.com/sQB7GW5.png 
https://i.imgur.com/sAJ0nPA.png

But this is not consistent into general library idea.

Just look into official one library, where the "components" are just an "type of components" instead of just one lingle, released @system component:

https://i.imgur.com/f5AZIRq.png

And if you try to think about a "device" as fully defined component, this is impossible to introduce necessary data (e.g. manufacturer, partnumber, manufacturer datasheet link, PCBA (internal) datasheet, PCBA THT bending type number) and so one).

https://i.imgur.com/nBtngqN.png

##### SUMMARY
Please make the library concept usable for real PCB industry (or just for proffessional hobbist usage). Let's avoid to create next "library monster", unusable and living it's own live). Let's follow the real industry way. 

##### SOLUTION
Please modify in a reasonable way or introduce into library the group for real, released, unchangable, fully defined components (e.g. assembled from all other groups, but finally unconfigurable, just possible to use and notheing more).

ubruhin:
## Summary

This is the first step to implement support for manufacturer part numbers (MPNs). In the end, MPNs are needed to make BOMs much more accurate (and thus more useful) than they are today. This PR extends the library structure with the following features:

- Support adding arbitrary attributes to devices (identical to how it is already possible for components).
- Support adding new objects called *parts* to devices. Parts represent the orderable variants of a device and are uniquely identified by an MPN (the part number) and optionally the name of the manufacturer. In addition, parts can contain arbitrary attributes as well, for example to specify the resistance of a resistor. This way it is possible to represent a whole resistor family with a single device - each resistance (with its MPN) is added as a separate part within the device.
- Add a new (optional) property *manufacturer* to libraries. This property is not critical at all, it is only used for convenience to initialize the *manufacturer* property of new parts with this value. For example within the library "Texas Instruments", you never need to specify the manufacturer for parts since it is automatically taken from the library.

Example structure of a device with parts:

- Device "R-0603":
  - Part "RC1608F102CS" from manufacturer "Samsung":
    - Attribute "RESISTANCE=1kΩ"
  - Part "RC1608F103CS" from manufacturer "Samsung":
    - Attribute "RESISTANCE=10kΩ"
  - Part "RC1608F104CS" from manufacturer "Samsung":
    - Attribute "RESISTANCE=100kΩ"

Note that the usage of parts in projects (for the BOM) will be implemented in a separate PR. This PR is only about the extension of the library structure.

## User Interface

The device editor UI has been rearranged a bit to get some space for the new attributes and part editors. The part editor is now at the bottom left, and the attribute editor at the bottom right. Due to limited space, the attribute editor is context dependent - if a part is selected, the part attributes are shown; if no part is selected, the (common) device attributes are shown. It's not very intuitive unfortunately, but to get v1.0.0 finished sooner I'd improve that in a later minor release.

image

## File Format

Representation of the parts from the screenshot above:

```scheme
 (part "BAT54" (manufacturer "ON Semiconductor")
  (attribute "PACKING" (type string) (unit none) (value "3k T&R"))
 )
 (part "BAT54_D87Z" (manufacturer "ON Semiconductor")
  (attribute "PACKING" (type string) (unit none) (value "10k T&R"))
 )
```

## Concept Decisions

Unfortunately there are many different workflows and requirements imaginable regarding MPN management. So it is not easy at all to find a good concept how to implement this feature. For the record, here are some of my thoughts why I ended up with this concept.

#### Alternative 1: Parts as Separate Library Elements

First of all, a different way of implementation would be to make parts completely standalone library elements which reference a device by its UUID (like devices are standalone library elements referencing components & packages). This would have the advantage to easily extend devices from official libraries (which are installed read-only) because you could extend them with parts stored in your own local libraries. However, it would also have several disadvantages:

- Even more complex overall library structure
- More dependencies generally lead to more issues (e.g. deleting a device would break all parts referencing it)
- Much more overhead (number of files & file size) leading to much higher disk usage, longer library download times and longer library scanning time

#### Alternative 2: All Parts in a Single File

The overhead could be reduced by storing all parts of a library within a single file, but this would be quite inconsistent to the overall library concept and also isn't very practical for version control.

#### Alternative 3: Each Device is a Part

Another way would be to simply consider the device name as an MPN - basically this means we don't even need to change the library structure so this concept would not introduce any additional complexity at all. However, it means that a separate device needs to be created for every single MPN. Although this would be fine for most devices, devices like resistors would extremely bloat the library since there exist thousands of resistor MPNs.

#### Summary

So in the end I decided to store parts within their devices, as a compromise between flexibility, complexity and efficiency. But I'd appreciate any opinion/thoughts about this decision :slightly_smiling_face: 

#### SKU Handling

Another big question is whether we also need to store SKUs (stock keeping units), i.e. the corresponding order numbers of suppliers. For now I decided to *not* store this information at all because this would be an additional huge amount of data, which is even harder to maintain than MPNs. Maybe it makes more sense to allow the schematic editor to query this information from services like Octopart (with the drawback that this only works with Internet connection). As an alternative, users might also use attributes of parts to add this information to the library anyway (e.g. "SUPPLIER=DigiKey; SKU=1276-3484-1-ND"), although this wouldn't be possible for official LibrePCB libraries.

So for now I'd completely ignore SKUs (at least for the next major release of LibrePCB), but of course we could re-think about this topic at any time. And any ideas are welcome of course.

## Related Issues

- Closes #782
- Closes #653
- Required for #965

### DIFF
--- a/libs/librepcb/core/CMakeLists.txt
+++ b/libs/librepcb/core/CMakeLists.txt
@@ -214,6 +214,8 @@ add_library(
   library/dev/devicecheckmessages.h
   library/dev/devicepadsignalmap.cpp
   library/dev/devicepadsignalmap.h
+  library/dev/part.cpp
+  library/dev/part.h
   library/library.cpp
   library/library.h
   library/librarybaseelement.cpp
--- a/libs/librepcb/core/CMakeLists.txt
+++ b/libs/librepcb/core/CMakeLists.txt
@@ -411,6 +411,7 @@ add_library(
   types/ratio.h
   types/signalrole.cpp
   types/signalrole.h
+  types/simplestring.h
   types/stroketextspacing.cpp
   types/stroketextspacing.h
   types/uuid.cpp
--- a/libs/librepcb/core/library/dev/device.cpp
+++ b/libs/librepcb/core/library/dev/device.cpp
@@ -44,8 +44,8 @@ Device::Device(const Uuid& uuid, const Version& version, const QString& author,
                    author, name_en_US, description_en_US, keywords_en_US),
     mComponentUuid(component),
     mPackageUuid(package),
-    mAttributes(),
-    mPadSignalMap() {
+    mPadSignalMap(),
+    mAttributes() {
 }
 
 Device::Device(std::unique_ptr<TransactionalDirectory> directory,
@@ -54,8 +54,8 @@ Device::Device(std::unique_ptr<TransactionalDirectory> directory,
                    std::move(directory), root),
     mComponentUuid(deserialize<Uuid>(root.getChild("component/@0"))),
     mPackageUuid(deserialize<Uuid>(root.getChild("package/@0"))),
-    mAttributes(root),
-    mPadSignalMap(root) {
+    mPadSignalMap(root),
+    mAttributes(root) {
 }
 
 Device::~Device() noexcept {
@@ -120,9 +120,9 @@ void Device::serialize(SExpression& root) const {
   root.ensureLineBreak();
   root.appendChild("package", mPackageUuid);
   root.ensureLineBreak();
-  mAttributes.serialize(root);
-  root.ensureLineBreak();
   mPadSignalMap.sortedByUuid().serialize(root);
+  root.ensureLineBreak();
+  mAttributes.serialize(root);
   root.ensureLineBreak();
   serializeMessageApprovals(root);
   root.ensureLineBreak();
--- a/libs/librepcb/core/library/dev/device.cpp
+++ b/libs/librepcb/core/library/dev/device.cpp
@@ -45,7 +45,8 @@ Device::Device(const Uuid& uuid, const Version& version, const QString& author,
     mComponentUuid(component),
     mPackageUuid(package),
     mPadSignalMap(),
-    mAttributes() {
+    mAttributes(),
+    mParts() {
 }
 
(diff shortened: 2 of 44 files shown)


dbrgn:
Nice, that sounds reasonable!

PR #1180, opened 2023-07-20
Support part numbers & assembly variants in projects

### LINKED ISSUE #965: Allow selecting devices for BOM export (assembly variants)

ubruhin:
# Summary

This is the follow-up of #1177 to make the part numbers from the library now usable in projects, including advanced assembly variant management to cover various use-cases.

## Specifying MPNs in the Schematic Editor

Until now, LibrePCB projects had no official support for manufacturer part numbers (MPNs), thus BOM/PnP exports were not accurate enough for ordering the parts. Now the "Add Component" dialog in the schematic editor allows to add a particular part instead of just a device or component:

image

If a part is added to the schematic, the MPN will be stored in the project and used for the BOM/PnP export. 

However, it's still possible to add just plain devices or components to the schematic, then an MPN could still be specified later in the design phase. This can be done in the component properties dialog:

image

Unfortunately, this new widget might be a bit confusing/complex until you know how it works. I hope we can make it a bit simpler and more intuitive some day. But let's first get some experience whether the underlying concept is working as intended.

## Assembly Variant Management

Many real-world projects are not as simple as placing a 10k resistor on the schematic and having it exported to the BOM. Instead, sometimes there are different variants how the same PCB could be assembled. Maybe one variant with an RS485 interface, and a different variant with RS232 on the same RX/TX lines. Or a DC/DC converter with a fixed output voltage of either 5V or 3.3V, depending on the assembled voltage divider resistors. So the same PCB can be used to produce different products. Therefore you'll also need different BOMs.

All the different assembly variants can now be specified under *Project -> Project Setup -> Assembly Variants*:

image

There must always be at least one assembly variant. Therefore LibrePCB automatically adds a variant called ~~"AV"~~ "Std" for every new project, or when migrating a v0.1 project. If multiple variants are not needed, just don't care about it.

## Omitting Components in BOM/PnP Export

For components which shall be omitted during BOM/PnP export of particular (or all) assembly variants, the corresponding checkboxes (in a pop-up) need to be unchecked:

image

For the simple case of only one assembly variant, the row directly contains a checkbox (no popup) which can be toggled for either "mount" or "do not mount":

image

## Different Parts in Different Assembly Variants

For the DC/DC converter example of above, where different parts need to be assembled by different assembly variants, it is possible to add additional rows to this table and enabling them only for the corresponding assembly variant. Either the MPN could differ then, or it's also possible to specify arbitrary attributes (like resistance):

image

So the BOM for the 5V assembly variant will specify 8.2k for this component, while the BOM for the 3V3 assembly variant specifies 5.6k for the same component.

## 2nd Source Parts

Another real-world requirement nowadays is to specify a second source for each critical part in a project, for example to be prepared for chip shortages. Specifying these second source part numbers outside of the EDA project however is not practical as it needs to be manually kept in sync with the project, and generated BOMs might need to be patched manually which is error-prone and time-consuming.

Therefore LibrePCB simply allows adding more than one part to a component in the schematic. These are then listed as alternative parts:

image

If there are any components with alternative part numbers, they will be exported as additional columns in the BOM (including the "Value", since it can be different for different parts):

image

## BOM/PnP Export Dialogs

Since the mounting of components is now configurable, you now can make mistakes e.g. by omitting components from the BOM by accident. To minimize this risk, both the BOM and PnP export dialogs not only show the components which are exported, but also those *not* contained in the exported files (gray background):

image

# Architecture

Fore those looking for a single UML diagram describing all the mentioned features :grin: 

mpn_management

# Notes / ToDo

- The first column of the parts table (title "Board Device") makes the table a bit complicated, but it's caused by the fact that every MPN is valid only for a particular device (because it specifies the package & pinout). Because a component might be represented by different devices on different boards, the MPNs need to be specified for each of these devices separately. Maybe we could make the UI a bit simpler some day for the case of having only one board.
- Currently the schematic is neither aware of assembly variants nor alternative parts. If a component specifies different parts for different assembly variants, or different alternative part numbers, the schematic just displays the first one of these parts.
- The 3D board viewer is not (yet) aware of assembly variants either, it always shows all components as assembled.
- Although this new feature does introduce quite some complexity, for typical, simple use-cases it's not as complicated as it might look in this PR. If there's only one board and you don't add assembly variants or alternative parts by intention, the parts list usually consists of only one row where you can just type in the MPN (and not even that is mandatory).

# Related Issues

Closes #965


dbrgn:
I wonder if the default name "AV" could be confusing. To me, I immediately think of "Audio/Video" 😄

Maybe "Default Assembly" instead?

ubruhin:
> I wonder if the default name "AV" could be confusing. To me, I immediately think of "Audio/Video" smile
> 
> Maybe "Default Assembly" instead?

The thing is, this name is added to BOM/PnP output file names. Therefore it must be short, without spaces and without special characters.

So I just abbreviated "assembly variant", and initially appended the number "0" to indicate that this is the only assembly variant: "AV0" - but then removed the zero as it might be confusing.

My ideas of assembly variant names:

- `AV`
- `AV0`
- `Std` ("standard" assembly?)
- `Default` (already a bit long, but might be acceptable. However, boards are usually named "default" as well, so there can be cases where the term "default" is contained two times in output filenames).

Maybe `Std` would indeed be a good alternative to `AV`? At least it is not used anywhere else in LibrePCB so it does not confuse with other default objects (board name, netclass name, ...)

doom-fr:
DFLT ?

ubruhin:
Changed it to `Std` for now (but still accepting other suggestions :see_no_evil:). And hiding assembly variants now everywhere in the UI if there's only one:

image

image

dbrgn:
Std is better than AV 🙂 I don't have a better suggestion.

Issue #706, opened 2020-06-01
Display and offer link to datasheet


doom-fr:
##### SUMMARY
Electronics is datasheet, datasheet and datasheet.
But in LibrePCB, there's nothing provided for that.

##### SOLUTION
For each device, we should add datasheet link in the library.
This is spread in the schematic and board editors. Just click to see the datasheet.
For generic component, the field can be manually added in the schematic page.
If not, the user can right click on the field containing the MPN and launch a search in is favorite search engine.

PS : there might be sometimes multiple datasheets for one device. But one is already a great improvement

dbrgn:
This would be useful indeed. See also the extended discussion here: https://github.com/LibrePCB/LibrePCB/issues/653 The datasheet could be implemented as a predefined `DATASHEET` attribute.

ubruhin:
Implemented in #1460. Not 100% finished yet (UI currently only supports one datasheet per device), but the main functionality is there so I'll close the issue.

PR #1181, opened 2023-07-23
Prepare file format for adding resources to library elements

ubruhin:
As suggested in #706, it would be useful to have links to datasheets etc. available right in the schematic editor (e.g. in the context menu of symbols). One option might be to "abuse" the attributes system (e.g. by adding a `DATASHEET` attribute). However, I think it makes sense to implement this as a dedicated feature - especially because many ICs require links to several documents, not just a single datasheet.

This PR does not yet implement this feature, but extends the file format to allow implementing it in a later minor release. I called the documents "resources", but I'm open for suggestions of another naming. In the file format they look like this:

```scheme
(resource "Datasheet" (mediatype "application/pdf")
 (url "https://example.com/datasheet.pdf")
)
```

The `mediatype` key is intended to allow the UI to display a corresponding icon (e.g. a PDF symbol) where useful. In addition, the application might download PDFs directly and open them with the local PDF reader instead of just opening the URL in the webbrowser.

Resources will be supported for the library elements symbols, components, packages and devices.

Once we implement this feature in the UI, we should consider to extend the CLI to validate the URLs used in a library against a domain whitelist since contributing arbitrary URLs could be dangerous. In addition, URLs used in libraries should probably periodically be tested to automatically remove broken URLs since they are very annoying. After all, I don't know if datasheet URLs are usually stable enough for adding them to libraries - if not, we need to reconsider this feature in general.

### DIFF
--- a/libs/librepcb/core/CMakeLists.txt
+++ b/libs/librepcb/core/CMakeLists.txt
@@ -244,6 +244,8 @@ add_library(
   library/pkg/packagemodel.h
   library/pkg/packagepad.cpp
   library/pkg/packagepad.h
+  library/resource.cpp
+  library/resource.h
   library/sym/symbol.cpp
   library/sym/symbol.h
   library/sym/symbolcheck.cpp
--- a/libs/librepcb/core/library/libraryelement.h
+++ b/libs/librepcb/core/library/libraryelement.h
@@ -24,6 +24,7 @@
  *  Includes
  ******************************************************************************/
 #include "librarybaseelement.h"
+#include "resource.h"
 
 #include <QtCore>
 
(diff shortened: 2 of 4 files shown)


dbrgn:
I would rename `type` to `mediatype` (as defined [here](https://www.iana.org/assignments/media-types/media-types.xhtml)).

Additionally, an actual `type` field could be useful, for categories, e.g. "Datasheet", "Mechanical Drawings", etc. I'm not sure how this would be implemented in a best way, maybe with an initial predefined set of entries, and "Other" as fallback?

ubruhin:
> I would rename type to mediatype (as defined [by the IETF and IANA](https://www.iana.org/assignments/media-types/media-types.xhtml)).

Sounds good, thanks for the hint! :+1:

> Additionally, an actual type field could be useful

Actually I had *exactly* the same idea, including the value "Other". But then I was wondering what this information is actually useful for. I ended up with the decision that it leads to additional effort (for specifying it) without a clear usage. If you see a case where this information is useful, please let me know :slightly_smiling_face: 

Note that for the case that we need some kind of a *main* resource (e.g. *the* datasheet link), I'd go with the concept already known from various other objects in LibrePCB - simply considering the *first* entry of a list (the resources in this case) as the most important one.

dbrgn:
The use case I had in mind: If a part has multiple data sheets linked, it's useful to know the difference between them. I could imagine that often this could be: Datasheet / Reference Manual / Mechanical Drawings.

If you just have 3 URLs ending with a cryptic name, that's not very helpful.

ubruhin:
Ah, that's actually what the string after `(resource` is intended for ("Datasheet" in the example above). That's the identifier/name of a particular resource and should just be set to an arbitrary string like "Reference Manual" or "Datasheet". I have in mind to show only this text in a context menu - the URL could be shown in the status bar on hover, just like a hyperlink in a web browser.

dbrgn:
Ahh OK, that makes sense then 🙂

PR #1192, opened 2023-08-07
About dialog: Add credits section

ubruhin:
Add a dedicated credits tab to the about dialog, mentioning NGI0 and icons8 (moved from the 'About' tab):

image


libs/librepcb/editor/dialogs/aboutdialog.cpp
```diff
@@ -146,6 +145,30 @@ AboutDialog::AboutDialog(const WorkspaceSettings& settings,
     mUi->lblContributing->setText(text);
   }
 
+  // Set credits text (not using Qt designer to avoid layouting issues).
+  {
+    QString text;
+    text += "<p>" %
+        tr("This project relies on many contributors, sponsors and other "
```
dbrgn, line 152: You could link "contributors" to https://github.com/LibrePCB/LibrePCB/graphs/contributors 🙂

ubruhin: I didn't link to this site because this link already exists on the first tab and I don't like a redundant link:

image

Not sure if this information should be moved to the "Credits" tab as well, but then I don't know what to write on the first tab :see_no_evil: If you have any suggestion, let me know :)

dbrgn: Ah, true. I don't think it would hurt to link to it twice though.

ubruhin: Okay, added the link :+1:

PR #1195, opened 2023-08-10
Add "generated by" property to library elements

ubruhin:
Just adding the node `(generated_by "")` to symbol/package/component/device files to allow marking library elements which were generated by a script like `librepcb-parts-generator`. The property is not used yet by LibrePCB, but in future we could show a note in the UI to make it clear when a library element is generated.

I think this helps to avoid modifying or duplicating such elements manually although adjusting the generator makes more sense. A custom UI warning can be made much more visible than the current "Generated by ..." sentence in the description of library elements (which is then no longer needed).

### DIFF
--- a/libs/librepcb/core/library/libraryelement.cpp
+++ b/libs/librepcb/core/library/libraryelement.cpp
@@ -43,7 +43,8 @@ LibraryElement::LibraryElement(const QString& shortElementName,
                                const QString& description_en_US,
                                const QString& keywords_en_US)
   : LibraryBaseElement(shortElementName, longElementName, uuid, version, author,
-                       name_en_US, description_en_US, keywords_en_US) {
+                       name_en_US, description_en_US, keywords_en_US),
+    mGeneratedBy() {
 }
 
 LibraryElement::LibraryElement(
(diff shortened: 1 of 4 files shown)


dbrgn:
Yeah, that makes sense!

Would it be useful to support an URL as well?

```
(generated_by "librepcb-parts-generator" (url "https://github.com/LibrePCB/librepcb-parts-generator/"))
```

ubruhin:
Yeah I was thinking about that as well, but somehow didn't like implementing even two properties for something which most library elements don't use at all :see_no_evil: Also both properties on one line is too long, but with a linebreak it looks like wasted space in the usual case:

```lisp
(generated_by ""
 (url "")
)
```

An alternative would be to allow using Qt's HTML subset, then we could set the value to this:

```lisp
(generated_by "<a href=\"https://github.com/LibrePCB/librepcb-parts-generator/\">librepcb-parts-generator</a>")
```

It might not be the perfect solution, but I think for this special use-case (a property which is *not* visible/editable in the UI, and can easily be re-generated) it should be fine.

Whether to actually allow HTML or not could still be decided finally when implementing the corresponding note in the UI.

dbrgn:
I wouldn't allow HTML, it seems messy 🙂 However, a value like `librepcb-parts-generator (https://github.com/LibrePCB/librepcb-parts-generator/)` would be fine as well, especially if it just serves documentation purposes.

PR #1199, opened 2023-08-13
Add alternative names to packages file format

ubruhin:
The naming of packages already lead to several discussions (IPC naming vs. JEDEC/EIAJ vs. manufacturer names) and is probably confusing users since the IPC names are not (yet) as well known as names like "SOT-23". So far, our only solution is to add these well known names to the package description (for clarity) and keywords (for the search function).

However, maybe it makes sense to explicitly specify these alternative names in packages, so that the UI can somehow make them more visible than just a description. We might consider listing packages multiple times in list views, once for each name so the user will find it for sure, no matter which name he looks for. Or maybe just show the alternative names within the list views next to the main (IPC) names.

I didn't implement anything in the UI yet, but this PR adds the necessary file format extension to package files so we can extend the UI in a minor release. But I already extended the search function to respect these alternative names, so for example our generator can already start using this feature.

File format:

```lisp
(name "DIOM5226X240")
[...some other nodes, to keep the header identical for any library element type...]
(alternative_name "DO-214AC" (reference "JEDEC DO-214"))
(alternative_name "SMA" (reference ""))
```

The `reference` is optional and can for example be set to "TI" for a Texas Instruments specific name.

~~Still thinking about renaming the node to something like `aka` or `known_as` instead of `alternative_name`...~~

/cc @dbrgn

### DIFF
--- a/libs/librepcb/core/library/pkg/package.cpp
+++ b/libs/librepcb/core/library/pkg/package.cpp
@@ -87,6 +87,7 @@ Package::Package(const Uuid& uuid, const Version& version,
                  const QString& keywords_en_US, AssemblyType assemblyType)
   : LibraryElement(getShortElementName(), getLongElementName(), uuid, version,
                    author, name_en_US, description_en_US, keywords_en_US),
+    mAlternativeNames(),
     mAssemblyType(assemblyType),
     mPads(),
     mModels(),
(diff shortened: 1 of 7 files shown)


dbrgn:
I think I'd sightly prefer `alternative_name` over `known_as`, but no strong opinion.

dbrgn:
I think `reference` is fine. In general, I think the proposal makes sense, both with `known_as` and with `alternative_name`.

ubruhin:
Alright, reverted it to `alternative_name` (and kept `reference`). Thanks for sharing your opinion :+1:

PR #1200, opened 2023-08-15
Simplify development & release workflow

ubruhin:
*Note: This PR is only about the documentation of the development & release workflow, but the intention is to start following this workflow from now on.*

# Summary

So far, the general development workflow was to accept any kind of changes (including file format changes) at any time to the `master` branch, thus considering the `master` branch always as unstable. For stable releases, a release branch was created and minor changes/bugfixes were cherry-picked form `master` to the release branch. The main reason for this decision was to keep the hurdle for contributors low since they don't need to care about this workflow and any change is accepted at any time.

However, now after a few years and many releases following this workflow, it turned out that contributions are made rather rarely (especially major ones). On the other side, I've spent many many hours with cherry-picking and with setting up Transifex to be somehow compatible with this workflow. With every minor release made, this workflow becomes more cumbersome as it's hard to keep track of which commits are safe to cherry-pick and which not.

Thus I started to rethink about this workflow. Probably the only efficient workflow would be to keep the `master` branch stable (i.e. not accepting file format changes) unless the next release is planned to be a major release. I think it would simplify many things, and still shouldn't hurt contributors too much since there are only a few of them :wink:

See details below.

# Current Workflow

https://developers.librepcb.org/da/dbc/doc_release_workflow.html

image

# Suggested New Workflow

https://developers.librepcb.org/_branches/simplify-development-workflow/da/dbc/doc_release_workflow.html

image

image

Any opinions or other ideas welcome!


dbrgn:
Makes sense, I'm all for simplifications to the current (rather complicated) workflow.

Would you merge hotfixes back to `master`?

(By the way, if you ever wanted to rename `master` to `main`, which is the git default nowadays and is the technically better term in my opinon, now would be a good moment. However, I don't really mind any variant 😄)

ubruhin:
Opened a separate PR for the branch renaming: #1201 

Will keep this PR open until #1201 is merged and update the diagrams with the new branch name then.

PR #1201, opened 2023-08-16
Rename 'master' branch to 'main'

ubruhin:
Replacing all occurrences of the branch name `master` by `main`. 

As mentioned in https://github.com/LibrePCB/LibrePCB/pull/1200#issuecomment-1679678241, now is a good moment to rename our `master` branch to `main`. Probably sooner or later we have to do it anyway because `main` gets the de-facto standard and thus people expect there's a `main` branch.

I think I'll still keep the `master` branch for a while to keep tools working weren't updated yet for the new branch name. But I won't keep it updated with new commits anymore. So generally my plan after merging this is:

1. Add a new branch named `main`, based on `master`.
2. Set `main` as the main branch of the repository.
3. Update all scripts on our server to access `main` instead of `master`

/cc @dbrgn


ubruhin:
Oops, I just realized the branch name `main` is *by far* not as established yet as I though :sob:  Even with the latest Git release the default branch name when creating a new repository is still `master`. Also the Git project itself is still using `master` as the main branch, just like basically *any* well known project.

This makes me thinking it's probably too early for the rename because it rather confuses users instead of feeling consistent with other projects. In the end I'd do the rename for consistency, but at the moment it would lead to the opposite. I expect **at least Git itself** needs to be using `main` when creating a new repository before we start renaming branches on existing projects.

So I'm sorry but I'll postpone this task until `main` is *really* the de-facto standard :see_no_evil:

rnestler:
> Even with the latest Git release the default branch name when creating a new repository is still master. 

This is not true anymore, is it? At least on my system `git init` prints the following hint:
```
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint:
hint: 	git config --global init.defaultBranch <name>
hint:
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint:
hint: 	git branch -m <name>
```

Which suggests, that you should configure git to use something like `main` instead.

ubruhin:
> This isn't fully true anymore, is it?

According your posted output, yes it is:

> hint: Using 'master' as the name for the initial branch.

web-coder-of-rpi:
> The main branch is already named `master`.

Maybe you should rename the pull request, as like I said, the main branch is already master.

PR #1206, opened 2023-08-23
Board editor: Render pad & via texts with Fontobene

ubruhin:
So far the texts within pads and vias (pad number resp. net name) were drawn with a TrueType font, but it turned out this rendering was a huge performance bottleneck, slowing down the whole board editor on large designs. Especially on Windows the board editor became basically unusable (depending on zoom level).

Now rendering these texts with a Fontobene stroke font which is much faster.

### DIFF
--- a/libs/librepcb/editor/graphics/primitivefootprintpadgraphicsitem.cpp
+++ b/libs/librepcb/editor/graphics/primitivefootprintpadgraphicsitem.cpp
@@ -24,13 +24,16 @@
 
 #include "origincrossgraphicsitem.h"
 #include "primitivepathgraphicsitem.h"
-#include "primitivetextgraphicsitem.h"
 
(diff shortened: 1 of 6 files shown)


dbrgn:
> rendering was a huge performance bottleneck

Really? 😂 🤦🏻

ubruhin:
> Really? 😂 🤦🏻

Yes, really :slightly_smiling_face:  I never really experienced this problem on Linux, but on Windows it's a very serious issue. And now with this change, I realized it even slightly speeds up the board editor on Linux, pan/zoom feels completely *instant* now even with larger designs :muscle: 

I don't know if there was some issue with our implementation of TrueType font rendering. Maybe there would be a solution to speed that up somehow, but replacing it by Fontobene was not a big deal and since it's a simple stroke font it's no surprise that performance is better...

PR #1215, opened 2023-08-26
Use lighter red/yellow background colors in parts table

ubruhin:
Because @dbrgn was scared by the strong red color :grin:

image


dbrgn:
Yellow is certainly better than red! Is this yellow color used for other "data missing" fields as well?

ubruhin:
> Yellow is certainly better than red!

Red is still used, see screenshot. It's just lighter now. Red means "no part set at all", yellow means "part set, but its MPN is empty" (a part could also be represented by attributes, e.g. `CAPACITANCE=100nF`, so the MPN is not mandatory).

> Is this yellow color used for other "data missing" fields as well?

Only in the device editor (updated as well).

PR #1216, opened 2023-08-28
Show deprecated library elements with red color

ubruhin:
In most list views, draw deprecated library elements with red color. Otherwise the user has no chance to know these library elements should no longer be used. By far not perfect yet, but still much better than nothing...

image


doom-fr:
Regarding HMI, won't it be better to write this greyed and italic ?
And maybe add ' (deprecated)' after the name ? 
And reduce opacity of the icon ?

Like this : 
image

ubruhin:
@doom-fr yeah, I agree that would look better! But on some systems these lists might use a dark theme which I have no control of, and thus I need colors which work both on light and dark themes. And I guess changing the font to italic is quite inefficient, making the UI even slower than they are already if there are many items.

But the main reason is actually that I plan to create a completely new UI anyway and don't like to put too much time working on legacy UI code :see_no_evil: So it's just a temporary solution.

PR #1219, opened 2023-08-29
Show OpenCascade version in about dialog

ubruhin:
It turned out that loading STEP files is a bit fragile, depending on the STEP files and the used OpenCascade version there could be issues. But without knowing which OpenCascade edition (OCCT vs OCE) and version is used, it's hard to reproduce and debug the issues. Thus now printing the used OCC version in the about dialog:

image


ubruhin:
Now also included the OpenCascade version in `librepcb-cli --version` since the CLI also requires OCC for STEP related tasks:

```
LibrePCB CLI Version 1.0.0-unstable
File Format 1 (stable)
Git Revision 9e5115209
Qt Version 5.15.3 (compiled against 5.15.3)
OpenCascade OCE 6.9.1
Built at 31.08.23 11:59
```

PR #1222, opened 2023-08-31
CLI: Implement new command 'open-step'

ubruhin:
Add a new CLI command to minify and validate individual STEP files without requiring to open a whole LibrePCB library. This is intended mainly for `librepcb-parts-generator` to post-process generated STEP files.

### Usage

```
Usage: ./librepcb-cli [options] open-step [command_options] file
LibrePCB Command Line Interface

Options:
  -h, --help        Print this message.
  -V, --version     Displays version information.
  -v, --verbose     Verbose output.
  --minify          Minify the STEP model before validating it. Use in
                    conjunction with '--save-to' to save the output of the
                    operation.
  --tesselate       Tesselate the loaded STEP model to check if LibrePCB is
                    able to render it. Reports failure (exit code = 1) if no
                    content is detected.
  --save-to <file>  Write the (modified) STEP file to this output location (may
                    be equal to the opened file path). Only makes sense in
                    conjunction with '--minify'.

Arguments:
  open-step         Open a STEP model to execute STEP-related tasks outside of
                    a library.
  file              Path to the STEP file (*.step).
```

### Example

```
$ librepcb-cli open-step --minify --tesselate --save-to=minified.step model.step
Open STEP file 'model.step'...
Perform minify...
 - Minified from 138,951 bytes to 54,462 bytes (-61%)
Save to 'minified.step'...
Load model...
Tesselate model...
 - Built 6360 vertices with 1 different colors
SUCCESS
```

@dbrgn Maybe you like to take a look at this API before it gets released(?)

### DIFF
--- a/apps/librepcb-cli/commandlineinterface.cpp
+++ b/apps/librepcb-cli/commandlineinterface.cpp
@@ -345,6 +345,7 @@ int CommandLineInterface::execute(const QStringList& args) noexcept {
   // --verbose
   if (parser.isSet(verboseOption)) {
     Debug::instance()->setDebugLevelStderr(Debug::DebugLevel_t::All);
+    OccModel::setVerboseOutput(true);
   }
 
   // --help (also shown if no arguments supplied)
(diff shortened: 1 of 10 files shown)


dbrgn:
I wonder if it makes sense to have separate flags for `--minify` and `--save-to`, given that "[save-to] only makes sense in conjunction with '--minify'".

However, there might be other modification commands in the future, so it is probably good as-is.

ubruhin:
> However, there might be other modification commands in the future, so it is probably good as-is.

That was exactly my idea behind this concept :slightly_smiling_face:

PR #1227, opened 2023-09-19
Make list_tests.py callable from within subfolder

Luz:
It took me a while, until I noticed the script is not broken, but rather intended to be executed from the directory above.
`./dev/list_tests.py`
The changes intend to allow usage from within the subdirectory:

```
cd dev
python ./list_test.py
```
and/or
```
cd dev
python list_test.py
```
and/or
```
cd dev
./list_test.py
```


ubruhin (approved):
Thanks! :slightly_smiling_face:

PR #1229, opened 2023-09-20
Use bash from env for better portability

Luz:
Hi
According to [Stackoverflow](https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang#10383546) it looks like the portability of the development scripts could be improved by having this new style for the bash shebang.
What do you think?
If possible, verify the line 58 of "dev/format_code.sh" as I did not test this inside docker yet. :(


dbrgn (approved):
NixOS user identified 😂

ubruhin:
Thanks! :muscle:

PR #1230, opened 2023-09-21
Change backticks to POSIX style

Luz:
According [Stackoverflow](https://stackoverflow.com/questions/9449778/what-is-the-benefit-of-using-instead-of-backticks-in-shell-scripts) the backticks are legacy, so I think it would be stylish to have this new syntax.


ubruhin:
I always wondered which style is preferred, but was too lazy to ask the Internet :grinning:

dbrgn:
Yes, `${...}` is definitely preferred 🙂

Issue #121, opened 2016-09-10
Write unit tests for common classes


ubruhin:
There are still only a few unit tests. I think first of all, we should write unit tests for the most important common classes, so those located in [libs/librepcbcommon](https://github.com/LibrePCB/LibrePCB/tree/master/libs/librepcbcommon).

Most important unit test (list will be updated regularly):
- [x] DirectoryLock
- [x] FilePath
- [ ] FileUtils _[high priority, easy]_
- [ ] SmartFile, SmartTextFile, SmartVersionFile, SmartXmlFile
- [ ] XmlDomDocument, XmlDomElement _[not yet ready, should be refactored first]_
- [ ] Angle, Length, LengthUnit, Point (already started)  _[high priority, easy]_
- [ ] Alignment _[easy]_
- [x] Application
- [x] ScopeGuard, ScopeGuardList
- [x] SystemInfo
- [ ] UndoCommand, UndoCommandGroup, UndoStack _[high priority]_
- [x] Uuid
- [x] Version

javsanpar:
Hello!

I want to learn C++ and I have thought that can practice helping with this issue.

Which tests you think I should start?

Thank you in advance. :smile:

ubruhin:
Hi @javsanpar,

> I want to learn C++ and I have thought that can practice helping with this issue.

Great! :smiley: 

> Which tests you think I should start?

I just updated the list in the issue description. You can choose freely which tests to implement, I think `Alignment` or `VersionFile` are easy and thus good units to start with. `FileUtils` requires a bit more effort, and `UndoStack` is probably the most complicated one.

javsanpar:
Sorry, I have been doing my final exams and I haven't had time to do the tests until a few days ago. :sweat_smile:

I just did a prototype, I wasn't sure what methods to test.: https://privatebin.net/?c05a7d078ce86f6c#eOwj9zGpDoEP0XjIjKhDA6nb6sbxPmEevmIPbXumWnk=

What do you think?

ubruhin:
> Sorry, I have been doing my final exams and I haven't had time to do the tests until a few days ago. sweat_smile

No problem :)

> What do you think?

Looks good so far, except that the test data contains 4 times exactly the same values :grin:

> I wasn't sure what methods to test

The more, the better ;)

Btw, feel free to open a pull request, it's easier to review than an uploaded diff.

Shivam-jainn:
Hey ! 
[...]

PR #1233, opened 2023-09-25
Add unit test for FileUtils

BuFran:
This PR realizes issue #121 

During testing, I found two bugs inside `FileUtils::getFilesInDirectory` functions. 

I created issues #1234 and #1236 for them.

One of them is simple copy-paste mistake inside `FileUtils::getFilesInDirectory` and can be simply corrected in #1234. 

But the second needs deeper examination and testing under different OS or discussion on API used.

### DIFF
--- a/tests/unittests/core/fileio/fileutilstest.cpp
+++ b/tests/unittests/core/fileio/fileutilstest.cpp
@@ -1,0 +1,313 @@
+/*
+ * LibrePCB - Professional EDA for everyone!
+ * Copyright (C) 2013 LibrePCB Developers, see AUTHORS.md for contributors.
+ * https://librepcb.org/
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*******************************************************************************
+ *  Includes
+ ******************************************************************************/
+
+#include <gtest/gtest.h>
+#include <librepcb/core/exceptions.h>
+#include <librepcb/core/fileio/filepath.h>
+#include <librepcb/core/fileio/fileutils.h>
+#if defined(Q_OS_WIN32) || defined(Q_OS_WIN64)  // Windows
+#include <fileapi.h>
+#endif
+
+#include <QtCore>
+
+/*******************************************************************************
+ *  Namespace
+ ******************************************************************************/
+namespace librepcb {
+namespace tests {
+
+/*******************************************************************************
+ *  Test Class
+ ******************************************************************************/
+
+class FileUtilsTest : public ::testing::Test {
+protected:
+  FilePath root;
+  FilePath rootFile;  // source for all operations
+  FilePath rootFileHidden;
+
+  FilePath subdir;
+  FilePath subdirFile;
+  FilePath subdirSubdir;
+  FilePath subdirSubdirFile;
+  FilePath subdirSubdirFileHidden;
+
+  FilePath rootFileMissing;
+  FilePath rootFileCopy;
+  FilePath subdirCopy;
+  FilePath subdirCopyFile;
+  FilePath subdirCopySubdir;
+  FilePath subdirCopySubdirFile;
+  FilePath subdirCopySubdirFileHidden;
+
+  QStringList filter;
+
+  static void SetupFile(const FilePath& pth, const QByteArray& content) {
+    QFile f(pth.toNative());
+    if (f.open(QIODevice::ReadWrite | QIODevice::NewOnly)) f.write(content);
+    f.close();
+  }
+
+  void SetUp() override {
+    root = FilePath::getRandomTempPath();
+
+    // the sources of already existing files and directories
+    rootFile = FilePath::fromRelative(root, "file.txt");
+    rootFileHidden = FilePath::fromRelative(root, ".hidden.txt");
+    subdir = FilePath::fromRelative(root, "subdir");
+    subdirFile = FilePath::fromRelative(subdir, "file.txt");
+    subdirSubdir = FilePath::fromRelative(subdir, "subdir");
+    subdirSubdirFile = FilePath::fromRelative(subdirSubdir, "file.txt");
+    subdirSubdirFileHidden =
+        FilePath::fromRelative(subdirSubdir, ".hidden.txt");
+
+    // the destinations for copying files, nonexistent at start of test
+    rootFileCopy = FilePath::fromRelative(root, "fileCopy.txt");
+    rootFileCopy = FilePath::fromRelative(root, "missing.txt");
+    subdirCopy = FilePath::fromRelative(root, "subdirCopy");
+    subdirCopyFile = FilePath::fromRelative(subdirCopy, "file.txt");
+    subdirCopySubdir = FilePath::fromRelative(subdirCopy, "subdir");
+    subdirCopySubdirFile = FilePath::fromRelative(subdirCopySubdir, "file.txt");
+    subdirCopySubdirFileHidden =
+        FilePath::fromRelative(subdirCopySubdir, ".hidden.txt");
+
+    QDir().mkdir(root.toNative());
+    QDir().mkdir(subdir.toNative());
+    QDir().mkdir(subdirSubdir.toNative());
+
+    SetupFile(rootFile, "test\n");
+    SetupFile(rootFileHidden, "hiddenContent\n");
+    SetupFile(subdirFile, "test\n");
+    SetupFile(subdirSubdirFile, "test\n");
+    SetupFile(subdirSubdirFileHidden, "hiddenContent\n");
+
+    // TODO make hidden files already hidden under Windows !
+#if defined(Q_OS_WIN32) || defined(Q_OS_WIN64)
+    SetFileAttributes(rootFileHidden.toNative().toStdWString().c_str(), FILE_ATTRIBUTE_HIDDEN);
+    SetFileAttributes(subdirSubdirFileHidden.toNative().toStdWString().c_str(), FILE_ATTRIBUTE_HIDDEN);
+#endif
+
+    filter << "*.txt";
+  }
+
+  void TearDown() override { QDir(root.toNative()).removeRecursively(); }
+
+  static void ExpectList(const QList<FilePath>& result,
+                         const QList<FilePath>& expected) {
+    std::string resultStr;
+    std::string expStr;
+    for (const auto& i : result)
+      resultStr += "  " + i.toStr().toStdString() + "\n";
+    for (const auto& i : expected)
+      expStr += "  " + i.toStr().toStdString() + "\n";
+
+    ASSERT_EQ(result.size(), expected.size())
+        << "\nActual:\n" << resultStr
+        << "\nExpected:\n" << expStr;
+    for (const auto& i : expected)
+      ASSERT_TRUE(result.contains(i))
+          << "\nActual:\n" << resultStr
+          << "\nExpected:\n" << expStr;
+  }
+};
+
+/*******************************************************************************
+ *  Test Methods
+ ******************************************************************************/
+
+TEST_F(FileUtilsTest, testReadExistingFile) {
+  auto p = FileUtils::readFile(rootFile);
+
+  EXPECT_EQ("test\n", p.toStdString());
+}
+
+TEST_F(FileUtilsTest, testReadNonexistentFileShouldThrow) {
+  EXPECT_THROW(FileUtils::readFile(rootFileMissing), Exception);
+}
+
+TEST_F(FileUtilsTest, testWrittenDataShouldBeReadBack) {
+  FileUtils::writeFile(rootFile, "someData\n");
+  auto p = FileUtils::readFile(rootFile);
+
+  EXPECT_EQ("someData\n", p.toStdString());
+}
+
+TEST_F(FileUtilsTest, testCopyValidFile) {
+  FileUtils::copyFile(rootFile, rootFileCopy);
+  auto p1 = FileUtils::readFile(rootFile);
+  auto p2 = FileUtils::readFile(rootFileCopy);  // throws if file not found
+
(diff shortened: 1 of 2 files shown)


BuFran:
I separated #1238 from this PR, that can go on its own way now.

This will be failing until #1236 will be resolved first.

BuFran:
This is now solved in #1241, this tests are not failing now.

ubruhin:
> I don't see any option to remove "draft" flag here so i'd like make this PR active back.

It's a button called "Ready for review". But Windows CI seems to be still failing :see_no_evil:

BuFran:
> It's a button called "Ready for review". But Windows CI seems to be still failing 🙈

Yes I see it now, after CI fails due to other reasons ... 

The problem is that I don't know how to make some file hidden using qt platform independent objects on Windows. 

Running CI is the only test I can do here, i have no computer with Windows. 

I will try to fix it soon, but it will require many commits :-( Any help of someone who is familiar with Qt on windows platform can speed it up. (it is commented as TODO in my code now)

tests/unittests/core/fileio/fileutilstest.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+  QStringList filter;
+
+  static void SetupFile(const FilePath& pth, const QByteArray& content) {
+    QFile f(pth.toNative());
+    if (f.open(QIODevice::ReadWrite | QIODevice::NewOnly)) f.write(content);
+    f.close();
+  }
+
+  void SetUp() override {
```
ubruhin, line 72: For consistency with all other tests, could you use the constructor/destructor instead of setup/teardown functions? It would also allow to directly initialize members like `filter` right in the constructor's initialization list.

Btw, I also wonder if it actually makes sense to create all these objects and files for each of the tests, although most tests only need one or two of them. Especially on Windows creating files is quite slow so it unnecessarily increases test time (although not too much). I think I'd have moved some of these things right into the corresponding test functions, but I'm also fine with keeping it that way.

tests/unittests/core/fileio/fileutilstest.cpp
```diff
@@ -0,0 +1,313 @@
...
+    root = FilePath::getRandomTempPath();
+
+    // the sources of already existing files and directories
+    rootFile = FilePath::fromRelative(root, "file.txt");
```
ubruhin, line 76: Just FYI but nothing critical no no need to change here: Basically everywhere else we use this form (it is also shorter):

```suggestion
    rootFile = root.getPathTo("file.txt");
```

tests/unittests/core/fileio/fileutilstest.cpp
```diff
@@ -0,0 +1,313 @@
...
+    QDir().mkdir(subdir.toNative());
+    QDir().mkdir(subdirSubdir.toNative());
+
+    SetupFile(rootFile, "test\n");
+    SetupFile(rootFileHidden, "hiddenContent\n");
+    SetupFile(subdirFile, "test\n");
+    SetupFile(subdirSubdirFile, "test\n");
+    SetupFile(subdirSubdirFileHidden, "hiddenContent\n");
+
+    // TODO make hidden files already hidden under Windows !
```
ubruhin, line 105: The TODO is now obsolete, right?

tests/unittests/core/fileio/fileutilstest.cpp
```diff
@@ -0,0 +1,313 @@
...
+    SetFileAttributes(rootFileHidden.toNative().toStdWString().c_str(), FILE_ATTRIBUTE_HIDDEN);
+    SetFileAttributes(subdirSubdirFileHidden.toNative().toStdWString().c_str(), FILE_ATTRIBUTE_HIDDEN);
+#endif
+
+    filter << "*.txt";
+  }
+
+  void TearDown() override { QDir(root.toNative()).removeRecursively(); }
+
+  static void ExpectList(const QList<FilePath>& result,
```
ubruhin, line 116: Inconsistent function naming here as well. 

And btw, actually we avoid using `ASSERT_()` and `EXPECT_()` calls in some helper function because it causes the "wrong" line number to be printed in the output of failed tests, making it harder to find the statements which fail. What we do typically is to provide a `str()` function which only serializes the passed (complex) data to a `std::string`, and compare the strings right in the test functions. This also makes it obsolete to use the `<<` operator for pretty printing. See for example `PathTest`. Could you maybe also apply this pattern here for consistency?

tests/unittests/core/fileio/fileutilstest.cpp
```diff
@@ -0,0 +1,313 @@
...
+  EXPECT_TRUE(subdirCopySubdirFileHidden.isExistingFile());
+}
+
+TEST_F(FileUtilsTest, testFindDirectories) {
+  auto p = FileUtils::findDirectories(root);
+
+  ExpectList(p, {subdir});
+}
+
+TEST_F(FileUtilsTest, testFindFileInDirectory) {
```
ubruhin, line 226: Just a note (can be left as-is): Generally try to keep the test names consistent with the function under test, it makes reading tests and their output more easy and intuitive. For example here the function under test is called `getFilesInDirectory` so the test could be called `testGetFilesInDirectory` ("get" vs. "find", "file" vs. "files"). This way it's immediately clear from the test name which function it tests.

If there are multiple tests for the same function, of course additional terms need to be added to the test name - ideally (but not necessarily) at the end of the test name.

tests/unittests/core/fileio/fileutilstest.cpp
BuFran: Yeah, I like the other norm, I will corect the code to use that way. 

I am still new in codebase, looking what and how it is made in other tests :-).

tests/unittests/core/fileio/fileutilstest.cpp
BuFran: Yeah, I'll remove it.

tests/unittests/core/fileio/fileutilstest.cpp
BuFran: Yes, you have right there, my original intent was to describe what the test does initially (the obvious result) when making the test instead of the function name that is little bit cryptic for me. 

And I forgot to rename it your way before opening PR. I'll correct it.

tests/unittests/core/fileio/fileutilstest.cpp
BuFran: I'll look at PathTest but if I remember I looked there before (my code is modified copy of `PathTest` - mainly using delete key :smiley: ). 

I agree that `ASSERT` and `EXPECT` in helpers are not descriptive, but the original has not been too. I'll look at other ways to do.

tests/unittests/core/fileio/fileutilstest.cpp
BuFran: No, there is intent to make the test this way! 

I need for each test to have clean folder structure because the functions have *side effects*. I like tools like `gtest-parallel` which rapidly speeds up testing by starting all tests **in parallel** on machines that supports multiple cores (and this is the future of PCs) and there is no guide in the `gtest` in which order the tests should be run. I write always all tests compatible with this idiom (even when your code doesn't use it) because in the future it can be simply implemented without change in code of test as the test base will grow.

Other tests (like `FilePath`) have no side effects, so they don't need to use this idiom, they could use your way. Your lib is clever designed so the objects have little or no side effects and when I looked at some tests, they can be run in either order without problems (even in parallel)

According to `filter`, yes that can be made in initialization of variable, i'll move it. It's result of digging what the hell the function returns when I search for bug in your codebase.

tests/unittests/core/fileio/fileutilstest.cpp
```diff
@@ -0,0 +1,313 @@
...
+  FilePath rootFileCopy;
+  FilePath subdirCopy;
+  FilePath subdirCopyFile;
+  FilePath subdirCopySubdir;
+  FilePath subdirCopySubdirFile;
+  FilePath subdirCopySubdirFileHidden;
+
+  QStringList filter;
+
+  static void SetupFile(const FilePath& pth, const QByteArray& content) {
```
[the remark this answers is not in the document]
BuFran: Well there was two options about capitals:

 * naming conventions of `gtest`
 * naming conventions of `LibrePCB`

Because my test code *inherits* from `gtest`, and there are *overriden* functions from `gtest`, i chosen the `gtest` version because of consistency across entire object. The other way will generate `setupFile` and `SetUp` functions and this could look ugly for someone (inconsistent across object).

It's not problem for me to use name `setupFile` if that will be the way You want. I'm happy that I predicted that this will be commented in PR before I created PR. Those assumptions makes code cleaner and I don't know which selection You will use.

tests/unittests/core/fileio/fileutilstest.cpp
```diff
@@ -0,0 +1,313 @@
...
+    QFile f(pth.toNative());
+    if (f.open(QIODevice::ReadWrite | QIODevice::NewOnly)) f.write(content);
```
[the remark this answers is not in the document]
BuFran: Good point is not managing fail inside helper function. It is nice facepalm i got and accepting it :-)

The second is nice point from the test perspective. Test on another technology than it runs can reveal more bugs. My previous assumption is "use app codebase that is already used" so when You use Qt everywhere, I should use Qt here too, to be consistent. This assumption seems to be wrong.

In `c++17` there is `std::filesystem` as platform independent native and with that lot of filesystem code in librepcb can be outdated but yeah, i know you don't like it, there *is* an reason by now so I should wait ... :-) 

So the question is I should rewrite that by using:

* plain C `fread`/`fwrite`
* C++ `std::fstream`
* Qt `QFile`

tests/unittests/core/fileio/fileutilstest.cpp
ubruhin: Hm, according [googletest FAQ](https://google.github.io/googletest/faq.html#CtorVsSetUp) the fixture object is explicitly created/destroyed for every single test, so the constructor/destructor behave the same way as SetUp/TearDown (with very small differences), or am I missing something?

tests/unittests/core/fileio/fileutilstest.cpp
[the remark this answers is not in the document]
ubruhin: Yeah even if the gtest naming convention is different, I prefer to use our own naming convention whenever possible. It avoids propagating the gtest naming conventions to other parts of our code by copy&paste.

But I still think the SetUp/TearDown functions are not needed anyway (see other comment) which would eliminate the inconsistency anyway :grin:

tests/unittests/core/fileio/fileutilstest.cpp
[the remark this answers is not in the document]
ubruhin: > My previous assumption is "use app codebase that is already used" so when You use Qt everywhere, I should use Qt here too, to be consistent. This assumption seems to be wrong.

It's not wrong, generally of course we use Qt everywhere to keep the code as portable as possible :slightly_frowning_face: The `FileUtils` test is a bit a special case since we're not just testing logic inside our application - in fact, the function under test forwards data to the operating system (leaving our application) and the test reads back the data from the operating system. So the behavior of the operating system (with its filesystem) is actually also verified by the test, which is very unique for the `FileUtils` test (only very few other test are involving the operating system). Just because of this, it came in my mind that using a different API for the readback might make sense, to hopefully cover issues related to Qt's file functions. For (theoretical) example, if Qt cached read/write operation, a readback might be successful even though the data has not been written to the file system yet.

> So the question is I should rewrite that by using:

I'd give `std::fstream` a try. I don't like using C functions if not absolutely necessary, as they are much more cumbersome and errorprone than the C++ API (e.g. error handling is usually simpler with C++).

However, it is just an idea an maybe it's not as simple as it sounds. So generally I'd also be fine using `QFile` for verification. For example if `std::fstream` turns out to be difficult, don't waste time with it and just stick with `QFile`. In the end, Qt developers should have tested `QFile` already :see_no_evil:

tests/unittests/core/fileio/fileutilstest.cpp
BuFran: Yeah, from that point of view, implementing YAGNI, you are right. 

I made those tests using `INSTANTIATE_TEST_SUITE_P` macro, and those tests was only single parametrized test. In that scenario, the constructor is called only once per all parametrized items. (well the behavior on the page isn't right) Therefore I needed the SetUp/TearDown overrides.

Second assumtpion was, when SetUp/TearDown crashes (due to filesystem rejecting file creation / out of space - it is obvious problem on my system as you know, and that was my case), it is only failing test case with different message. The filesystem exception can't be handled in destructor or constructor (during cleaning or creation) and that terminates the test completely. 

This comes from my practice that file i/o *may* throw so I always suppose on first sight that they may throw, even in this scenario probably they may not. 

Implementing YAGNI, rewriting to `std::fstream` it can be done using constructors as You said, so I'll rewrite that way.

BuFran:
Prepared for review.

The test output on failed test results now in (shortened):
```
Expected equality of these values:
  comparable(actual)
    Which is: "/tmp/librepcb/1696570756777_2075782095/file.txt..."
  comparable(expected)
    Which is: "/tmp/librepcb/1696570756777_2075782095/file.txt..."
With diff:
@@ +1,3 @@
+/tmp/librepcb/1696570756777_2075782095/file.txt
 /tmp/librepcb/1696570756777_2075782095/file.txt
 /tmp/librepcb/1696570756777_2075782095/subdir/file.txt
```

I believe I resolved all of Your comments.

tests/unittests/core/fileio/fileutilstest.cpp
```diff
@@ -0,0 +1,320 @@
...
+    auto join = [](std::string a, std::string b) {
+      return !a.empty() ? (std::move(a) + '\n' + std::move(b)) : std::move(b);
+    };
+    auto conv = [](const FilePath& filepath) {
+      return filepath.toStr().toStdString();
+    };
+    std::vector<std::string> paths{};
+    std::transform(path.begin(), path.end(), std::back_inserter(paths), conv);
+    std::sort(paths.begin(), paths.end());
+    return std::accumulate(paths.begin(), paths.end(), std::string(), join);
```
ubruhin, line 141: Huh this function looks very complicated for the simple thing it does :see_no_evil: I'm not even sure if the `std::move()` calls actually make sense in this context.

With Qt it would be not only much easier to understand, but even shorter:

```
QStringList pathsStr;
foreach (const FilePath& fp, paths) {
  pathsStr.append(fp.toStr());
}
pathsStr.sort();
return pathsStr.join("\n").toStdString();
```
Anyway, your solution seems to work so this time I could live with that...

BuFran: I forgot that Qt has "join" function that is missing in plain c++, and I agree that using it produces more readable code. 

But this code part you recommended is Yours, during squashing, your authorship will be deleted. I mean it will be better that you will push your commit rewriting this function content just after squash merge of this PR. Both versions will be in the history for comparison.

PR #1238, opened 2023-09-27
Add unit test for VersionFile

BuFran:
This test isn't failing, it can move forward faster than failing #1233 that needs supporting PR solving found issue.


ubruhin (approved):
Looks good, thanks!

I just added a commit to apply auto-formatting. Will merge after CI succeeds.

PR #1239, opened 2023-09-27
Upgrade to c++17 standard as recommended by both Google / Microsoft rules

BuFran:
[Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html)

"Currently, code should target C++17, i.e., should not use C++2x features"

[Enforcement](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#enforcement-1)
Use an up-to-date C++ compiler (currently C++20 or C++17) with a set of options that do not accept extensions.


ubruhin:
I appreciate your initiative in contributing to LibrePCB. However, currently I do not agree with switching to a newer C++ standard. This is not a task of simply changing a number in a CMake file. It's a strategic decision - and one of LibrePCB's goals is to be as portable as possible, so the more portable C++11 makes more sense than C++14/17. For example, LibrePCB is still being packaged for OpenPandora, which doesn't ship a compiler fully supporting C++17. Of course one could argue that OpenPandora is not an important target, but the question is how much we want to reduce portability for how much advantage C++17 provides.

IMHO the advantages of C++17 over C++11 are rather small (and only for developers, not for users!). So I do not like to give up portability for just a small improvement for developers. Of course some day we need to switch, latest when we start supporting Qt6 since it requires C++17. But that's a different topic.

Since you are referencing Google guidelines, they also state:

> Consider portability to other environments before using features from C++14 and C++17 in your project.

I guess they mean exactly what I am trying to explain here :wink: 

And that CppCoreGuidelines (initially _not_ from Microsoft) recommend to use the latest C++ standard is not a surprise so for me that's not a valid argument :see_no_evil: 

Closing this PR, sorry!

Issue #1179, opened 2023-07-19
Select-all in library editor only selects items within 14 squares of the center


kotchin:
Thank you for creating LibrePCB. I really enjoy using it.

I'm sorry to have to report a bug. This probably qualifies as a minor bug.

##### SUMMARY
When using CTRL+A to select all in the library editor, the pins, shapes and lines being selected are only those within 14 squares of the center (vertically and horizontally). There is no issue selecting any/all of them when using the mouse.
This bug does not occur in the schematic editor.

##### VERSION / OS / ENVIRONMENT
```
LibrePCB Version: 0.1.7
Git Revision:     ba553c6
Build Date:       2022-10-03T14:10:33
Qt Version:       5.15.10 (built against 5.15.6)
CPU Architecture: x86_64
Operating System: KDE Flatpak runtime
Platform Plugin:  wayland
TLS Library:      OpenSSL 3.0.8 7 Feb 2023
```

##### EXPECTED RESULTS
All items and pins should be selected.

##### ACTUAL RESULTS
Only pins within 14 squares of the center are selected, horizontally and vertically.

##### ADDITIONAL INFORMATION
Below is the result of using CTRL+A, with the symbol in 3 various positions (notice how the selected pins differ in each case).

![lib_ed_1](https://github.com/LibrePCB/LibrePCB/assets/6894486/1008a3d1-cfcf-435e-9f57-fc541be9b112)

![lib_ed_2](https://github.com/LibrePCB/LibrePCB/assets/6894486/9ec9f0f3-ece5-49f1-abe6-96952465d81b)

![lib_ed_3](https://github.com/LibrePCB/LibrePCB/assets/6894486/46e11f6a-8cca-4be4-afda-9649740d7cde)

![lib_ed_4](https://github.com/LibrePCB/LibrePCB/assets/6894486/e3405dc2-298d-4b7b-b2c0-bbec2417bc4f)

I'm not that great at programming, but if you need more help with identifying the bug, I can see what I can do.

ubruhin:
> Thank you for creating LibrePCB. I really enjoy using it.

Glad to hear! :heart: 

> When using CTRL+A to select all in the library editor, the pins, shapes and lines being selected are only those within 14 squares of the center (vertically and horizontally). 

Oh that's indeed a bug, I was able to reproduce it!  Thanks for reporting with all the details! :+1:

BuFran:
[...]

PR #1243, opened 2023-10-02
Select-all functionality in library (fixing #1179)

BuFran:
The `currentGraphicsItem` is always empty (there is no insert anywhere) so it returned always empty rect.

After this PR, for `selectAll` functionality, the entire `graphicsScene` for retrieving entire content rect is used.

Maybe the solution is wrong, probably I missed intent of `currentGraphicsItem`. The board and schematic does this differently.

One architecture misconception is that on symbol it is **reference**, on package it is **shared pointer** so probably there is really unfinished rework.

There should be no style bugs because I tried to run `dev/format_code.sh` :smiley:

### DIFF
--- a/libs/librepcb/editor/library/pkg/fsm/packageeditorstate_select.cpp
+++ b/libs/librepcb/editor/library/pkg/fsm/packageeditorstate_select.cpp
@@ -373,8 +373,9 @@ bool PackageEditorState_Select::processSelectAll() noexcept {
       if (auto item = mContext.currentGraphicsItem) {
         // Set a selection rect slightly larger than the total items bounding
         // rect to get all items selected.
-        item->setSelectionRect(
-            item->boundingRect().adjusted(-100, -100, 100, 100));
+        auto bounds = mContext.graphicsScene.itemsBoundingRect();
+        bounds.adjust(-100, -100, 100, 100);
+        item->setSelectionRect(bounds);
         emit availableFeaturesChanged();  // Selection might have changed.
         return true;
       }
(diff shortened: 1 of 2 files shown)


ubruhin:
I think the code makes sense :+1: Will quickly test it soon.

> One architecture misconception is that on symbol it is reference, on package it is shared pointer so probably there is really unfinished rework.

No, this is on purpose. A symbol always consists of exactly one graphics item (_the_ symbol) so we can use a reference. But a package consists of 0..n footprints, and thus 0..n graphics items. In case of 0 footprints, we don't have any graphics to display so it will be `nullptr`.

> There should be no style bugs because I tried to run dev/format_code.sh 😃

Awesome :) By the way, since you are contributing regularly, I may tell you some more things about our preferred workflow :slightly_smiling_face:

We prefer commit messages according [this recommendation](http://chris.beams.io/posts/git-commit/) (see also [`CONTRIBUTING.md`](https://github.com/LibrePCB/LibrePCB/blob/master/CONTRIBUTING.md)). It may sound picky, but it really helps to keep the Git log readable and messages easy to understand if every contributor follows the same style. 

For example the GitHub issue should not be mentioned in the first line of the commit message. Just describe what you did in this commit (e.g. "Library editor: Use GraphicsScene::itemsBoundingRect() to select all items"). If the commit closes a particular GitHub issue, just write "Closes #nnn" at the end of the PR body - this will make GitHub to close the issue automatically when merging the PR so we don't have to do it manually.

So far our preference, but especially for simpler PRs where a single commit is enough, don't worry too much about it. This is also an advantage of squash-merge that the final commit message can easily be adjusted in the GitHub UI :see_no_evil:

PR #1247, opened 2023-10-03
Add more licenses

BuFran:
In New project wizard, the user can select only some of CC licenses. This commit adds all versions of CC licenses, and some Open Hardware licenses as proposed in #659.

Licenses added:

 - CC-BY-NC
 - CC-BY-NC-ND
 - CC-BY-NC-SA
 - CC-BY-ND
 - TAPR-OHL
 - CERN-OHL-P
 - CERN-OHL-W
 - CERN-OHL-S

Resolves: LibrePCB#659


dbrgn:
Some of the licenses require additional actions. For example, the CERN OHL 2.0 (see https://ohwr.org/project/cernohl/wikis/Documents/CERN-OHL-version-2) suggests (does not strictly require) to add the user guide:

screenshot-2023-10-03T09:28:38+02:00

However, it requires to add a license header or license file to every source file:

screenshot-2023-10-03T09:29:35+02:00

It also requires adding a visible note to the PCB itself:

screenshot-2023-10-03T09:30:54+02:00

LibrePCB could easily fulfil the first one (adding the user guide). The second item is a bit more tricky, although it would be possible that LibrePCB is aware of the license used, and automatically generates a license file for every newly added source file.

The third one _could_ be done in an automated way by automatically adding a text object to the schematic. Alternatively, we could include a "list of user actions" for every license, and simply show those to the user ("when applying the license: 1. read through the user guide, 2. add a visible note to the PCB, 3. ...").

See #666 for my initial idea (which should probably be done in a more simple / declarative way).

(For now I guess it would also be OK to simply add the license text files, but if folks simply assume that this is all they need to do, the projects won't be correctly licensed.)

ubruhin:
> However, it requires to add a license header or license file to every source file:
> [...]
> The second item is a bit more tricky, although it would be possible that LibrePCB is aware of the license used, and automatically generates a license file for every newly added source file.

Is there actually somewhere described what they mean with the term "source file"? For me this term is not intuitive at all for a hardware project. Strictly speaking, our source files are S-Expression files but it makes absolutely no sense to add license text to those files. On from the point of hardware, "source file" possibly means just the "project file" (which is a directory for LibrePCB) so adding the license file once to the project directory is enough? At least this makes most sense to me.

> The 2.2.9 can be solved by new custom variable {{LICENSE}} that can be prefilled by selected text in project settings and user should place this text (via license component, for example) on the board.

Actually I like this idea. Probably this could be done without introducing any file format changes by using project-level attributes.

But I wonder what text should be placed in that variable. I'd put only the license name (e.g. "CERN-OHL-S v2") into it, but placing a text `{{LICENSE}}` on the board would still not comply with the OHL guide since it requires to put the whole sentence "Licensed under CERN-OHL-S v2" on the PCB. IMHO this requirement is completely crap, do the people writing those guidelines know how small free space is available on most PCBS? :sob:

Maybe call the variable `LICENSE_TEXT` and for OHL licenses put the whole sentence into it (including the link to the source location), while for other licenses only put its name into it?

Another topic: @dbrgn  do you have any opinion if it makes sense to add all the different CC licenses?

BuFran:
According to "what is souce" question, it is answered inside the license text:

obrazek

So the `LICENSE.txt` inside project folder should be sufficient, but there should be published with generated PDFs because our software is proprietary ?

ubruhin:
> [...] because our software is proprietary ?

I hope LibrePCB is *not* proprietary :see_no_evil: 

> edit2: When the LibrePCB will be approved by FSF or OSI, user will don't need the PDF generation step

As far as I understand the license text in your screenshot, not the *software* needs to be approved, but its *license*. Without checking this explicitly, assume our GPLv3+ is approved by both FSF and OSI.

ubruhin:
Sorry for the long delay - finally I was able to review this.

Generally I like it, including the "Additional actions might be required" note. However, IMHO the placement of this note is not ideal as it is far away from the license dropdown. In addition, to me the wizard now feels too cluttered. With the default wizard size on my system the text is even slightly cropped which looks ugly. And actually for the typical case where no license is chosen, this note is not even relevant at all.

I also noticed that the hyperlink of the "info" icon next to the license dropdown is not suitable anymore since it links to the Creative Commons Wikipedia page, which doesn't make sense for the new licenses. This brings me to the idea to move the "Additional actions..." note into the tooltip of the info icon (to declutter the UI) and make the hyperlink pointing to https://github.com/LibrePCB/LibrePCB/tree/master/share/librepcb/licenses because the tooltip recommends to "see the license text" and this hyperlink would actually point to the license text. The hyperlink is not ideal as it is not stable, but I think for this non-production-critical feature it's good enough.

Any thoughts on this idea? Or other ideas?

FYI, there were encoding problems in some added license files. Therefore I downloaded them again from https://creativecommons.org/2014/01/07/plaintext-versions-of-creative-commons-4-0-licenses/ which fixed the encoding. It seems there were even some small updates to the already existing licenses so I updated them as well. I just pushed these changes to your branch so you don't need to do it again :see_no_evil:

BuFran:
Thanks for review ! 

  The encoding seems strange (I downloaded it now and i have obviously different server because my fresh download from [this link](https://creativecommons.org/2014/01/07/plaintext-versions-of-creative-commons-4-0-licenses/) it still contains wrong characters. So thanks for adjusting.

 About the UX, my intent is to change it as small as possible for now. I have some proposals for changes here but the issue touched has been only "adding more licenses", not "changing UX behavior".

My proposal is to remove a lot of code for filling license dropbox and make it automatically searched from the directory [licenses](https://github.com/LibrePCB/LibrePCB/tree/master/share/librepcb/licenses) where each license would be in separate directory named as the license name. Entire directory content (nonhidden) will be copied to the project when license be accepted. It will benefit the user with:

 * User can add his own license just by putting their files to the [licenses](https://github.com/LibrePCB/LibrePCB/tree/master/share/librepcb/licenses)  folder
 * The license can contain more files when needed (license PDFs for example)
 * The link can point to installed directory with correct license text (no web connection required)
 * Code can "detect" more actions required some way from the license directory
 * Adding new license will not need to change the codebase.
 * The installer can have "check" options to install only licenses the user really wants to use in production.
 * Users can share "license packages" without touching librepcb team at all.
 * There can be button in UX "install license" that can unpack zip-packaged license into that directory ?

The problem I am solving now is how to tell LibrePCB some metainfo about license in that directory. 

 * Displayed name
 * Translated name (the name will be not included in 'ts' file anymore)
 * Variables to fill inside schematic/board (`{{LICENSE_TEXT}}` for example)
 * Actions that will be needed for DRC/ERC run that project is correctly licensed
 * Some longer text describing the license in "info" icon
 * Timestamp of license download

In my first thought, it is some user-readable  `ini` file inside license directory, but I am open for discussion here. I will not prefer the one big metainfo for all licenses because I like the copy license directory/delete it to enable/disable that license in the dialog.

But that is BIG change and needs more discussion in separate issue so therefore it will not be included in this PR. It will contain some thoughts from #666. Moreover, it will be split to two parts, the core (searching and licensing) and UX to enable licensing over commandline without UX objects.

BuFran:
In #1266 I have different proposal how to display "Additional actions..." message. If you don't like way it's done in this PR, you can skip that commit (da1e4d3) during merge and we could discuss the ways how the user can be informed about other actions more in #1266 (there are plenty other options).

dbrgn:
Nice 🙂

Issue #608, opened 2019-11-29
Board editor: Change color of inner layer 2 (too close of inner layer 1 color)


doom-fr:
##### VERSION / OS / ENVIRONMENT

```
LibrePCB Version: 0.1.3-unstable
Git Revision:     bd616e14d1366486cf53cad8b8c7e3e67119d960
Build Date:       2019-11-17T15:33:39
Qt Version:       5.12.3 (built against 5.12.3)
CPU Architecture: x86_64
Operating System: Ubuntu 18.04.3 LTS
Platform Plugin:  xcb
```

##### SUMMARY
inner layer 1 & 2 are quite same color

##### STEPS TO REPRODUCE
1. set to 2 inner layers
2. draw something in each layer

##### EXPECTED RESULTS
all copper layers (at least the 4 firsts) should have different colors that are easily distinguishable.

##### ACTUAL RESULTS
inner layer 1 & 2 are quite same color

![image](https://user-images.githubusercontent.com/7809481/69250084-4f8c5280-0baf-11ea-9028-e2a176788330.png)

##### ADDITIONAL INFORMATION
And when inner1 is selected, it is really quite the same as inner2

![image](https://user-images.githubusercontent.com/7809481/69250155-6cc12100-0baf-11ea-968d-84589041c734.png)

##### SOLUTION
exchange between inner layer 2 color and inner layer 4 color.
that will give this stack of color : 
![image](https://user-images.githubusercontent.com/7809481/69251187-561bc980-0bb1-11ea-9159-db2416d00665.png)

doom-fr:
editing `boards/board/settings.user.lp` file allow to change all colors.

ubruhin:
@dbrgn Since you are the master of colors, maybe that's an issue for you? :grin:

PR #1248, opened 2023-10-03
Visually separate colors in board layers

BuFran:
The colors of Documentation layer and inner layers has not been visually differentiated.

This PR uses recommended colors by doom-fr in LibrePCB#707 and LibrePCB#608
therefore fully resolves both issues.

Resolves LibrePCB#707 
Resolves LibrePCB#608


ubruhin:
For such PRs it would actually be nice to provide some screenshots to make the contribution *saving* me time instead of *taking* me the same time for review than if I implemented it on my own :see_no_evil: For now I did a comparison (old left, new right):

image

I'm not sure if we now have the same problem as described in #608, just between the top layer and inner layer 2 :thinking: But at least in total it shouldn't be worse than before so I'm fine with it.

No comment on the documentation layer color yet because the new color is not applied to Top/Bottom Documentation (the adjusted Board Documentation layer is used very rarely).

BuFran:
according to documentation layer comparison, there is some pictures, old above, proposed below:

obrazek

obrazek

I think it is better differentiated

ubruhin:
Thanks. Although I'm still not fully convinced by the colors, probably it's indeed a little bit better than before :)

Just one thing: Could you change the hex characters to uppercase for consistency with the rest of the code?

PR #1259, opened 2023-10-16
CI: Update to macOS 12

ubruhin:
It seems for macOS 11 the qt5 bottle (binary release) suddenly disappeared which caused Qt to be installed from source which takes ages, leading in CI timeout. I don't know any other solution than upgrading to macOS 12, leading to worse binary compatibility of our releases (there were already several users complaining that LibrePCb doesn't run on macOS 10 anymore, so I think we'll now get additional complains for macOS 11 :sob:).


ubruhin:
> Isn't there a way to build static image with macos 11 and precompiled qt and using the custom image in the CI ? Ie process "the ages" once and reuse it for tomorrow ?

We do exactly this for both Linux and Windows. But unfortunately something like Docker doesn't seem to exist for macOS. At least not at the time when I set up macOS CI, but feel free to check Azure Pipeline docs again :wink:

BuFran:
Just to stick:

[Docker OSX emulation](https://github.com/sickcodes/Docker-OSX)

There are older versions MacOSX, they are docker images and running OSX on hosted Linux so possibility to use in our CI.

Is it runnable under MS Azure ?

ubruhin:
> Is it runnable under MS Azure ?

Might be worth a try, but those images are ~20GB in size :sob: Also it seems to focus on GUI and the actual `docker run` ends up in a QEMU shell, but we need it to end up in the macOS shell. Not sure if this could be done.

PR #1272, opened 2023-11-22
Refactor format_code.sh and add QML formatter

BuFran:
In **first commit** of series, the QML formatter is added so programmer could autoformat modified QML files before submitting pull request

In **second commit** of series, there was plenty copy-pasted code inside that script. Adding new formatters and options can introduce new bugs. Proposed refactoring shortened all copied code to two bash function calls.

The `search_files` function searches for all intended files with specified extensions
they should be checked in the concrete linter.

The `update_file` checks, if the file content has been modified by linter tool and if
it's true, updates file and logs about the change on screen

**NOTE:** The `NEW_CONTENT` is transferred to the `update_file` via global variable, but it can be read from stdin until eof so the linter calls can be one-liners somewhere in the future, but for now the reason is to keep the trace of changes from previous version of script.

**Third commit** allows specification of branch to be compared against in the case of searching file modifications.

In current code, only master branch is hardcoded as reference to compare and track changes against them.
This modifies lot of files when working on branch that are modified by other people in history when not
rooted on master.

I propose adding option `--compare-against <branch_name>` where you can select different branch (or tag or sha)
to compare local modifications against it.

There could be discussion about the argument and variable name. I can rework both.

Probably there can be implementation detail where `ALL` and `COMPARE_AGAINST` have same meaning in the manner
of selecting which method to call, so in future the `ALL` variable can be left out and `COMPARE_AGAINST` will
be set to empty when `--all` argument is passed (KISS principle).

### DIFF
--- a/dev/format_code.sh
+++ b/dev/format_code.sh
@@ -1,11 +1,11 @@
 #!/usr/bin/env bash
-set -euo pipefail
+set -eufo pipefail
 
 # Formats files according our coding style.
 #
 # Usage:
 #
-#   format_code.sh [--all] [--sudo] [--docker]
+#   format_code.sh [--all] [--sudo] [--docker] [--base <base>] [--check]
 #
 # Notes:
 #   - Run the command "./dev/format_code.sh" in the root of the repository.
@@ -18,11 +18,18 @@ set -euo pipefail
 #   - To format all files (instead of only modified ones), add the "--all"
 #     parameter. This is intended only for LibrePCB maintainers, usually you
 #     should not use this!
+#   - To update files modified against some branch <branch> in history, use
+#     option "--base <base>". If this is not specified, the default "master"
+#     branch will be used.
+#   - To only check which files are wrongly formatted, use option "--check".
+#     No file modifications will be made, nonzero exit code will be thrown
+#     instead.
 
 DOCKER=""
 DOCKER_CMD="docker"
 CLANGFORMAT=${CLANGFORMAT:-clang-format}
-ALL=""
+BASE="master"
+CHECK=""
 for i in "$@"
 do
 case $i in
@@ -35,8 +42,17 @@ case $i in
   shift
   ;;
   --all)
-  ALL="--all"
+  BASE=""
   shift
+  ;;
+  --base)
+  shift
+  BASE=$1
+  shift
+  ;;
+  --check)
+  CHECK="--check"
+  shift;
   ;;
 esac
 done
@@ -55,13 +71,56 @@ if [ "$DOCKER" == "--docker" ]; then
   $DOCKER_CMD run --rm -t --user "$(id -u):$(id -g)" \
     -v "$REPO_ROOT:/code" \
     $DOCKER_IMAGE \
-    /usr/bin/env bash -c "cd /code && dev/format_code.sh $ALL"
+    /usr/bin/env bash -c "cd /code && dev/format_code.sh --base \"$BASE\" $CHECK"
 
   echo "[Docker done.]"
   exit 0
 fi
 
 COUNTER=0
+
+# This function searches for all files for a given Git match pattern.
+# It searchs sorted for tracked files first, then following untracked files.
+#
+# All found files are put on stdout so it is parsable by foreach loop.
+search_files() {
+  if [ "$BASE" == "" ]; then
+    TRACKED=$(git ls-files -- $@)
+  else
+    # Only files which differ from the base branch
+    TRACKED=$(git diff --name-only ${BASE} -- $@)
+  fi
+  UNTRACKED=$(git ls-files --others --exclude-standard -- $@)
+
+  for file in $TRACKED $UNTRACKED; do
+    if [ -f "$file" ]; then
+      echo "$file"
+    fi
+  done
+}
+
+# This function tracks modifications of file and prints out information that
+# file has been processed by the script. It increments processed file counter.
+#
+# Note: Do NOT use in-place edition of the tools because these causes "make"
+# to detect the files as changed every time, even if the content was not
+# modified! So we only overwrite the files here if their content has changed.
+update_file() {
+  NEW_CONTENT=$(cat)
+  OLD_CONTENT=$(cat "$1")
+
+  if [ "$NEW_CONTENT" != "$OLD_CONTENT" ]; then
+    if [ "$CHECK" == "" ]; then
+      printf "%s\n" "$NEW_CONTENT" > "$1"
+    fi
+    echo "[M] $1"
+    COUNTER=$((COUNTER+1))
+  else
+    if [ "$CHECK" == "" ]; then
+      echo "[ ] $1"
+    fi
+  fi
+}
 
 # Format source files with clang-format and Python 3.
 clang_format_failed() {
@@ -73,33 +132,9 @@ clang_format_failed() {
   exit 7
 }
 echo "Formatting sources with $CLANGFORMAT and Python..."
-for dir in apps/ libs/librepcb/ tests/unittests/
-do
-  if [ "$ALL" == "--all" ]; then
-    TRACKED=$(git ls-files -- "${dir}**.cpp" "${dir}**.hpp" "${dir}**.h")
-  else
-    # Only files which differ from the master branch
-    TRACKED=$(git diff --name-only master -- "${dir}**.cpp" "${dir}**.hpp" "${dir}**.h")
-  fi
-  UNTRACKED=$(git ls-files --others --exclude-standard -- "${dir}**.cpp" "${dir}**.hpp" "${dir}**.h")
-  for file in $TRACKED $UNTRACKED
-  do
-    if [ -f "$file" ]; then
-      # Note: Do NOT use in-place edition of clang-format because this causes
-      # "make" to detect the files as changed every time, even if the content was
-      # not modified! So we only overwrite the files if their content has changed.
-      OLD_CONTENT=$(cat "$file")
-      NEW_CONTENT=$($CLANGFORMAT -style=file "$file" || clang_format_failed)
-      NEW_CONTENT=$(echo "$NEW_CONTENT" | "$REPO_ROOT/dev/format_code_helper.py" "$file" || clang_format_failed)
-      if [ "$NEW_CONTENT" != "$OLD_CONTENT" ]
-      then
-        printf "%s\n" "$NEW_CONTENT" > "$file"
-        echo "[M] $file"
-        COUNTER=$((COUNTER+1))
-      else
-        echo "[ ] $file"
-      fi
-    fi
+for dir in apps/ libs/librepcb/ tests/unittests/ share/; do
+  for file in $(search_files "${dir}**.cpp" "${dir}**.hpp" "${dir}**.h" "${dir}**.js"); do
+    $CLANGFORMAT -style=file "$file" | "$REPO_ROOT/dev/format_code_helper.py" "$file" | update_file "$file" || clang_format_failed
   done
 done
 
@@ -113,29 +148,9 @@ ui_format_failed() {
   exit 7
 }
 echo "Formatting UI files with Python..."
-for dir in apps/ libs/librepcb/ tests/unittests/
-do
-  if [ "$ALL" == "--all" ]; then
-    TRACKED=$(git ls-files -- "${dir}**.ui")
-  else
-    # Only files which differ from the master branch
-    TRACKED=$(git diff --name-only master -- "${dir}**.ui")
-  fi
-  UNTRACKED=$(git ls-files --others --exclude-standard -- "${dir}**.ui")
-  for file in $TRACKED $UNTRACKED
-  do
-    if [ -f "$file" ]; then
-      OLD_CONTENT=$(cat "$file")
-      NEW_CONTENT=$(echo "$OLD_CONTENT" | "$REPO_ROOT/dev/format_code_helper.py" "$file" || ui_format_failed)
-      if [ "$NEW_CONTENT" != "$OLD_CONTENT" ]
-      then
-        printf "%s\n" "$NEW_CONTENT" > "$file"
-        echo "[M] $file"
-        COUNTER=$((COUNTER+1))
-      else
-        echo "[ ] $file"
-      fi
-    fi
+for dir in apps/ libs/librepcb/ tests/unittests/; do
+  for file in $(search_files "${dir}**.ui"); do
+    cat "$file" | "$REPO_ROOT/dev/format_code_helper.py" "$file" | update_file "$file" || ui_format_failed
   done
 done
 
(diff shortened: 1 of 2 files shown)


dev/format_code.sh
```diff
@@ -38,6 +40,11 @@ case $i in
   ALL="--all"
   shift
   ;;
+  --compare-against)
```
ubruhin, line 43: Just as an idea, what would you think about renaming it to `--base`? In version control we usually talk about the _base_ branch (or commit) for such things so I think this would still be very intuitive and at the same time it's even shorter :)

Of course the `COMPARE_AGAINST` should then be renamed as well.

dev/format_code.sh
```diff
@@ -63,6 +70,57 @@ fi
...
+  MASKS=""
+  while (($#)); do
+    MASKS+=" ${DIRECTORY}**$1"
+    shift
+  done
+
+  if [ "$ALL" == "--all" ]; then
+    TRACKED=$(git ls-files -- $MASKS)
+  else
+    # Only files which differ from the master branch
```
ubruhin, line 91:
```suggestion
    # Only files which differ from the base branch
```

(or any other way to make the statement more accurate)

dev/format_code.sh
```diff
@@ -182,28 +186,35 @@ xmlsort_failed() {
...
 echo "Formatting resource files with xmlsort..."
-if [ "$ALL" == "--all" ]; then
-  TRACKED=$(git ls-files -- "**.qrc")
-else
-  # Only files which differ from the master branch
-  TRACKED=$(git diff --name-only master -- "**.qrc")
-fi
-UNTRACKED=$(git ls-files --others --exclude-standard -- "**.qrc")
-for file in $TRACKED $UNTRACKED
+for dir in img/ libs/
```
ubruhin, line 189: Hm, previously the file search was not restricted to these directories, why restricting it now?

Generally I prefer to format all files in the repository, and only blacklisting those coming from third-parties (that's why we limit *.cpp/*.h files to our own source directories).

dev/format_code.sh
```diff
@@ -182,28 +186,35 @@ xmlsort_failed() {
...
+qml_format_failed() {
+  echo "" >&2
+  echo "ERROR: qmlformat failed!" >&2
+  echo "  Make sure that qmlformat (qtdeclarative5-dev-tools) are installed." >&2
+  echo "  On Linux, you can also run this script in a docker" >&2
+  echo "  container by using the '--docker' argument." >&2
+  exit 7
+}
+echo "Formatting QML sources with qmlformat..."
+for dir in share/
```
ubruhin, line 209: Same here, I wouldn't limit this to the share directory, there don't seem to be any third-party qml files in the repository.

dev/format_code.sh
```diff
@@ -182,28 +186,35 @@ xmlsort_failed() {
...
+do
+  for file in $(search_files "${dir}" .qml)
+  do
+    if [ -x "${QMLFORMAT}" ]; then
```
ubruhin, line 213: If I understand correctly, this will silently ignore qml files if `qmlformat` is not found. I agree the script should not abort in that case, but I'd highly prefer to see some message in the output that qml files could not be sorted. Isn't this already the case thanks to `qml_format_failed`, so the missing `qmlformat` does not need to be handled separately?

ubruhin (changes requested):
Looks pretty good, the refactoring makes definitively sense!

In addition to the comments: Am I correct that the filter pattern passed to `git` now leads to different output? For example this call:

    git ls-files -- "**CMakeLists.txt" "*.cmake"

is now replaced by this:

    git ls-files -- CMakeLists.txt .cmake

which leads to different result when running manually in terminal. So for me it looks like the result is not correct anymore. Same applies to all other filters, for example the recursive glob `**` is no longer used in the filter pattern. Couldn't we simply keep the current filter patterns to be sure it behaves identical as before?

Another thing: I saw this option in `qmlformat --help`:

````
-l, --newline <newline>  Override the new line format to use (native macos
                           unix windows).
````

I don't know what is the default value, but it might be `native`. This is not desired (we always want UNIX line endings) so it might be worth explicitly passing this flag.

Btw, I didn't test the script yet by running it. Will test its behavior after these things are clarified.

File: dev/format_code.sh (hunk 1 above)
Comment (ubruhin) [reply dev/format_code.sh:43]: Ah and btw, this new CLI flag should also be added to the documentation comment block at the top of the script ;)

File: dev/format_code.sh (hunk 4 above)
Comment (BuFran) [reply dev/format_code.sh:209]: The problem is that in other case it formats the cmake-build-debug directory too and it is really huge as it contains a lot of file copies already changed. Mainly in CMakeFiles scripts but its same for other files. With this limit the speed of stylecheck/formater is comparably faster (0.5sec vs 30sec measured by eye)

File: dev/format_code.sh (hunk 3 above)
Comment (BuFran) [reply dev/format_code.sh:189]: The problem is that in other case it formats the cmake-build-debug directory too and it is really huge as it contains a lot of file copies already changed. Mainly in CMakeFiles scripts but its same for other files. With this limit the speed of stylecheck/formater is comparably faster (0.5sec vs 30sec measured by eye)

File: dev/format_code.sh (hunk 1 above)
Comment (BuFran) [reply dev/format_code.sh:43]: :-D There is documentation ? My IDE collapsed it automatically so I didn't realized there is one ... :facepalm: 

the `--base` wording is definitely better so I use that

File: dev/format_code.sh (hunk 4 above)
Comment (ubruhin) [reply dev/format_code.sh:209]: I can't reproduce this problem. The build directory created by qtcreator is located outside the source directory, thus it doesn't cause any troubles. And when building manually, the build directory should be named `build` to ensure it is ignored by Git (its *always* bad if the build directory is not ignored). With this directory ignored, `git ls-files` should not be slow.

Or am I misunderstanding something?

Note: If you prefer a different name for the build directory, you can add it to `.git/info/exclude` manually to make sure it is ignored as well.

**Comment by BuFran** [comment]
according CMakeLists no, they are replaced by

```
git ls-files -- **CMakeLists.txt **.cmake
```

The only difference is the `"` chars is missing - will be probably problem on windows, needs to be checked

according to `qmlformat -l` I agree and I will update the code.

File: dev/format_code.sh (hunk 5 above)
Comment (ubruhin) [reply dev/format_code.sh:213]: > According to error message, I guess to continue rework to enable some formatters on commandline (so user wants only to format qml, not cpp for example). This PR is just test if You agree with changes and I should continue and do it not overengineer ...

So far I agree with the changes, but to be honest I wouldn't add more options to it. Long-term developers shall install `qmlformat` anyway and for short-term developers it's no problem if there appear many warnings in the output. So in both cases we don't need an additional flag (when adding this flag, the question arises why not making all the other tools optional as well, ending up in several more flags not really needed).

**Comment by BuFran** [comment]
My imagination is to add `--stylecheck` option which disables file writing, just prints "there is problem with file XY" so the `format_code.sh` could be later reused in `stylecheck.sh` which needs some touches too (in another PR)

File: dev/format_code.sh (hunk 5 above)
Comment (BuFran) [reply dev/format_code.sh:215]: `qml_format_failed` solves it. :facepalm: thanks.

File: dev/format_code.sh (hunk 4 above)
Comment (BuFran) [reply dev/format_code.sh:211]: default build directory for CLion IDE is `cmake-build-debug` and `cmake-build-release` depending on the target, and it is in the root directory of the project. But yes, probably the `.git/info/exclude` solution will work for me. I will just try.

File: dev/format_code.sh (hunk 3 above)
Comment (BuFran) [reply dev/format_code.sh]: Restriction removed.

File: dev/format_code.sh (hunk 4 above)
Comment (BuFran) [reply dev/format_code.sh]: Restriction removed

**Comment by BuFran** [comment]
According to `$(git ls-files -- "**CMakeLists.txt" "*.cmake")` line, there is really diference that in new script it runs `$(git ls-files -- **CMakeLists.txt **.cmake)`

There is really functional difference (good point :+1:), that in previous version the `CMakeLists.txt` are parsed **recursively** by bash file globbing, but `.cmake` was included it only from current directory.

It makes me wonder if this was intended behavior or be an mistake as there are .cmake files inside subdirectories already. In my opinion, the second can be true so my code corrected bug.

NOTE: For the sake of completeness, the ** are parsed differently in bash 4.0 and in 5.0, there is well known problem with symbolic links and recursion. But as far as I know, we don't use symlinks anywhere.

**Comment by ubruhin** [comment]
> There is really functional difference (good point 👍), that in previous version the CMakeLists.txt are parsed recursively by bash file globbing, but .cmake was included it only from current directory.
>
> It makes me wonder if this was intended behavior or be an mistake as there are .cmake files inside subdirectories already. In my opinion, the second can be true so my code corrected bug.

Both filters `*.cmake` and `**.cmake` lead to the same output of `git ls-files` because no directory is specified as a prefix. So the old version of the script is working as intended even without using recursive glob.

Anyway, please make sure to pass the asterisks as strings to `git ls-files`, they must not be interpreted as globbing by bash. This is important because we want to fetch the list of files from Git, not from the file system.

File: dev/format_code.sh [hunk 6]
```diff
@@ -182,28 +190,27 @@ xmlsort_failed() {
...
+}
+echo "Formatting QML sources with qmlformat..."
+for file in $(search_files "" .qml); do
+  ${QMLFORMAT} -l unix "$file" | update_file "$file" || qml_format_failed
 done
 
 echo "Finished: $COUNTER files modified."
+
+# nonzero exit code
+if [ "$CHECK" == ""] && [ $COUNTER -ge 0 ]; then
```
ubruhin, line 214: Are you sure this is correct? :see_no_evil: It looks like `exit 1` is executed if check mode is **not** enabled (instead of enabled) and if `$COUNTER` is **greater or equal** than zero (which is **always** the case). Or are my Bash skills that bad? :joy:

ubruhin:
> What is your opinion about deleting ALL variable by reusing BASE (if BASE is empty it means no branch filtering, so output all files) ?

Sounds good :+1:

> according to qmlformat -l I agree and I will update the code. There should be .qmlformat.ini config file same as .clang-format config file, but I didn't generated it yet just for case you have different version...

Hm initially I thought that's a great idea, but then I realized that `qmlformat` is a very new tool which didn't even exist before Qt 5.15 and even with 5.15 it seems to be lacking many options - including INI file parsing as far as I can see. So this is not usable for us yet.

Another thing I realized is that `qmlformat` is not installed in the `devtools` Docker image so it won't work with the `--docker` flag. This is an absolute must-have requirement to merge this PR, so it needs to be installed in the Docker image. I just suspect this is not trivial since there doesn't seem to be an Ubuntu version which has both `qmlformat` and `clang-format-6` available :wink:

BuFran:
:upside_down_face: So isn't the time to upgrade `clang-format-6.0` to newer supported version ? Current stable version is `clang-format-14.0` :upside_down_face: 

We can support qmlformat when it is available, but when isn't it can silently skip the qml format step. That is my recommendation for this issue. Man who develops QtQuick will probably have newer version, and for development on old code, it is not needed to format files the programmer will not touch :-)

AFAIK as I tested QtQuick (that your new UI you like to root at) contains in Qt5 nasty display bugs and in qt5 is really not usable therefore, so it will need to upgrade everything to Qt6 soon, (where the qmlformat is mandatory component). Even on my linux it is impossible to install `clang-format-6.0` version, so I'm using `clang-format-14.0` with four-line hack. And if You don't like my formatting you can always run your version and push "format code" commit with few little changes. Oldest supported on my platform is 8.0.

After merging this PR (with silence on missing qmlformat) it will be one small step forward ...

BuFran:
After some search, if this [this Dockerfile](https://github.com/LibrePCB/docker-librepcb/blob/master/devel/Dockerfile) is the dockerfile for that ubuntu you mentioned (i have no docker at all, due to space and security considerations), seems you are using 16.04 which **isn't supported from 2021** [Ubuntu Releases](https://wiki.ubuntu.com/Releases). Even 18.04 is unsupported for now, 20.04 will end of support 2025 so my opinion is to switch to supported operating system (and keep this old reference system only as backlog for "special" unmaintained projects which support died early and we - you - like to stay compilable).

Citing from official web:

>It is important to take action – either by migrating to the next LTS or upgrading to Ubuntu Pro. Unless you take action, your 18.04 LTS machines will not receive any security updates after 31 May 2023.

Have you purchased Ubuntu Pro in the docker image to have LTS support ($225.00/year)? You really like to open all of vulnerabilities of the system for each user who like to use LibrePCB on his computer ?

anyways, the package you are searching is `qtdeclarative5-dev-tools` which contains `qmlformat`, probably it is still available on 16.04

BuFran:
Ah sorry, probably [This repo](https://github.com/LibrePCB/docker-librepcb-dev) is for docker and there is Ubuntu 18, still EOS but this year on the edge of the support. I didn't find that before as i'm not using docker.

BuFran:
And seems that the qmllint is on not packaged version in already installed [link](https://download.qt.io/online/qtsdkrepository/linux_x64/desktop/qt5_5152/qt.qt5.5152.gcc_64/5.15.2-0-202011130601qtdeclarative-Linux-RHEL_7_6-GCC-Linux-RHEL_7_6-X86_64.7z) from your ubuntu docker file with another path than packaged qt5.15 from newer ubuntu

```
/opt/qt/5.15.2/gcc_64/bin/qmllint
```

So what I need is to test if you are using this version of qmllint and hardcode this path to my script and be compatible with you :-)

ubruhin:
@BuFran The devtools image is now updated and qmlformat installed (#1273, #1276). Could you rebase this branch to the current `master` now? It might lead to a conflict but it should be easy to resolve (just see my changes in the linked PRs).

BuFran:
Rebased, no conflict (if there was, it has been resolved automatically). I'm holding my breath if something happened... 

As I pushed a lot of fixes You recommended, I can squash all changes into single commit to be mergable (there can be better commit message as this PR moved through time) if You like.

ubruhin:
I started to test it, but unfortunately the Docker mode doesn't work anymore:

```
$ ./dev/format_code.sh --docker --all
[Re-running format_code.sh inside Docker container]
./dev/format_code.sh: line 71: ALL: unbound variable
```

You really should get Docker running on your system, it makes things soooooo much easier :see_no_evil: 

Regarding quotes, couldn't you just add double quotes around the values appended to `$MASKS` by escaping them? It's easier to just do the same thing as before (which is known to be working) instead of changing something we don't know if it behaves the same way...

BuFran:
Hello,

According to `$MASKS`, the arguments are quoted by single quote. It's not single string, but multiple strings, each quoted. Each quoted string is put inside git command line as separate argument, and due to single quotes, the argument globbing is parsed by git, not the shell.

I have no idea, where the single quotes comes from. I have not written them anywhere, but when you `set -x` before commands, they are here. Each my try to fill double quotes in the while loop it is crippled by single quotes following double quotes, thus not working.

The `$ALL` is real revealed bug.

Anyways, you may close this PR not merging, same as license PR, I am giving up any volunteering on this project as the direction is moving toward huge containers I have no space for, and towards technologies which are not well implemented yet and has big rendering issues in current operating systems (QtQuick - I'm against this technology stronger than before after some testing on your dev branches). 

It's nice piece of software in current state and I apologize I like to help you by implementing something or trying to correct bugs.

I wish You happy christmas time!

ubruhin:
> I have no idea, where the single quotes comes from. I have not written them anywhere, but when you set -x before commands, they are here. Each my try to fill double quotes in the while loop it is crippled by single quotes following double quotes, thus not working.

Ah OK I see :+1: So once `--all` is working, I'll test if the `ls-files` is leading to the correct result. If it looks good, I'm fine with it.

> Anyways, you may close this PR not merging, same as license PR, I am giving up any volunteering on this project as the direction is moving toward huge containers I have no space for,

Hm sad to hear that :-/ Thanks anyway for your time already spent for LibrePCB.

> and towards technologies which are not well implemented yet and has big rendering issues in current operating systems (QtQuick - I'm against this technology stronger than before after some testing on your dev branches).

May I ask you for more details about this? Could also be a private conversation on Telegram. I'd be really interested in your findings about QtQuick. For me, it's still a very hard decision which GUI toolkit makes most sense for the future. I'm neither happy with QtWidgets nor with QtQuick, nor any other GUI toolkit I have seen yet. In the end, it's a choice of the least of the evils. And my decision is still not definitive, if there are more reasons against QtQuick than against QtWidgets we can still stick with QtWidgets. But I need more input to make the correct decision.

> I wish You happy christmas time!

Thanks you, happy christmas time too! :sparkles:

ubruhin:
Since the refactoring makes sense anyway (no matter whether we will need `qmlformat` in future or not), I don't like to close this PR. So I tested this in more detail and fixed some issues. Globbing was still a problem, now after disabling it with `set -f` it works properly. I think now everything works fine so let's merge it. Thanks again @BuFran.

PR #1277, opened 2023-12-12
Fix file extension in Protel-style Gerber output job

ubruhin:
The Protel-style suffix for the board outlines layer was "gml" instead of "gm1". Very nasty bug, I hope it doesn't lead to serious issues for users :sob:


dbrgn:
Wait, what, it's really "gm1"? 😂🙈

dbrgn:
Seems it's not so clear: https://bugs.launchpad.net/kicad/+bug/1694049

ubruhin:
Yeah the file extensions of Gerber is really a mess. No standard at all. Although Gerber specifies `.gbr` as official extension, it doesn't specify how to put the layer name into the file name.

Anyway, our legacy Gerber export dialog used `.gm1`, so the new Gerber output job shall use the same extension...

PR #1307, opened 2024-02-15
Initialize new workspace with some example projects

### LINKED ISSUE #1267: Provide example projects
##### SUMMARY

For evaluation of an EDA tool, it's helpful to open an existing project to play around with. It gives a much better impression of the capabilities of the tool compared to creating a new project from scratch. Thus it would be nice to provide some example projects which can be opened from the control panel.

##### SOLUTION

I think there are two main questions:

- Should the example projects be shipped together with the application, or downloaded from the internet on demand? Although I like having resources available offline, I tend to the online solution here to avoid bloating the application with data that is only useful during evaluation. But the question then is how to discover and where to download these projects from.
- Where should example projects be stored for opening them? If they are bundled together with the application, they will be read-only so they should be copied to a writable location before opening them. Probably they should just be copied into the workspace's projects folder? Maybe into a subdirectory called `examples`? In the end it's easy to just delete the projects from the control panel after evaluation.

We could also consider extending the workspace creation wizard with a checkbox "add example projects" which automatically adds one or a few example projects to the newly created workspace.

ubruhin:
So far, new workspaces didn't contain any projects after their initial creation so the Control Panel project tree was completely empty. However, as suggested in #1267 it would be useful to have some example projects available to help new users evaluating LibrePCB. Thus this PR implements this feature in two ways:

### Auto-Initialize New Workspaces

The initial idea was actually to provide a way to open example projects, for example through a menu item *File -> Open -> Example Project*. However, the question was where they will be opened from and if they are read-only or writable. Also the menu option might not be visible and convenient enough for new users to make them actually using this feature.

Now I decided to go with a different approach. When creating a new workspace, just automatically add some example projects in a subdirectory named "Examples" to the workspace. So the projects will immediately be shown in the Control Panel for new users, ready to be opened like normal projects (also writable!). One might argue that not everyone likes to have these example projects, but in the end it's very easy to just delete them and you only need to do this once. Also I remember EAGLE did it the same way and it never felt annoying to me.

The only question left was whether these example projects shall be bundled with the application or downloaded from the internet on demand. Since this feature is not really important (no problem if the download takes some time or even fails), I decided for the on-demand download to avoid bloating the application with example projects. We just should store them at some reliable, persistent URL.

### Manually Add Example Projects

To also allow adding the example projects to existing workspaces, the Control Panel contains a new menu item *File -> Import -> Add Example Projects* which initiates the download and installation of these projects asynchronously, exactly the same way as described above.

### Demo

librepcb-example-projects

### ToDo

Decide which and how many example projects to install, and make sure to host them at a reliable location.

Closes #1267

### DIFF
--- a/libs/librepcb/editor/editorcommandset.h
+++ b/libs/librepcb/editor/editorcommandset.h
@@ -448,6 +448,15 @@ public:
 
   EditorCommandCategory categoryImportExport{
       "categoryImportExport", QT_TR_NOOP("Import/Export"), true, &categoryRoot};
+  EditorCommand addExampleProjects{
+      "add_example_projects",  // clang-format break
+      QT_TR_NOOP("Add Example Projects"),
+      QT_TR_NOOP("Add some example projects to the workspace"),
+      ":/img/logo/32x32.png",
+      EditorCommand::Flag::OpensPopup,
+      {},
+      nullptr,  // Exclude from shortcuts overview & configuration
+  };
   EditorCommand importDxf{
       "import_dxf",  // clang-format break
       QT_TR_NOOP("Import DXF"),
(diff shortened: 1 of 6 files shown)


ubruhin:
Created a new repository to store example projects as *.lppz files: https://github.com/LibrePCB/librepcb-example-projects

For now, let's go with my own projects. We can replace them with community-contributed projects at any time later when we get such contributions. Some requirements for contributions are listed in the mentioned repository.

PR #1313, opened 2024-02-23
Schematic editor: Display live parts information (prices etc.)

ubruhin:
### Summary

This PR adds an initial integration of live parts information to the schematic editor, in particular to the "Add Component" dialog and the "Generate BOM" dialog. Live part information means part lifecycle status, supplier availability, pricing and more. Since this information is very volatile, it is fetched from the internet on demand and not stored in any file (except for caching).

Note that this PR adds the feature to the application (i.e. client-side), but it also requires a server-side implementation which provides the data to be displayed. The server-side implementation is added in https://github.com/LibrePCB/librepcb-api-server/pull/1. And the actual availability of this feature in the application depends on whether the server provides this data or not - if we don't have access to a corresponding database, the server replies with a "not available" status and the application silently doesn't show any life part information.

### API Specification

The underlying API for the server requests/responses is specified here: https://developers.librepcb.org/_branches/live-part-information/d1/dcb/doc_server_api.html#doc_server_api_resources_parts (link might be down if the branch is not pushed for some time).

### Add-Component Dialog

Preview with dummy data since we don't have access to a parts information database yet:

librepcb-live-part-info

### BOM Export

The same information is provided in the BOM export (incl. mouse hover), plus the total price for all parts of the project:

image

### Internet Access

Whenever the user opens a window where life parts information should be displayed, the application requests the information from the configured API server (api.librepcb.org by default). The received data is cached locally to avoid extensive network traffic if the same data is requested multiple times in a short time span.

The part information itself is all requested through the API, i.e. LibrePCB only talks to our own server. This allows us to ensure proper data privacy (no tracking etc.), independent of the underlying data provider. So the data provider only gets anonymous requests from our own server, not from our users directly (keeping e.g. their IP address private). However, images are not embedded in the API responses, so they might (and probably will be) downloaded directly from other domains (e.g. semiconductor companies). I hope this compromise is okay for our users, as it would be a big overhead to also route this traffic through our own server. I think it's not really relevant for privacy since requests do not contain any useful information (just fetching static content) and are spread across many domains, not possible for a company to track all user requests.

However, for users who are concerned about internet traffic, data privacy or something like that, the feature can be disabled in the workspace settings (details are provided as tooltip):

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,12 @@ productive use, please install an official release as described in the
 
 ### Requirements
 
-To compile LibrePCB, you need the following software components:
+To compile and run LibrePCB, you need the following software components:
 
 - g++ >= 4.8, MinGW >= 4.8, or Clang >= 3.3 (C++11 support is required)
-- [Qt](http://www.qt.io/download-open-source/) >= 5.5
+- [Qt](http://www.qt.io/download-open-source/) >= 5.5 (make sure the
+  [imageformats](https://doc.qt.io/qt-5/qtimageformats-index.html) plugin
+  is installed too as it will be needed at runtime!).
 - [OpenCASCADE](https://www.opencascade.com/) OCCT or OCE (optional)
 - [OpenGL Utility Library](https://en.wikipedia.org/wiki/OpenGL_Utility_Library)
   GLU (optional)
@@ -75,7 +77,7 @@ actually used for CI, but are also useful to build LibrePCB locally.
 sudo apt-get install build-essential git cmake openssl zlib1g zlib1g-dev \
      qtbase5-dev qtdeclarative5-dev qttools5-dev-tools qttools5-dev \
      qtquickcontrols2-5-dev libqt5opengl5-dev libqt5svg5-dev \
-     libglu1-mesa-dev liboce-*-dev
+     qt5-image-formats-plugins libglu1-mesa-dev liboce-*-dev
 sudo apt-get install qt5-doc qtcreator # optional
 ```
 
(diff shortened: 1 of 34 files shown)


dbrgn:
Very fancy!

In theory the server could live-proxy the requests, but I also doubt it's worth the effort.

ubruhin:
> In theory the server could live-proxy the requests, but I also doubt it's worth the effort.

True, it's the server's choice where images are loaded from so this privacy feature could in theory be added any time later without changing the API (though privacy "by design" would be nicer).

Btw for the record, the PR is marked as draft since I might change some parts of the API before merging:

- Remove the `suggestions` key from the API (resp. mark it as "preview") since it is not implemented yet and could easily be added some day later without introducing a breaking change.
- I need to think about how to handle the `provider_logo_url` for dark window themes - if the logo contains dark text on transparent background, it wouldn't be readable with a dark window theme. Maybe a `provider_logo_dark_url` should be added, or requiring an SVG which can be recolored client side :thinking: 

Another question might be whether this new feature should be opt-in instead of opt-out (maybe asking it in the first-run wizard), but so far I think opt-out is okay due to the privacy respecting implementation (also the server-side implementation is open-source!). So from my point of view, I'd go with opt-out for now and listen for user feedback.

klaids:
Hi, A great feature!! 
Do you think it also makes sense to implement creating a manual solution based with adding a price manually. For example, when a PN does not have an online match but we still want to keep track of it on the BOM. 
So it will also be possible using an API to upload prices that are managed internally through quotations, right?

ubruhin:
Hi @klaids,

> Do you think it also makes sense to implement creating a manual solution based with adding a price manually. For example, when a PN does not have an online match but we still want to keep track of it on the BOM.

Currently there are no concrete plans for such a feature, but I also thought about it already and I agree it would be useful in some cases. The online API is a first step, but maybe some day we extend this feature with more functionality.

>  So it will also be possible using an API to upload prices that are managed internally through quotations, right?

Hm, I'm not sure if I understand your question correctly. Uploading a price through the API is not planned (would be difficult to ensure correctness of the uploaded prices). But theoretically it will be possible to set up your own API server as kind of proxy where you can deliver your own prices (e.g. from quotations) to LibrePCB and retrieving the remaining prices from our API server. For a company this might make sense (also to deploy your own libraries) but of course involves some effort to set up the API server.

ubruhin:
Small update on the API:
- Marked `suggestions` key as proposal
- Added `provider_logo_dark_url` as a proposal (we don't need it yet)
- Added `info_url` key, identically to the `/order` endpoint which already contained this key

PR #1314, opened 2024-02-24
Update AppStream MetaInfo to conform to latest Freedesktop spec

Alexander-Wilms:


share/metainfo/org.librepcb.LibrePCB.metainfo.xml
```diff
@@ -5,7 +5,9 @@
   <project_license>GPL-3.0</project_license>
   <name>LibrePCB</name>
   <summary>Design Schematics and PCBs</summary>
-  <developer_name>The LibrePCB Developers</developer_name>
+  <developer id="org.librepcb">
```
ubruhin, line 8: Shouldn't this be the other way around?

```suggestion
  <developer id="librepcb.org">
```

At least this is what the [specs](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer) mention as an example:

> The element should have a id property, containing a unique ID to identify the respective developer. It is recommended to use a web URL, like gnome.org as ID to achieve a higher chance of uniqueness.

And btw, I worry a bit if this new `<developer>` tag also works on older systems resp. if it will be a problem if it doesn't :thinking: But hopefully in worst case it would just lead to this information being hidden in some software center UI...

dbrgn: Are we reading the same spec? Was the spec just updated? 😄 

> The element should have a id property, containing a unique ID to identify the respective developer. It is recommended to use a reverse-DNS name, like org.gnome or io.github.ximion, or a Fediverse handle (like @user@example.org) as ID to achieve a higher chance of uniqueness.

Alexander-Wilms: I should have mentioned that I based this on this edit that will be part of the next minor release of the spec:

https://github.com/ximion/appstream/commit/4537dceb743d010fd79b59b29c6ad9ab3b7dc950

Also, I think one can just keep the old tag for some time in addition to the new one until all big distros use AppStream 1.x

ubruhin: Ahh seems we are bleeding edge xD

> Also, I think one can just keep the old tag for some time in addition to the new one until all big distros use AppStream 1.x

Sounds reasonable! :+1:

PR #1327, opened 2024-03-08
Adding multithreading to design rule checker

DoctorScience:
Added multi-threading to the DRC and some changes to the emitProgress function to allow design steps to allow checks to be completed in any order. Each check is simply given its own thread. It sped up a DRC on my project by about 70%. I made this patch for my personal use but hopefully there's something useful in there :)


ubruhin:
Well, I wish it is that simple :wink: But unfortunately this solution is not safe at all. The `Board` object and its whole content is not thread-safe. So crashes are only a matter of time. Also the global variables do not take into account that multiple boards can be opened at the same time.

Good if it works for you and thanks for your motivation to contribute it, but it's nothing we can merge :see_no_evil: Thus closing this.

PR #1316, opened 2024-02-27
Add licensing data for https://reuse.software/

FedericoCeratto:
This could be used to later remove `img/README.md` and simplify license management using `reuse lint` in CI
Please review the licensing data before merging.
The project can then be registered at https://api.reuse.software/register to obtain a badge similar to https://api.reuse.software/info/codeberg.org/FedericoCeratto/rrdnsd

PR #1336, opened 2024-03-20
Replace Icon8 icons

FedericoCeratto:
The following icons are from https://icons.getbootstrap.com/ under MIT license:
img/actions/export.png
img/actions/import.png
img/actions/view_3d.png
img/library/part.png
img/actions/move_to.png

The following are made by me with parts from the bootstrap collection, under MIT license:
img/actions/clone.png
img/actions/export_pick_place_file.png
img/actions/flip_horizontal.png
img/actions/flip_vertical.png
img/actions/ruler.png
img/actions/drc.png
img/actions/export_pixmap.png
img/actions/order_pcb.png


ubruhin:
Hmm I appreciate your work, but to be honest I'm not confident with some of these icons :see_no_evil: 

The only one I totally agree with is `export_pick_place_file.png`, it is even more precise than the old icon.

For `clone.png` and `export_pixmap.png` IMHO they are not descriptive enough. The clone icon looks like a normal "copy" operation, but cloning means something different than copying. And for the pixmap it is important to express *pixel format* (like photos, thus the old icon), but just a schematic does not imply any image format.

The `ruler.png` looks not bad, but in the toolbar it's simply too small, the ticks are almost not visible.

And all the monochrome icons (although some look really nice) have two problems: They are not readable on dark themes (could be fixed by also providing a white variant) and they make the UI even more inconsistent. To me, the toolbars look really ugly with this mix of different icon styles:

image

Note that for long term I'm even considering using monochrome-icons only since they work very well on dark themes. But for short-term I really want to keep colored icons to keep the look more or less consistent.

FedericoCeratto:
@ubruhin based on the conversation on Telegram I understood you seemed happy to consider Bootstrap and I starting converting **all** icons and drawing new ones in SVG and PNG. Please let me know if I should stop or continue :)
Regarding the readability, AFAIK monochrome icons are used because they can be flipped to negative at runtime as needed and also painted with any hue to make applications "themed".

ubruhin:
Hmm yes I am happy considering Bootstrap icons, but I thought this PR is only about getting rid of Icons8 icons because of their license :thinking: Changing *all* icons though is a much bigger task, probably taking a long time to get finished (also with some uncertainty whether we really like monochrome icons or not, this needs to be tested).

Also inverting colors at runtime is an additional task which would need to be implemented and tested (I already did a quick test some time ago, but it was taking way too much time so I decided to use separate PNGs for dark and light themes). SVG is another topic which needs to be investigated (performance?).

So I'd like to do this step by step, beginning with clarifying licenses (#1316), then getting rid of icons8 (this PR) and for long-term monochrome icons (maybe together with a dark-only LibrePCB theme, then we even don't need black icons).

ubruhin:
The new UI now contains monochrome icons from Bootstrap and Font Awesome, and all Icons8 images have been removed. So this PR is no longer needed.

PR #1351, opened 2024-03-30
Control panel: Add donate button

ubruhin:
Not sure if this is too prominent or not, but donations are important for the long-term funding of LibrePCB and probably most users are not even thinking about donating because in the application itself (in contrast to the website) it's not promoted at all. I tried to implement it in a non-disruptive way to hopefully not annoy users. Anyway, I'm open for opinions.

image

The button opens https://librepcb.org/donate/ in a browser.

Initially I wanted to implement it in the status bar (very small) but couldn't find a good looking solution.


dbrgn:
Looks good to me 🙂

xotho:
I like, 
**it does not matter the position since it makes it's job** ....

X!

ubruhin:
> Anyhow, the right position without comments:

That would also be my preferred position but it's not possible to place a button in the menu bar (for example on macOS, the menu bar is not even part of the application window).

xotho:
....
image

Do not get mad, it work fine so.

take care,
X!

Issue #1390, opened 2024-05-19
Add auto dark mode for Windows


mi4code:
##### SUMMARY

![(ASUS-PC, 2024-05-19 20-23-38)](https://github.com/LibrePCB/LibrePCB/assets/114027049/93078a51-951d-4a0a-9458-59a4c08ac71b)

##### SOLUTION

I can open PR for my current implementation if desired but I think that additional settings entry should be added to have option to disable this feature (and I have no idea how to do that so I will need some help).

ubruhin:
Hi,

can you quickly describe how you implemented that? Or feel free to open a PR with your current implementation. I'm not sure yet if and how this feature should be configurable, first I'd like to test the look&feel on my own. Also I'm curious what you mean with "auto", does Qt detect if Windows is set to a dark mode or is there any manual detection required?

ubruhin:
OK thanks for the PR. So basically everything is manually done (detecting Windows configuration & applying different colors).

I found [here](https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5) that Qt now has a dark theme built-in, which can be tried by passing `-style fusion` to the `librepcb.exe`. It only works with our 64-bit (Qt6) build, but that's fine as 32-bit (Qt5) builds are deprecated.

I quickly tried this and it seems to work very well. So probably the only thing is to detect the Windows theme setting, and if dark then enable the Fusion style. No need for changing colors manually.

Still not sure about configurability... maybe for now it's good enough to ignore it. If the user selected dark theme, he is probably not interested in light theme of LibrePCB :thinking:

mi4code:
> I quickly tried this and it seems to work very well. So probably the only thing is to detect the Windows theme setting, and if dark then enable the Fusion style. No need for changing colors manually.

Actually the Fusion style setting already does this (it uses dark/light theme according to the system) so using it on Windows by default would solve everything.

Only downside would be that the light theme will become slightly different - see comparsion below (left current, right fusion).

![image](https://github.com/LibrePCB/LibrePCB/assets/114027049/227c71c7-d8cd-47f8-8596-ed31ade6cc5a)

ubruhin:
> Only downside would be that the light theme will become slightly different - see comparsion below (left current, right fusion).

[...]

PR #1391, opened 2024-05-20
Implement automatic dark theme for Windows

mi4code:
#1390


ubruhin:
For the record, I added a commit as described in https://github.com/LibrePCB/LibrePCB/issues/1390#issuecomment-2139250928.

Thanks for initiating this! :tada:

xotho:
I confirm I like, maybe some accent should have to be reviewed.

But I Like !!!

thank you.
X

PR #1395, opened 2024-05-26
Replace pick&place icon from Icons8

FedericoCeratto:
Based on the discussion on Telegram, @ubruhin preferred colorful icons to keep the same style as the existing ones.
I tried to make the icons a bit "3D" using gradients.


ubruhin:
Thanks for your ongoing work on this!

Note that my comment from https://github.com/LibrePCB/LibrePCB/pull/1336#issuecomment-2014953650 still applies:

> The only one I totally agree with is export_pick_place_file.png, it is even more precise than the old icon.
>
> For clone.png and export_pixmap.png IMHO they are not descriptive enough. The clone icon looks like a normal "copy" operation, but cloning means something different than copying. And for the pixmap it is important to express pixel format (like photos, thus the old icon), but just a schematic does not imply any image format.
>
> The ruler.png looks not bad, but in the toolbar it's simply too small, the ticks are almost not visible.

The following images look not bad on light theme, but they don't work well on dark themes:

- `export.png`
- `import.png`
- `move_to.png`
- `view_3d.png`

For the remaining icons I'm not convinced yet as well to be honest. I think the old ones look better / more intuitive. Maybe `drc.png`, `flip_horizontal.png` and `flip_vertical.png` could easily be drawn manually in a similar way to the old ones (instead of using bootstrap icons) since they are quite simple :thinking: 

I'm really not sure if it makes sense to continue working on this as it turns out to be very hard (the old icons are really nice, and designing new ones which are intuitive and look good on both light & dark themes is difficult). If the license is a big problem, what about applying your suggested icons as a patch for the Debian package? I mean it would not be forever, the patch could be removed once we switch to a fully monochrome icon theme. Maybe this way would be much more efficient and doesn't affect users on other operating systems...

FedericoCeratto:
Ok, I'll patch them out in the Debian package for the time being. In the meantime do you want to extract export_pick_place_file.png and use it perhaps?

ubruhin:
> Ok, I'll patch them out in the Debian package for the time being.

Great, thanks!

> In the meantime do you want to extract export_pick_place_file.png and use it perhaps?

Yeah I think that would be nice. I could update the PR accordingly if you don't mind (just make sure to create the patch first). Actually I'd also try to tune colors a little bit as they are still not ideal for dark theme, maybe replacing black lines with gray would work better :thinking: 

Just a question: In `dep5` currently `export_pick_place_file.png` is licensed as MIT with comment "based on bootstrap" but actually the icon is entirely drawn by yourself, isn't it? Would you then mind to license it under CC0-1.0 with "Copyright: LibrePCB Developers, see AUTHORS.md" as all our other self-drawn images? I think it would help to keep licensing simple/clean and it's still correct as you're listed in `AUTHORS.md`. But if you don't like it, no worries.

FedericoCeratto:
@ubruhin good spot, `export_pick_place_file.png` is indeed drawn by me, feel free to use and modify it under CC0-1.0 with "Copyright: LibrePCB Developers, see AUTHORS.md".

PR #1404, opened 2024-06-13
Update REUSE copyright metadata

FedericoCeratto:
Minor improvements

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -1,7 +1,6 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: librepcb
 Source: https://librepcb.org/
-Files-Excluded: libs/parseagle/specs/eagle*dtd
 Comment:
  Exclude libs/parseagle/specs/eagle_7.5.0.dtd as it is released
  under CC-BY-ND-3 and not required for building or installing.
@@ -301,19 +300,20 @@ Copyright: Icons8
 License: LicenseRef-Icons8
 
 Files: libs/delaunay-triangulation/*
-Copyright: 2015 Simon Zeni
+Copyright: 2015, Simon Zeni
 License: MIT
 
 Files: libs/dxflib/*
-Copyright: 2001 Robert J. Campbell Jr., 2001-2013 RibbonSoft GmbH
+Copyright: 2001-2013, RibbonSoft GmbH
 License: GPL-2.0-or-later
 
-Files: libs/fontobene-qt/*
-Copyright: Fontobene Developers
+Files: libs/fontobene-qt5/*
+Copyright: 2012-2020, Urban Bruhin
 License: MIT or Apache-2.0
 
 Files: libs/googletest/*
-Copyright: Google and various contributors
+Copyright: 2005-2020, Google Inc and various contributors
+           2018-2019, Google LLC
 License: BSD-3-Clause
 
 Files: libs/googletest/googlemock/scripts/generator/*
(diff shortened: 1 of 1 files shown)


.reuse/dep5
```diff
@@ -301,43 +300,48 @@ Copyright: Icons8
...
 
 Files: libs/dxflib/*
-Copyright: 2001 Robert J. Campbell Jr., 2001-2013 RibbonSoft GmbH
+Copyright: 2001-2013, RibbonSoft GmbH
 License: GPL-2.0-or-later
 
-Files: libs/fontobene-qt/*
-Copyright: Fontobene Developers
+Files: libs/fontobene-qt5/*
+Copyright: 2012-2020, Urban Bruhin
```
ubruhin, line 311: Hmm this change makes no sense, the submodule is called `fontobene-qt`, not `fontobene-qt5` (anymore). Also I'd prefer keeping "Fontobene Developers" as copyright holder.

.reuse/dep5
```diff
@@ -301,43 +300,48 @@ Copyright: Icons8
...
-Copyright: 2008-2014 Natacha Porté, Vicent Martí, Xavier Mendez, Devin Torres and the Hoedown authors
+Copyright: 2008-2014, Natacha Porté, Vicent Martí, Xavier Mendez, Devin Torres and the Hoedown authors
 License: ISC
 
 Files: libs/muparser/*
-Copyright: 2020 Ingo Berg
+Copyright: 2004-2020, Ingo Berg
 License: BSD-2-Clause
 
+Files: libs/polyclipping/
```
ubruhin, line 331: Polyclipping is already listed below, what's wrong with it?

.reuse/dep5
```diff
@@ -301,43 +300,48 @@ Copyright: Icons8
...
+Copyright: 2010-2017, Angus Johnson
+License: BSL-1
+
 Files: libs/optional/*
-Copyright: 2017 by Simon Brand
+Copyright: 2017, Simon Brand <tartanllama@gmail.com>
```
ubruhin, line 336: I personally don't like if my email is published by somebody else, thus I generally also don't publish the email of other people. I think it's not relevant at all in this file anyway.

.reuse/dep5
```diff
@@ -353,10 +357,6 @@ Files: libs/type_safe/*
 Copyright: 2016-2019, Jonathan Müller
 License: MIT
 
-Files: libs/type_safe/external/debug_assert/*
```
ubruhin, line 356: Why removing this?

.reuse/dep5
```diff
@@ -301,43 +300,48 @@ Copyright: Icons8
...
 
 Files: libs/optional/tests/catch.hpp
-Copyright: 2017 Two Blue Cubes Ltd
-License: BSL-1.0
+Copyright: 2017, Two Blue Cubes Ltd
+License: BSL-1
 
 Files: libs/parseagle/*
-Copyright: U. Bruhin
+Copyright: 2012-2020, Urban Bruhin
```
ubruhin, line 344: I prefer to keep the short form "U. Bruhin".

ubruhin (changes requested):
Could you please elaborate a bit why these changes are made? Is it generated somehow?

And why the commas in the Copyright statements? In both the specification and the example provided at https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/ (DEP5) I don't see any comma used :thinking:

.reuse/dep5
FedericoCeratto: Copyright and license appears to be already described by `libs/type_safe/*`

.reuse/dep5
FedericoCeratto: Rolled back.

.reuse/dep5
FedericoCeratto: Good spot, removed!

.reuse/dep5
FedericoCeratto: (Email addresses are recommended, among other things, to prevent ambiguities around the author's identity.) Anyhow, I removed it.

.reuse/dep5
FedericoCeratto: Updated.

FedericoCeratto:
The file is handwritten. The commas are not mandatory but customary: see the `Example 4. Complex` section in https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/ - this is just minor linting to keep the format consistent.

.reuse/dep5
ubruhin: But it's not the same license. type_safe is MIT, while debug_assert is Zlib.

.reuse/dep5
```diff
@@ -1,7 +1,6 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: librepcb
 Source: https://librepcb.org/
-Files-Excluded: libs/parseagle/specs/eagle*dtd
 Comment:
  Exclude libs/parseagle/specs/eagle_7.5.0.dtd as it is released
  under CC-BY-ND-3 and not required for building or installing.
```
ubruhin, line 6: I just realized this comment is related to `Files-Excluded` which was removed, so I think this comment should be removed too.

FedericoCeratto: Updated.

.reuse/dep5
FedericoCeratto: You are right! Confusingly, the `libs/type_safe/external/debug_assert/CMakeLists.txt` file mentions a LICENSE file but I verified on the upstream repo - it's still under Zlib. Updated.

.reuse/dep5
```diff
@@ -301,47 +297,48 @@ Copyright: Icons8
...
 Files: libs/optional/*
-Copyright: 2017 by Simon Brand
+Copyright: 2017, Simon Brand
 License: CC0-1.0
 
 Files: libs/optional/tests/catch.hpp
-Copyright: 2017 Two Blue Cubes Ltd
-License: BSL-1.0
+Copyright: 2017, Two Blue Cubes Ltd
+License: BSL-1
```
ubruhin, line 334: `reuse supported-licenses` lists it as "BSL-1.0".

```suggestion
License: BSL-1.0
```

ubruhin (approved):
Thanks for the updates, looks good to me now.

Ready to merge from your side too?

PR #1411, opened 2024-06-27
Schematic editor: Ask for swapping component names

ubruhin:
Sometimes one wants to explicitly set the name of a component, for example to have components with consecutive numbers close to each other (R1, R2, R3) for readability reasons or to make assembly easier.

However, when renaming a component and the new name already exists in the schematic, an error was shown and one had to rename the other component first to "release" its name. This was very cumbersome.

Now a message box asks to simply swap the components name in this case so LibrePCB doesn't prevent you from renaming a component anymore -> much better user experience.

image


xotho:
It's really usefull !

thank you.

PR #1412, opened 2024-06-28
Package editor: Support copying properties to other objects

ubruhin:
In the package editor it is a common task to modify pad geometries (e.g. width, height or drill diameter) of already added pads. But for packages with a lot of pads, this is *very* annoying because each pad has to be edited individually. So it would be very helpful to allow modifying many pads at once.

I implemented this as a quite generic feature. After copying a single object of any type (pad, polygon, hole, ...) to the clipboard as usual, the context menu contains a new item "Paste Geometry" which applies only the geometry from the copied object to the selected objects. The usual keyboard shortcuts Ctrl+C and Ctrl+V work as well since Ctrl+V automatically detects whether only the geometry shall be applied or the normal paste operation takes place. With this feature, it is now very efficient to apply changes to many objects at once - just modify one object and then copy it to the others.

librepcb-paste-geometry

Note that it is not possible to choose *which* properties will be copy&pasted. Basically everything except placement (X/Y/ROT; and for pads, the package pad assignment) are copied. Btw, the copy&paste works even across tabs/windows, so you can copy the pad geometry of any package in any library to another one.

Unfortunately this new powerful feature is not really promoted in the UI so maybe it is not so intuitive to find it - the (conditional) context menu item is the only visible part of the feature. However, I didn't find a better solution yet :thinking:

### DIFF
--- a/libs/librepcb/editor/graphics/circlegraphicsitem.h
+++ b/libs/librepcb/editor/graphics/circlegraphicsitem.h
@@ -55,7 +55,7 @@ public:
   virtual ~CircleGraphicsItem() noexcept;
 
   // Getters
-  Circle& getCircle() noexcept { return mCircle; }
+  Circle& getObj() noexcept { return mCircle; }
 
   // Operator Overloadings
   CircleGraphicsItem& operator=(const CircleGraphicsItem& rhs) = delete;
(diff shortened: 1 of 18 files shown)


xotho:
Checked the functionality of this feature, I like it.

PR #1431, opened 2024-08-16
Symbol editor: Support mass import of pins through clipboard

ubruhin:
Especially when drawing symbols of ICs with many pins (e.g. microcontrollers), it can be cumbersome and time consuming to manually add every pin individually. Importing them from datasheets would be very helpful, though parsing them out of a PDF is almost impossible. However, I found a semi-automated way to do it through the clipboard: Just select the whole pinout table from a PDF into the the clipboard and let LibrePCB extract the pin names from this piece of text.

Usually the copied table cells will be separated by whitespaces, so LibrePCB interprets the clipboard content as a space-separated list and lets the user choose which column contains the pin names - though by default it even tries to auto-detect the column:

librepcb-import-pins-1

However, this does not work well if table cells contain linebreaks as in the clipboard we can't distinguish them from table rows. For this case, another mode is supported: LibrePCB constantly monitors the clipboard and records every text copied into it. In this mode, just go through the datasheet and copy pin-by-pin into the clipboard:

librepcb-import-pins-2

It's not perfect by far, but I think it already helps to save time in many cases.


dbrgn:
Clever solution 🙂 Not very intuitive UX, but once you learn it, it's going to save a lot of time.

PR #1450, opened 2024-10-23
Package editor: Support displaying datasheet image overlay

### LINKED ISSUE #909: Library editor : Allow overlaying datasheet image for verification
##### SUMMARY

As already said in https://github.com/LibrePCB/LibrePCB/issues/740, creating new component is time consuming and source of error.

##### SOLUTION

Maybe the best solution is to import it from already existing component database.
It will open to a very large set of components.

But if not possible/not desired, is it possible to overlay the image of the datasheet with transparency for checking ?

It will make something like that : (and show that GND7 and GND8 are not at the good place :sob: )

image

ubruhin:
This PR allows to set an image as a semi-transparent overlay in the footprint editor, to verify the footprint geometry with a package drawing (e.g. from a datasheet). Images can be loaded from several sources:

- Take screenshot directly from within LibrePCB (with 3 seconds delay to allow window switching)
- Paste from clipboard
- Load from file

Then the transform of the image needs to be set by a few steps:

1. Crop the image (optional)
2. Mirror & rotate the image, if required
3. Specify two points in the image with their corresponding target coordinates in the footprint. LibrePCB will then calculate X- and Y scale factors and offset for the actual image placement. A sanity check warns e.g. in case the distance in one direction is not large enough, thus would lead in inaccurate scaling.

The image together with all the settings is stored in the user's cache directory (`~/.cache` on Linux) - not in the library element! - and automatically displayed again when opening the package the next time.

For display, the image colors get inverted depending on the window's background color and the image's background color is made transparent so only the drawing should be visible. And actually it's not an overlay but in the background, to avoid hiding footprint geometry in case the automatic transparency doesn't work.

librepcb-bg-image

Closes #909

### DIFF
--- a/libs/librepcb/editor/CMakeLists.txt
+++ b/libs/librepcb/editor/CMakeLists.txt
@@ -31,6 +31,9 @@ add_library(
   dialogs/aboutdialog.cpp
   dialogs/aboutdialog.h
   dialogs/aboutdialog.ui
+  dialogs/backgroundimagesetupdialog.cpp
+  dialogs/backgroundimagesetupdialog.h
+  dialogs/backgroundimagesetupdialog.ui
   dialogs/circlepropertiesdialog.cpp
   dialogs/circlepropertiesdialog.h
   dialogs/circlepropertiesdialog.ui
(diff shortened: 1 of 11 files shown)


doom-fr:
Awesome !

PR #1457, opened 2024-11-02
Import Specctra DSN export / SES import

ubruhin:
To allow using external routers or autorouters like Freerouting for LibrePCB boards, this PR implements Specctra export/import in the board editor. Some key features:

- Exports board outline, devices, pads, traces, vias, planes & keepout zones to Specctra DSN
- Imports device placement, traces & vias from Specctra SES
- The import tries to detect unmodified objects (fuzzy-compare coordinates & properties of imported objects with existing objects) and keeps their original UUID and other properties to avoid unnecessary file modifications
- During import, some status messages, potential problems or hint messages are displayed to the user
- Extensively tested with [Freerouting](https://freerouting.org/)
- Some initial testing with [Topola](https://topola.dev/) is done, with adjustments in both Topola and this PR, but since there's no official release of Topola yet we can't really guarantee compatibility yet

librepcb-specctra


xotho:
I checked both DSN and SES and seems to be fine.

Further checking in the next dayz.

Take care.
X!

PR #1460, opened 2024-11-07
Support specifying & opening datasheets in editors

ubruhin:
The support for storing datasheet URLs in components and devices has already been implemented in the LibrePCB 1.0 file format (#1181), but it has not been not implemented in the UI yet. This PR implements the missing UI integration, consisting of several parts:

### Library Editor

A new text field for components and devices in the library editor allows to enter a datasheet URL:

image

If a datasheet URL is set, the button next to the text field opens it. Useful for testing URLs but also when working on a library element to quickly access the datasheet.

Note that the file format actually allows to store multiple URLs, not just one. But to keep the UI simple, for now it only displays & modifies the first URL if there are multiple.

### SQLite Library Cache

The background library scan now also stores any URLs on components and devices in the SQLite database so they can be accessed quickly from the schematic- & board editors.

### Schematic- & Board Editors

The context menu of components in the schematic- and board editors now provide quick access to the stored resources:

image

The first item here is the datasheet URL fetched from the SQLite library cache (or from the project's library element, if not found in the cache). When clicked, the datasheet is downloaded & opened.

The second item here is shown only if an MPN is set on the component. Instead of relying on URLs stored in the library, it requests live part information on demand through the [LibrePCB API](https://developers.librepcb.org/d1/dcb/doc_server_api.html#doc_server_api_resources_parts). If the response contains a datasheet URL, it will be downloaded & opened.

### Download & Caching

When requesting to open a datasheet, the following steps are performed:

1. Download the resource from its URL (if not in cache)
2. Store the downloaded file in the user's cache directory (e.g. `~/.cache/LibrePCB`) so the next time, no download is needed (identified by the URL's MD5 hash)
3. Open the cached file with the local PDF reader

If the download fails (e.g. because it's not a direct URL, which is currently the case for our API responses), the URL will be opened in the web browser. This way even indirect URLs or URLs with a cloudflare wall etc. should work.

### DIFF
--- a/libs/librepcb/core/network/networkrequestbase.cpp
+++ b/libs/librepcb/core/network/networkrequestbase.cpp
@@ -109,6 +109,14 @@ void NetworkRequestBase::setMinimumCacheTime(int seconds) noexcept {
   mMinimumCacheTime = seconds;
 }
 
+void NetworkRequestBase::useBrowserUserAgent() noexcept {
+  Q_ASSERT(QThread::currentThread() != NetworkAccessManager::instance());
+  Q_ASSERT(!mStarted);
+  mRequest.setHeader(QNetworkRequest::UserAgentHeader,
+                     "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:131.0) "
+                     "Gecko/20100101 Firefox/131.0");
+}
+
 void NetworkRequestBase::start() noexcept {
   Q_ASSERT(QThread::currentThread() != NetworkAccessManager::instance());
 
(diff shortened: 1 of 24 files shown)


dbrgn:
Very nice 🙂

"Cache into project" would also be nice. I often keep datasheets in my project directory under `docs/` or `datasheets/` to ensure that they are still available if the URL is offline.

In the library manager, a "check for dead datasheet URLs" might also be an interesting extension.

ubruhin:
Good points :+1:  I opened #1461 and #1462.

PR #1477, opened 2024-12-09
Add initial integration of Rust

ubruhin:
## About This Change

As explained in [this blog post](https://librepcb.org/blog/#2024-10-17_roadmap_2.0), our long-term goal is to move from C++ towards Rust (see reasoning there). This PR is the very first step towards this goal. After merging this, the Rust toolchain is required to build LibrePCB.

The strategy is to start with implementing low-level functionality in Rust and providing a thin FFI to be wrapped with C++ (bottom-up approach). The Rust code is compiled to a static library and linked to our C++ application.

## Build System Changes

- Drop support for CMake versions 3.16..3.21, now requiring >=3.22
- Add [Corrosion](https://github.com/corrosion-rs/corrosion) as a submodule (required to integrate Rust crates in CMake)
- Add a Rust crate [`libs/librepcb/rust-core`](https://github.com/LibrePCB/LibrePCB/tree/integrate-rust/libs/librepcb/rust-core) where all our future Rust code will be located
- Use [`cbindgen`](https://github.com/mozilla/cbindgen) to generate FFI header file `ffi.h` (called from `build.rs`)
- Include the Rust crate in CMake and add it as a dependency of `libs/librepcb/core`
- Make `format_code.sh` formatting Rust code with `rustfmt`

## CI Changes

- Update to our new Docker images containing the Rust toolchain:
  - https://github.com/LibrePCB/docker-librepcb-dev/pull/50
  - https://github.com/LibrePCB/docker-librepcb-dev/pull/52
  - https://github.com/LibrePCB/docker-librepcb-dev/pull/48
  - https://github.com/LibrePCB/docker-librepcb-dev/pull/51
- Check formatting of *.rs files (fail on wrong formatting)
- Run [`cargo clippy`](https://doc.rust-lang.org/clippy/) to lint the Rust crate
- Run `cargo test` to run unit tests
- Run `cargo doc` to generate documentation and include it in our Doxygen documentation
- Setup caching of Rust build artifacts

## Rust Library

This PR only intends to set up the whole toolchain so we can start implementing Rust modules afterwards. For verification, a single function `Toolbox::incrementNumberInString()` has been re-implemented in Rust and wrapped in C++ so we can see the integration works properly. In addition, some code is implemented to work with `QString` and `QByteArray` in the FFI.

Documentation preview: https://developers.librepcb.org/_branches/integrate-rust/librepcb_rust_core/index.html

## Architecture

For a better overview of the architecture, I've drawn a completely new diagram:

architecture_overview drawio


ci/azure-pipelines.yml
```diff
@@ -4,6 +4,7 @@ variables:
   # Specify build author as an environment variable because on Apple Silicon
   # build_librepcb.sh is run locally where we want to use a different author.
   LIBREPCB_BUILD_AUTHOR: "LibrePCB CI"
+  CARGO_INCREMENTAL: 0  # Probably required for sccache(?)
```
rnestler, line 7: Since you usually have a clean build anyways in CI incremental compilation has no benefits.

ubruhin: Hmm honestly I'm a bit confused about the incremental build. I read that it creates additional artifacts which are useless on CI *because* every build is a clean build. For example at [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache) they write:

> In particular, the workspace crates themselves are not cached since doing so is generally not effective. For this reason, this action automatically sets CARGO_INCREMENTAL=0 to disable incremental compilation, so that the Rust compiler doesn't waste time creating the additional artifacts required for incremental builds.

So far I'm not sure what is the best way to speed up CI. My local tests with `sccache` were not successful at all, it did not speed up rebuilds. I even experienced CMake to hang randomly which *might* be related to `sccache`. [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache) sounds interesting to cache dependencies, but they write it's not speeding up building the own crate.

The Fluvio project wrote a [blog post about using sccache on CI](https://www.infinyon.com/blog/2021/04/github-actions-best-practices/) but afterwards they [removed sccache from CI](https://github.com/infinyon/fluvio/commit/7c55089253d5819f3e53de1634c6c50ac42e67cd).

After all, it sounds like caching Rust build artifacts is rather difficult. Maybe we should go with [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache) for now (no sccache) as so far dependencies are a much bigger part than our own Rust code.

rnestler: I never really looked into sccache. But for rust incremental compilation: The smallest compilation unit in Rust is a crate (as opposed to C where it is a file). But with incremental compilation the compiler caches with very fine granularity which speeds up small changes a lot. But it need to generate and store all the required metadata and intermediate artifacts, which makes building from a clean state slightly slower. More information on incremental compilation:
 * https://blog.rust-lang.org/2016/09/08/incremental.html
 * https://lwn.net/Articles/997784/
 * https://perf.rust-lang.org/dashboard.html
 * https://rustc-dev-guide.rust-lang.org/queries/incremental-compilation-in-detail.html

.gitmodules
```diff
@@ -34,3 +34,6 @@
 [submodule "libs/dxflib"]
     path = libs/dxflib
     url = https://github.com/LibrePCB/dxflib.git
+[submodule "libs/corrosion"]
+    path = libs/corrosion
+    url = https://github.com/LibrePCB/corrosion.git
```
rnestler, line 39: Why via submodule and not just install it in the Docker containers? I guess because of the old Ubuntu versions?

ci/azure-pipelines.yml
rnestler: In https://github.com/rnestler/reboot-arch-btw/blob/master/.github/workflows/ci.yml I just cache the `target` folder and the registry cache.

libs/librepcb/rust-core/ffi.cpp
```diff
@@ -0,0 +1,81 @@
...
+extern "C" std::size_t ffi_qbytearray_len(const QByteArray* obj) {
+  return static_cast<std::size_t>(obj->size());
+}
+
+extern "C" const uint8_t* ffi_qbytearray_data(const QByteArray* obj) {
+  return reinterpret_cast<const uint8_t*>(obj->data());
+}
+
+extern "C" uint8_t* ffi_qbytearray_data_mut(QByteArray* obj) {
+  return reinterpret_cast<uint8_t*>(obj->data());
```
rnestler, line 44: I would have expected `QByteArray` to use `uint8_t` internally anyways, but no it is using `char` and needs these ugly `reinterpret_cast` which may be invalid on exotic platforms where `unsigned char` isn't equal to `uint8_t`...

libs/librepcb/rust-core/ffi.cpp
```diff
@@ -0,0 +1,81 @@
...
+/*******************************************************************************
+ *  Functions
+ ******************************************************************************/
+
+extern "C" std::size_t ffi_qbytearray_len(const QByteArray* obj) {
+  return static_cast<std::size_t>(obj->size());
+}
+
+extern "C" const uint8_t* ffi_qbytearray_data(const QByteArray* obj) {
+  return reinterpret_cast<const uint8_t*>(obj->data());
```
rnestler, line 40:
```suggestion
  return reinterpret_cast<const uint8_t*>(obj->constData());
```

There is an explicit `constData`, but I'm not sure if it has any benefits.

libs/librepcb/rust-core/src/ffi/cpp_ffi.rs
```diff
@@ -0,0 +1,87 @@
...
+  fn ffi_qstring_set(obj: &mut QString, s: *const c_char, len: usize);
+}
+
+/// Convert `QString` to [String]
+#[allow(dead_code)]
+pub fn from_qstring(obj: &QString) -> String {
+  unsafe {
+    let data = ffi_qstring_utf16(obj);
+    let len = ffi_qstring_len(obj);
+    String::from_utf16_lossy(std::slice::from_raw_parts(data, len))
```
rnestler, line 79: Why not directly use `QString::toUtf8()`? Doesn't it guarantee that it contains only valid UTF-8?

rnestler: Ah then C++ would allocate for the `QByteArray`, this way it's a `String` allocated by Rust.

ci/azure-pipelines.yml
ubruhin: Thanks for the explanation, that halps. And the performance dashboard is nice :)

I just wonder, did you read the code change exactly? It sounds like you argue *against* using incremental builds while the change actually disables incremental builds.

.gitmodules
ubruhin: If you consider any Ubuntu version older than 24.10 as *old*, then yes. Not even the latest LTS 24.04 has corrosion in the package management as far as I can see.

And also simply because this way it's under our control which exact version of corrosion is used, and if needed we have the ability to include our own patches.

ubruhin: Ah and I didn't even talk about Windows. I'd expect having corrosion part of our sources is 100 times easier than installing it system-wide on a Windows...

libs/librepcb/rust-core/ffi.cpp
ubruhin: I was surprised about exactly the same thing when I implemented this cast :sweat_smile: But I assume `char` is always 8-bit (otherwise `QByteArray` makes not much sense) so I don't expect any issues with the cast...

Luckily we can remove this code once 100% of LibrePCB is migrated to Rust :joy:

ubruhin:
> I think it's important to handle String conversion between C++ and Rust in a good way, since it is a thing that can easily make you trip in a lot of cases if not done correctly, so this is a very good place to start. I helped a bit in the [newsboat rust port](https://github.com/newsboat/newsboat/labels/rust-port) and I think string handling is still not fully solved: https://github.com/newsboat/newsboat/issues/1344

Good point, actually I didn't think too much about it yet.

But I think you're right, conversion between Rust strings and `QString` *should* not lead to problems as both explicitly use unicode, while (if I understand correctly) newsboat used `std::string` without ensuring consistent encoding. I just need to be careful with system I/O on the Rust side as I'm used to just rely on Qt which handles all the low-level I/O correctly.

libs/librepcb/rust-core/ffi.cpp
ubruhin: Shouldn't have benefits here. AFAIK it's to avoid detaching the COW type `QByteArray` when called on a non-const object, but our `obj` is `const QByteArray*` so it won't detach anyway.

libs/librepcb/rust-core/src/ffi/cpp_ffi.rs
ubruhin: Oops didn't see your second comment.

> Ah then C++ would allocate for the QByteArray, this way it's a String allocated by Rust.

I'm not sure if I understand correctly. For the C++->Rust conversion I *do* wand to have the string allocated by Rust, as this heavily simplifies ownership (no long-living references to C++-allocated objects).

ubruhin:
Updated the architecture diagram to actually reflect the 2-way dependency of the C++/Rust core libraries:

image

Not ideal this circular dependency (I had to disable compilation of the FFI for `cargo test` because of unresolved linker symbols) but moving these few FFI functions implemented in C++ to a separate static library isn't really worth the effort I guess...

.gitmodules
rnestler: I guess this won't be a distribution problem, since it is only required at build time?

libs/librepcb/rust-core/src/ffi/cpp_ffi.rs
rnestler: > I'm not sure if I understand correctly. For the C++->Rust conversion I do want to have the string allocated by Rust, as this heavily simplifies ownership (no long-living references to C++-allocated objects). Or what would you propose?

Yeah exactly. I thought we could maybe avoid allocations altogether in some cases by just having an UTF-8 view into the `QString`. But since it uses UTF-16 internally, we need to allocate in all cases were we want to access it as UTF-8.

It's probably easiest and safest to just convert between UTF-16 and UTF-8 in the FFI.

rnestler:
> I just need to be careful with system I/O on the Rust side as I'm used to just rely on Qt which handles all the low-level I/O correctly.

Rust handles system I/O correctly as well. That's the reason there are things like [OsString](https://doc.rust-lang.org/std/ffi/struct.OsString.html)

> but moving these few FFI functions implemented in C++ to a separate static library isn't really worth the effort I guess...

It's probably cleaner to have it separate, no?

ci/azure-pipelines.yml
rnestler: > It sounds like you argue against using incremental builds while the change actually disables incremental builds.

Yes. My initial comment should have an "anyway" in the end. So not only because of sscache you'd want to disable it but also because it's usually not worth to do in CI.

.gitmodules
ubruhin: Probably yes, but it should also be easy to *compile* LibrePCB on Windows. I do need this by myself when debugging Windows problems. I guess there isn't any real downside of having corrosion added as a submodule (the clone time for this repo shouldn't be an issue) but it saves a lot of troubles on systems where corrosion is not easily available.

ci/azure-pipelines.yml
ubruhin: :+1: Improved the comment accordingly.

ubruhin:
Updated CI config to cache `~/.cargo/{bin,registry/index,registry/cache}` and the `target` directories instead of using sccache. And now on all CI jobs, not just Windows.

> Rust handles system I/O correctly as well. That's the reason there are things like [OsString](https://doc.rust-lang.org/std/ffi/struct.OsString.html)

Perfect :+1:

> It's probably cleaner to have it separate, no?

Probably yes, but still not really nice because this separate library (defined in CMake) would be a dependency of the Rust library which isn't built when compiling with `cargo build`. Since this approach doesn't provide advantages, I prefer to keep it simple with a circular dependency which in practice works fine...

PR #1487, opened 2025-01-22
DRC: Report vias as useless if connected on <2 layers

connorslade:
The previous implementation only detected vias as useless if they had no `drillLayerSpan`. Now they are also detected if there are traces from less than two layers connected. This identifies situations like below, where you may mistakenly have left a via:

<details><summary>Screenshot</summary>
<p>

Screenshot from 2025-01-21 at 22_50_14 807626492

</p>
</details>

### DIFF
--- a/libs/librepcb/core/project/board/drc/boarddesignrulecheck.cpp
+++ b/libs/librepcb/core/project/board/drc/boarddesignrulecheck.cpp
@@ -22,6 +22,7 @@
  ******************************************************************************/
 #include "boarddesignrulecheck.h"
 
+#include "../../../geometry/padgeometry.h"
 #include "../../../geometry/via.h"
 #include "../../../types/layer.h"
 #include "../../../utils/clipperhelpers.h"
@@ -29,6 +30,7 @@
 #include "../boardplanefragmentsbuilder.h"
 #include "boardclipperpathgenerator.h"
 #include "boarddesignrulecheckmessages.h"
+#include "polyclipping/clipper.hpp"
 
 #include <QtConcurrent>
 #include <QtCore>
@@ -245,6 +247,7 @@ BoardDesignRuleCheck::Result BoardDesignRuleCheck::run(
         },
         3);
   }
+
   addToStage2(&BoardDesignRuleCheck::checkCopperHoleClearances, 3);
   if (!data->quick) {
     addToStage2(&BoardDesignRuleCheck::checkMinimumPthAnnularRing, 2);
@@ -259,10 +262,10 @@ BoardDesignRuleCheck::Result BoardDesignRuleCheck::run(
     addIndependent(&BoardDesignRuleCheck::checkInvalidPadConnections, 2);
     addIndependent(&BoardDesignRuleCheck::checkDeviceClearances, 2);
     addIndependent(&BoardDesignRuleCheck::checkBoardOutline, 1);
+    addIndependent(&BoardDesignRuleCheck::checkVias, 1);
   }
   addSequential(&BoardDesignRuleCheck::checkMinimumCopperWidth);
   if (!data->quick) {
-    addSequential(&BoardDesignRuleCheck::checkVias);
     addSequential(&BoardDesignRuleCheck::checkAllowedNpthSlots);
     addSequential(&BoardDesignRuleCheck::checkAllowedPthSlots);
     addSequential(&BoardDesignRuleCheck::checkUsedLayers);
@@ -1637,18 +1640,32 @@ RuleCheckMessageList BoardDesignRuleCheck::checkZones(const Data& data) {
 RuleCheckMessageList BoardDesignRuleCheck::checkVias(const Data& data) {
   RuleCheckMessageList messages;
   emitStatus(tr("Check for useless or disallowed vias..."));
+
   for (const Data::Segment& ns : data.segments) {
     for (const Data::Via& via : ns.vias) {
-      if (!via.drillLayerSpan) {
-        messages.append(
-            std::make_shared<DrcMsgUselessVia>(ns, via, getViaLocation(via)));
-      } else if ((via.isBlind && (!data.settings.getBlindViasAllowed())) ||
-                 (via.isBuried && (!data.settings.getBuriedViasAllowed()))) {
+      if ((via.isBlind && (!data.settings.getBlindViasAllowed())) ||
+          (via.isBuried && (!data.settings.getBuriedViasAllowed()))) {
         messages.append(
             std::make_shared<DrcMsgForbiddenVia>(ns, via, getViaLocation(via)));
       }
+
+      // If the via has no drill layer span, emit an InvalidVia warning
+      if (!via.drillLayerSpan) {
+        messages.append(
+            std::make_shared<DrcMsgInvalidVia>(ns, via, getViaLocation(via)));
+        continue;
+      }
+
+      // If the total number of layers connected to the via is less than two,
+      // add a 'useless via' warning. Connections can be made by traces, planes,
+      // or pads.
+      if (isViaUseless(data, ns, via)) {
+        messages.append(
+            std::make_shared<DrcMsgUselessVia>(ns, via, getViaLocation(via)));
+      }
     }
   }
+
   return messages;
 }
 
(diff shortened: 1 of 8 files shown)


ubruhin:
Good point, I agree this case should emit a DRC warning. However, I think the warning should be emitted only when the via is connected on *exactly* one layer. Vias not connected on *any* layer are a common thing (e.g. vias used as plated holes for mechanical purposes) and should therefore not emit the warning.

And there's another important case to consider: Vias connected to a trace on one layer, and to a plane on another layer (or even just connected purely to planes, like stitching vias) are totally valid and should not emit the warning too. Currently, connections to planes are not counted, which would lead to the warning. I think for this check we have to consider connections to planes just like connections to traces, but unfortunately this is not trivial at all :-/

Another thing is that probably we should have separate message types for your case and for the current `DrcMsgUselessVia` case. The current case is really something invalid (a via with the drill only on one layer). Probably `DrcMsgUselessVia` was a bad name (and it's message is misleading) for the current case, it fits way better for your new check.

Two comments regarding implementation:
- Please always use `{}` blocks for any `if` and `for` :nerd_face: 
- Detecting the connected traces by comparing their coordinates with the via coordinate is not ideal to be honest... I think it would be worth adding a new member `QSet<const Layer*> connectedLayers;` to the `Via` struct and calculate it during the data preparation step.

connorslade:
Thank you for your feedback; I tried to address all of your suggestions.

As recommended, I modified the `BoardDesignRuleCheckData` constructor to identify the layers that each via is connected to using their registered net lines. I also added support for vias connected to planes based on how vias are handled in the `BoardPlaneFragmentsBuilder` class.

For the separate error types, I created `DrcMsgInvalidVia` for vias with no drill layer spans and use `DrcMsgUselessVia` for the detection system added in this PR.

Please let me know if there are any other changes I should make.

Also, currently CI is failing for this PR as the DRC warning changes break some of the board design rule check tests but I'm not sure now to update the test data as it is stored in a seprrate repo ([LibrePCB/librepcb-test-data](https://github.com/LibrePCB/librepcb-test-data)).

ubruhin:
Ok nice job, it's going into the right direction :slightly_smiling_face: 

One problem now is that the clipper stuff is an expensive operation and thus should not be done in the `BoardDesignRuleCheckData` constructor as it is executed in the GUI thread. Instead, `checkVias()` should be changed to an "independent job" (see `addIndependent()` and the plane connections calculated right in `checkVias()` (I think all the required data should be available at this point). Preparing `connectedLayers` is fine in the constructor as it is lightweight.

Btw, I wonder if it would make sense to simplify the plane connection detection to only check if the via's center coordinate is within a plane :thinking: It might be good enough (and faster, see `ClipperLib::PointInPolygon()`) and more reliable as "barely touching" planes would clearly emit the DRC warning. No strong opinion about that at the moment...

And just another case came into my mind: Thermal vias within SMD pads are another common use-case which on one layer neither have traces not planes, but still connect to a net. So I think the connection detection with clipper should also include pads in addition to planes.

And a small detail: If I see correctly, vias which emit `DrcMsgInvalidVia` will most probably (or even for sure?) also emit `DrcMsgUselessVia`. I think in that case it would be better to just emit `DrcMsgInvalidVia` (which is more critical) and ignore `DrcMsgUselessVia` as it is kind of redundant.

> Also, currently CI is failing for this PR as the DRC warning changes break some of the board design rule check tests but I'm not sure now to update the test data as it is stored in a seprrate repo ([LibrePCB/librepcb-test-data](https://github.com/LibrePCB/librepcb-test-data)).

Hmm yeah unfortunately for PRs from forks it's a bit cumbersome to change test data :( I could update the test project to also include tests for the new DRC warning and then we can point to the new commit of the submodule.

One thing we have to consider too is the file format change. I totally agree with the rename from `useless_via` to `invalid_via`, but it will have an impact on existing projects. The easiest would be to ship this PR with the next major version v2.x and do the rename in project files during file format upgrade. But if there will be a new v1.x release first (which is not clear yet), I'm not sure how we could smoothly include this PR. But let's ignore that for the moment...

ubruhin:
Ah, I have an idea: Within `DrcMsgInvalidVia`, let's change the key to `useless_via` for now. This will keep the message compatible with the current file format (and the current test data). And in `DrcMsgUselessVia`, change the key to `antennae_via` or something similar. We can do the rename (including file format upgrade) later in LibrePCB 2.x, no matter if a v1.x is published first or not.

connorslade:
Alright, I implemented your suggested changes:

- The clipper checks have been moved from the `BoardDesignRuleCheckData` constructor into functions (`getViaConnectedLayers` and `isViaConnectedToAnyPad`) called by `checkVias`, which is now an independent job;
- vias that are connected to device pads are ignored from the 'useless via' check;
- the changed via warning/error message keys have been reverted to remain backwards compatible; and
- the 'useless via' and 'invalid via' messages won't both be triggered at the same time, with the latter taking priority.


I also switched to just looking at the center of each via when checking if it's in each plane / pad, just because it didn't seem worth the extra computation if an error would be raised from the via being on the edge of a plane anyway. But I am just now realizing that no error would be raised for vias just barely on device pads. I don't know how bad it would be for the performance of a big project to multiply the number of `PointInPolygon` checks (currently every via is compared with every plane and every device pad) by the number of points making a via. So ill just leave it as is for now.

ubruhin:
Awesome!

The conditional to emit `DrcMsgUselessVia` looks unexpected though. Maybe we had a misunderstanding. I think it does not matter if a via is connected to a trace, a plane or a pad. Every connection is equal. And if a via is connected on less than two layers (no matter what kind of connections), the warning should be emitted. For example, a via connected only to a pad on one layer is still an antennae via which does not act as a thermal via as the heat is not conducted to another layer.

Another thing is that it is currently really bad in performance if I understand correctly. DRC checks really need to be optimized for performance, at least all the low-hanging fruits shall be implemented. And in this case, it is very simple to make it efficient: Just start counting the number of connected layers from top to bottom layer. As soon as 2 layers are found, abort the detection. And on each layer, abort the detection if one connection is found.

Since `connectedLayers` is already pre-filled with any connected traces (this doesn't need to be optimized as it is lightweight), we even don't have to do *any* clipper operation if this set contains >=2 entries. This will make the DRC check basically a no-op for 95% of the cases so we can save a lot of processing time.

I'm sorry that it takes many iterations to get this implemented... but I really care about every feature is being done right, and your implementation really goes into the right direction so I'm happy to assist :slightly_smiling_face:

connorslade:
I modified the `checkVia` method to count vias on pads as just connections to the layer the pad is on. Also, it will immediately move on to checking the next via once two connected layers have been identified.

libs/librepcb/core/project/board/drc/boarddesignrulecheckdata.h
```diff
@@ -63,13 +64,16 @@ struct BoardDesignRuleCheckData final {
...
+    // Only filled with the layers of traces directly connected to the via.
+    QSet<const Layer*> connectedLayers;
     const Layer* startLayer;
     const Layer* endLayer;
     std::optional<std::pair<const Layer*, const Layer*>> drillLayerSpan;
     bool isBuried;
     bool isBlind;
     std::optional<PositiveLength> stopMaskDiameterTop;
     std::optional<PositiveLength> stopMaskDiameterBot;
+    NetSignal* netSignal;  // May be null.
```
ubruhin, line 76: Unfortunately this pointer should not be added to the struct as it is not safe to be accessed from threads. I know currently it is never dereferenced and thus won't cause problems, but honestly it feels dangerous to have it here (easy to access it in future while forgetting its non-threadsafety). Fortunately we don't need it anyway - just use `Segment::net` instead which is already there and fully threadsafe.

libs/librepcb/core/project/board/drc/boarddesignrulecheckdata.h
```diff
@@ -100,6 +104,7 @@ struct BoardDesignRuleCheckData final {
     UnsignedLength minWidth;
     Path outline;
     QVector<Path> fragments;
+    NetSignal* netSignal;  // May be null.
```
ubruhin, line 107: Same here - the existing member `net` should be used instead.

And don't forget to also remove the `#include` for `netsignal.h` :nerd_face:

libs/librepcb/core/project/board/drc/boarddesignrulecheck.cpp
```diff
@@ -2284,6 +2301,76 @@ QVector<Path> BoardDesignRuleCheck::getViaLocation(
...
+      if (connectedLayers.count() >= 2) {
+        return false;
+      }
+    }
+  }
+
+  // For each pad geometry that shares a layer with the current via, check if
+  // the center of the via is in the outline of the pad.
+  for (const Data::Device& device : data.devices) {
+    for (const Data::Pad& pad : device.pads.values()) {
```
ubruhin, line 2342: Is the `.values()` added by intention? It will deep-copy the list but I guess there's no need to make a deep copy for this read-only access.

```suggestion
    for (const Data::Pad& pad : device.pads) {
```

libs/librepcb/core/project/board/drc/boarddesignrulecheck.cpp
```diff
@@ -2284,6 +2301,76 @@ QVector<Path> BoardDesignRuleCheck::getViaLocation(
...
+
+  // A via is considered to be connected to a plane if it contains
+  // the layer the plane is on, both of them are part of a net, they are in
+  // the same net, and the via is physically within the plane
+  for (const Data::Plane& plane : data.planes) {
+    const int copperNumber = plane.layer->getCopperNumber();
+    if (via.startLayer->getCopperNumber() > copperNumber ||
+        via.endLayer->getCopperNumber() < copperNumber || !plane.netSignal ||
+        !via.netSignal || plane.netSignal != via.netSignal ||
+        connectedLayers.contains(plane.layer)) {
```
ubruhin, line 2325: If every condition contained parenthesis, I'd sleep better :see_no_evil: I've seen too many nasty bugs due to unexpected operator precedence so I really prefer to make it explicit (IMHO it's even more readable). Just a note, I'll accept it as-is.

(same applies to the condition below)

libs/librepcb/core/project/board/drc/boarddesignrulecheckmessages.h
```diff
@@ -920,6 +920,28 @@ class DrcMsgForbiddenVia final : public RuleCheckMessage {
...
 
+/*******************************************************************************
+ *  Class DrcMsgInvalidVia
+ ******************************************************************************/
+
+/**
+ * @brief The DrcMsgInvalidVia class
+ */
+class DrcMsgInvalidVia final : public RuleCheckMessage {
+  Q_DECLARE_TR_FUNCTIONS(DrcMsgForbiddenVia)
```
ubruhin, line 931:
```suggestion
  Q_DECLARE_TR_FUNCTIONS(DrcMsgInvalidVia)
```

ubruhin (commented):
FYI there are several line breaks in the diff on otherwise unchanged code. Not critical (no need to revert) but generally it would be nice to avoid such unintended changes...

I'll take care of updating the test data now (and also test this new feature).

ubruhin:
Ok during testing I've found a bug: A via in a pad is currently not reported as unused because the pad's stop mask & solder paste layers are also counted as connected layers. Where you access `it.key()->getCopperNumber()`, you also need to check `it.key()->isCopper()` and discard any non-copper layers.

libs/librepcb/core/project/board/drc/boarddesignrulecheck.cpp
connorslade: I typically work in Rust where looping through a map yields items of type (key, value), so I added that without realizing it was unnecessary. Oops.

connorslade:
Thank you for taking the time to help me through this process. I fixed all the review comments you made and the stop mask / solder paste bug.

Should I update the test data submodule to point to your new commits?

ubruhin:
I pushed the unit test changes now, hope CI will now succeed. I'll take a final look tomorrow (today probably I don't have time).

libs/librepcb/core/project/board/drc/boarddesignrulecheck.cpp
```diff
@@ -2284,6 +2301,78 @@ QVector<Path> BoardDesignRuleCheck::getViaLocation(
...
+
+  // For each pad geometry that shares a layer with the current via, check if
+  // the center of the via is in the outline of the pad.
+  for (const Data::Device& device : data.devices) {
+    for (const Data::Pad& pad : device.pads) {
+      const Transform transform(pad.position, pad.rotation, pad.mirror);
+
+      for (auto it = pad.geometries.begin(); it != pad.geometries.end(); it++) {
+        const int copperNumber = it.key()->getCopperNumber();
+        if (it.key()->isCopper() ||
```
ubruhin, line 2348: The logic is the wrong way around :see_no_evil: 

```suggestion
        if ((!it.key()->isCopper()) ||
```

libs/librepcb/core/project/board/drc/boarddesignrulecheckdata.cpp
```diff
@@ -86,14 +87,25 @@ BoardDesignRuleCheckData::BoardDesignRuleCheckData(
...
+
+      // Add all the layers of traces directly connected to the current via to
+      // the connectedLayers set. The via may be connected to more layers
+      // through other mechanisms like planes, but identifying those connections
+      // can be expensive, so it's not done here.
+      foreach (const BI_NetLine* nl, bi_via->getNetLines()) {
+        connectedLayers.insert(&nl->getLayer());
+      }
+
+      const librepcb::Via via = bi_via->getVia();
```
ubruhin, line 101: I wanted to ignore that for now but as another change needs to be made anyway, this could be improved a bit too (avoid unnecessary copy):

```suggestion
      const librepcb::Via& via = bi_via->getVia();
```

Btw actually we never use snake_case variable names... I this case I simply wouldn't create the `via` variable to avoid the need for renaming `bi_via` but it's clearly a matter of taste so I'll accept your way :slightly_smiling_face: No need to change anything here, only the `&` would be nice.

ubruhin (commented):
Almost perfect! Just one wrong conditional prevented me from merging :sweat_smile: 

Ci is also good (it seems the wrong conditional didn't make any test fail, but that's acceptable as it is just an unlucky edge-case that now the stop mask layer counts as a copper connection).

connorslade:
Fixed!

ubruhin:
Thanks! I just squashed all your commits as-is to keep a clean history (of course retaining your author information). Will merge once CI succeeds.

Thanks also for following our coding style very well, much appreciated :pray:

PR #1505, opened 2025-03-29
Integrate AboutDialog into Slint main window

ubruhin:
One dialog eliminated :muscle:

Integrated into the new main window in a responsive way - the smaller the window, the less details shown:

image


dbrgn:
Is the content shown dependent on the window size, without any scrollbar? If yes, I don't think that's a good application of responsive design. Reactions to screen size should affect presentation and layout, not content 🙂

But maybe that's a misunderstanding.

ubruhin:
Yes, content is hidden on small windows, but I did it carefully:

- In very general, it's unlikely people are using LibrePCB on screens with less then 1000 pixels vertically, so in the real use-case no content is hidden
- Logos are no relevant information, they can be hidden if there's no space for them
- All the buttons are there just for convenience. Most of them are also in the top menu so they are not lost. The only one really hidden is the link to the GitHub repository, but I don't consider this as an important button.
- The prosa texts aren't really important either, at least not having them in the application. The credits section is also on the website where it reaches even more users than in the application.
- Hiding the contribution to Icons8 is not nice, but I'll get rid of the Icons8 images before releasing LibrePCB 2.0 so we don't need this contribution anymore at all ;)
- The only really really important information is the version information and is therefore never hidden.

PR #1582, opened 2025-08-17
Schematic editor: Show location of selected ERC message

### LINKED ISSUE #553: Schematic Editor : Add 'goto' function that center the problem when double-click on schematic error
##### SUMMARY
image

When you have errors, it is difficult to know where is located the problem in the different pages of schematics (here, the problem is caused by a problem with forced nets and occurs not just after having placed the item, but later, and in many places (the problem seems to be spread)).

##### SOLUTION
a 'goto' function that switch to the good page, center the element and if possible highlight it.
(but the first two will already be a good improvement).

If not possible or too hard, just tell the page and coordinates.

ubruhin:
Like in the board editor, show the location of ERC problems in the schematics. It's not perfect since it's much more complicated than DRC issues (ERC issues sometimes have no visual appearance at all, sometimes even multiple locations spread across schematics). For now, only the first visual location is shown (if there is any) which is already very helpful.

Closes #553


xotho:
absolutely useful !!!!!

PR #1591, opened 2025-08-29
Board editor: Automatically simplify net segments

### LINKED ISSUE #710: Board Editor : Remove unnecessary nodes of wires when extending them

ubruhin:
When drawing traces or deleting traces/vias in the board editor, the modified net segments will be automatically simplified in the following ways:

- Remove redundant junctions (same position, same layer)
- Remove redundant traces (same anchors, same layer), keeping only the thickest of them
- Remove zero-length traces
- Remove useless junctions within straight trace segments (join line segments into the same direction to a single trace)
- Split traces to connect with junctions on the path between start and end points
- Split intersecting traces, placing a new junction to connect them (only orthogonal intersections for now)

The simplification is added as a separate operation to the undo stack. If for any reason the user doesn't want the simplified result, just press Ctrl+Z and you'll have your modifications applied without the simplification.

The algorithm is implemented in a generic way so we can reuse it later for the schematic editor.

librepcb-simplify-board

Replaces #737
Closes #710

### DIFF
--- a/libs/librepcb/core/CMakeLists.txt
+++ b/libs/librepcb/core/CMakeLists.txt
@@ -75,6 +75,8 @@ add_library(
   3d/stepexport.h
   algorithm/airwiresbuilder.cpp
   algorithm/airwiresbuilder.h
+  algorithm/netsegmentsimplifier.cpp
+  algorithm/netsegmentsimplifier.h
   application.cpp
   application.h
   attribute/attribute.cpp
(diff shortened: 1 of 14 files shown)


dbrgn:
Very cool :)

xotho:
yessss! great!

PR #1595, opened 2025-09-05
Add note to README.md

web-coder-of-rpi:
The note explains that if the build is stopped, it can run again from where it left off.
This image shows that if the build folder is already there, it will operate quickly to about 38 percent before it starts to take longer.
make-j8


web-coder-of-rpi:
Ignore the first 13 commits

ubruhin:
Sorry but this is absolute basic `make` knowledge and is not specific to LibrePCB, so it makes no sense to document it again in every project that uses `make`.

web-coder-of-rpi:
Thanks for telling me, I'm no expert. I wanted everyone to know that you can stop it as when I did it for the first time my computer froze at 66%.

PR #1596, opened 2025-09-05
Support adding individual pads directly to boards

### LINKED ISSUE #1078: Add support for plated slots in boards
##### SUMMARY

Non-plated slots were implemented in #1071. Plated footprint pad slots were implemented in #1076. But pads can only be added to footprints (i.e. in the library editor), not directly to a board. However, I'm pretty sure sooner or later the need for plated slots in boards will arise as well, which is currently still not possible (except by wrapping it in a package, which is cumbersome).

Well, actually it would be possible to create plated slots by drawing polygons on the "Milling" layer. But since the polygon outline can be arbitrary, it's handled differently in the Gerber export (not part of drill files, separate milling file needs to be explained manually to the manufacturer, and a larger tool diameter might be required). It would be much better to create true slots which get exported to drill files.

##### SOLUTION

I'm really not sure how this should be implemented. Maybe extending via objects to allow creating slots? However, once blind/buried vias will be supported as well, I don't know if it's a good idea to support slots at the same time (i.e. blind/buried plated slots :sob:). In addition, vias currently can only be square-sized (round/square/octagon) which wouldn't work for slots.

Adding a "plated" property to hole objects is also strange, since holes are non-electrical objects but plated slots usually are connected to some net.

And allowing to add footprint pad objects to boards directly is also somewhat strange (would be mostly redundant to vias and thus confusing, while adding much more complexity).

Any ideas welcome ;-)

ubruhin:
The initial goal of this was to support adding plated slots to boards as requested in #1078. The solution I have chosen is to allow adding arbitrary, individual footprint pads directly to boards. Footprint pads already support configuring them as slots, so this feature implicitly allows to create slots in boards.

Unfortunately this solution is a bit complex (affecting file format, UI, FSM, DRC, Gerber export, PnP export, Netlist export, Specctra export/import). But I think it is worth the complexity because it allows not only to add plated slots, but also any other kind of pads. For example, it is quite handy to add test points, fiducials, solder pads etc. to a board without requiring to create footprints for them.

I didn't add dedicated toolbuttons in the board editor to not promote this way of adding pads too much (whenever a footprint is available for test points, fiducials etc., those should be preferred over board pads). The new features are put into a context menu of the "add via" toolbutton.

librepcb-board-pads-2

Closes #1078


dbrgn:
Very nice! 🚀

PR #1631, opened 2025-11-06
Support sharing DRC settings & output jobs via libraries

ubruhin:
## Goal

The idea of this feature is to allow sharing PCB design rules (respectively DRC settings) and fabrication output settings (via output jobs) through libraries. It makes selecting the correct settings for each PCB manufacturer much easier, faster and less error-prone.

## More DRC Settings

To make this feature more powerful, DRC settings were extended with new attributes to also store min/max PCB sizes, max. number of layers, available thicknesses, solder resist colors & silkscreen colors - i.e. the capabilities of a PCB manufacturer. In future this allows us to implement DRC warnings if the user configured a board setup which the PCB manufacturer is not able to manufacture.

## New Library Element Type

As the core of this feature, a new library element type named "Organization" is introduced. This generic name has been chosen to allow using it for more things in future, e.g. sharing BOM settings for part distributors.

An Organization can have 0..n sets of DRC settings, 0..n PCB fabrication output jobs, and 0..n PCB assembly output jobs. It can be created and edited in the library editor (the actual settings are in popup windows):

510842763-037f69c8-83a6-427d-a56e-3d0e954222e3

## Initialize new Boards With DRC Settings

For new (empty) boards in projects, the board editor suggests to import design rules from one of the organizations. A click on any of the listed presets loads them and dismisses the message. This encourages users to select the design rules *before* starting with the layout.

image

## Import DRC Settings

For already existing boards, this can also be done at any time later:

image

The board stores a reference to the loaded preset, therefore it will be possible in future to notify the user when the referenced presets were updated with new settings.

## Import Output Jobs

In the output jobs dialog, new jobs can now be added by importing one of the presets of the organizations:

image

## Notes

Since this feature is implemented in a generic way, users can also create their own organizations for the PCB manufacturers / assembly houses they use for their projects. But the goal is to add many of them to our official libraries to make them available for all LibrePCB users.

Note that when loading any organization preset, the organizations themselves are *not* copied into the project, they exist only in the workspace library. During the import, only the corresponding settings are copied into the board (DRC settings) resp. into the project (output jobs).

Due to the whole bunch of new features, this is currently considered as _experimental_. We will need to gather some real-world experience to see what works well and where some adjustments are needed. Possibly there are still some rough edges.


dbrgn:
Very nice!

Some bikeshedding: I wonder if Corporate is a good name. I mostly know the term as an adjective or as a compound noun (e.g. "corporate body", "corporate design").

Maybe Organization? Supplier? Company? Firm? LegalEntity?

Or more specific: Fab? (Would not work for distributors though.)

ubruhin:
Yes I'm still not happy with it either. You made some good proposals. Somehow I like Organization and Legal Entity but I'm not sure. Fab and Supplier sound a bit too specific IMO, though Supplier may still be good enough for any realistic use-case. The good about Supplier is that it is quite intuitive.

Another option which just came into my mind: Business

So hard to decide, strong opinions very welcome :sweat_smile:

dbrgn:
I think @rnestler also likes bikeshedding, so: Ping 😂

rnestler:
> Some bikeshedding: I wonder if Corporate is a good name. I mostly know the term as an adjective or as a compound noun (e.g. "corporate body", "corporate design").

The noun would be corporation, right?

Maybe Manufacturer?

PetoJanis:
I think a more neutral term could be "**Technology**" or something similar. 
The settings can include both manufacturing rules (for the PCB itself) and assembly rules (for the physical assembly of the PCB), which can be applied by companies and individuals alike. 
Furthermore, this approach leaves room for future expansion—such as adding rules for AOI or testing—and avoids being tied to corporate terminology. 
That’s why I believe it’s better to use "**Technology**" as the overarching term.

ubruhin:
Thanks for the inputs. I think _Manufacturer_ is too specific (wouldn't work for part distributors) while _Technology_ is too generic (not intuitive for users to understand what this library element is intended for).

At the moment I am considering Organization and Supplier as good candidates as they are both intuitive and generic enough to be usable for part distributors and assembly services.

ubruhin:
Renamed Corporate to Organization for now.

I was also thinking about a more generic term like "Preset" or "Configuration Set" but at the moment I'm not sure if that makes sense. In worst case, I think we can still rename the library element in a future release if we really need it to be more generic than Organization (though the change of the file format might be a bit painful).

PR #1633, opened 2025-11-07
Migrate from polyclipping (Clipper1) library to Clipper2

### LINKED ISSUE #1481: Update polyclipping library
##### SUMMARY

This issue is mainly a reminder to myself to update the polyclipping library before releasing LibrePCB 2.0, as this will be a good time to do it. Clipper2 seems to contain a lot of improvements: https://www.angusj.com/clipper2/Docs/Changes.htm

We might also consider using a Rust alternative, but I'm not sure if any exists. There are some wrappers for Clipper, but nothing mature as far as I can see.

- [ ] Evaluate if there is a good Rust alternative or wrapper
- [ ] If going with a Rust crate, discuss with me how to proceed
- [ ] If going with Clipper2 (C++), update `./libs/polyclipping/` to the latest version
    - To be checked: Do we only need the files from the `CPP/Clipper2Lib` directory? Or is it worth to include the whole Clipper2 repository as a submodule?
- [ ] Update usages of polyclipping in our code, update tests if needed, do some manual tests

ubruhin:
Work in progress...

Closes #1481

### DIFF
--- a/libs/librepcb/core/project/board/boardplanefragmentsbuilder.cpp
+++ b/libs/librepcb/core/project/board/boardplanefragmentsbuilder.cpp
@@ -359,7 +359,7 @@ BoardPlaneFragmentsBuilder::Result BoardPlaneFragmentsBuilder::run(
     ClipperHelpers::subtract(
         *data->boardArea,
         ClipperHelpers::convert(boardCutouts, maxArcTolerance()),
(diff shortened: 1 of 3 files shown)


ubruhin:
Hmm the performance difference to the old lib seems to be minimal (almost not measurable) at first glance (no representative measurement, just running the whole DRC). And the new API seems to miss some features we use currently, which makes the migration cumbersome and error-prone. Not sure if it is worth the effort & risk at this time :thinking:

PR #1697, opened 2026-02-20
Doxygen: Change ID naming convention from snake-case to kebab-case

megabyte6:
As far as I am aware, this isn't snake case. Snake case would look more like `snake_case`. This is [`kebab-case`](https://developer.mozilla.org/en-US/docs/Glossary/Kebab_case).


ubruhin:
Ah indeed, thanks!

PR #1709, opened 2026-03-06
Board editor: Support flipping view to bottom side

### LINKED ISSUE #1398: Board editor: View PCB from bottom side
Conceptual consideration:

In the Board Editor we have the possibility to 'see' the board from TOP layer or a 'view through' of BOTTOM layer.....

Why not implement a 180 degree board turn as a PUSH BUTTON function ?

In this way your working layer can let you work not in a 'view through' view...

There are more related questions to be seen obviously maybe about the coord system reference about one or more boards on the same BOARD EDITOR workspace....

Just an idea...

take care.

ubruhin:
Support flipping the board view, so the board can be viewed and edited from its bottom side. For now, only horizontal flipping is supported.

image

Closes #1398

### DIFF
--- a/img/images.qrc
+++ b/img/images.qrc
@@ -4,6 +4,7 @@
         <file alias="/bi/align-bottom.svg">../libs/bootstrap-icons/icons/align-bottom.svg</file>
         <file alias="/bi/align-center.svg">../libs/bootstrap-icons/icons/align-center.svg</file>
         <file alias="/bi/align-top.svg">../libs/bootstrap-icons/icons/align-top.svg</file>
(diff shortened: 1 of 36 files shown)


vicencb:
Hi @ubruhin, i am using this feature and it is very useful, but i would like the flipped state to be sticky.

As it is now, if a tab has its view flipped, focus is given to another tab and then focusing back the board tab, then the view is reset to normal.

ubruhin:
Hi @vicencb,
Yess I also observed this behavior and fixed it in 20a6f1fca693db4c30397a4d45ba5cf9af7f5862, it will be included in the next release.

Issue #1254, opened 2023-10-09
Support loading semi-transparent STEP models


ubruhin:
##### SUMMARY

I'm not sure if STEP supports transparency (RGBA) but at least with CadQuery it's possible to pass an alpha value to the `Color` class. This would actually be nice for semi-transparent 3D models like LEDs since it would look more realistic.

##### SOLUTION

- [ ] Check if and how STEP and OpenCascade support RGBA.
- [ ] If it is supported, make `OccModel::tesselate()` loading the alpha values and make sure the result in the 3D viewer looks as expected.
- [ ] Adjust [librepcb-parts-generator](https://github.com/LibrePCB/librepcb-parts-generator) to generate LEDs with semi-transparent body

PR #1755, opened 2026-04-12
Support loading semi-transparent STEP models

Girish4489:
This PR implements support for loading and rendering semi-transparent (RGBA) STEP models in the 3D viewer, enabling more realistic rendering of components such as LEDs.

## Key Changes:
   * **Infrastructure**: Updated `OccModel::Color`, `StepExport::Color`, and `OpenGlSceneBuilder::Color` from RGB (3-element tuple) to RGBA (4-element tuple).
   * **STEP Loading**: Modified `OccModel::loadStep` and its helpers to extract alpha values using OpenCascade's `Quantity_ColorRGBA` (supported in OCCT 7.5.0+). Added appropriate fallback logic for older OCCT versions (defaulting to alpha=1.0).
   * **PCB Transparency**: Updated `OccModel::createBoard` to support the alpha channel, allowing the PCB body to be rendered with transparency if specified.
   * **Rendering**: Adjusted `OpenGlSceneBuilder::publishDevice` to pass the extracted alpha channel to the OpenGL renderer via `QColor`.
   * **Verification**: Added a new unit test `OccModelTest.testTransparency` which programmatically creates a semi-transparent model and verifies that the alpha value is correctly preserved during the tessellation process.

 ## Visual Proof:
  
output

 **Closes**: #1254


Girish4489:
I checked the [librepcb-parts-generator](https://github.com/LibrePCB/librepcb-parts-generator) and found that the LED generation script ([generate_led.py](https://github.com/LibrePCB/librepcb-parts-generator/blob/master/generate_led.py)) was actually already configured with an alpha value of 0.5 in its color tuples.

tests/unittests/core/3d/occmodeltest.cpp
```diff
@@ -207,6 +207,31 @@ TEST_F(OccModelTest, testTesselate) {
...
+    const QMap<OccModel::Color, QVector<QVector3D>> result = model->tesselate();
+    EXPECT_EQ(result.count(), 1);
+
+    const OccModel::Color resultColor = result.keys().first();
+    EXPECT_NEAR(std::get<0>(resultColor), 1.0, 0.01);  // Red
+    EXPECT_NEAR(std::get<1>(resultColor), 0.0, 0.01);  // Green
+    EXPECT_NEAR(std::get<2>(resultColor), 0.0, 0.01);  // Blue
+    // Since we know the current build environment has OCCT 7.9.2,
+    // we expect alpha to be correctly loaded as 0.5.
+    // If it were an older version, it would be 1.0.
```
ubruhin: Here I'd still prefer to make the code compatible with older OCC versions (i.e. adding #if conditional) since the tests may also be run locally on developers computers where we don't know which OCC version is used.

tests/unittests/core/3d/occmodeltest.cpp
```diff
@@ -207,6 +207,31 @@ TEST_F(OccModelTest, testTesselate) {
...
+  if (OccModel::isAvailable()) {
+    // Create a board with a semi-transparent color.
+    const Path outline = Path::rect(Point(Length(0), Length(0)),
+                                   Point(Length(10000), Length(10000)));
+    const QColor color(255, 0, 0, 128);  // Alpha = 0.5
+    std::unique_ptr<OccModel> model = OccModel::createBoard(
+        outline, {}, PositiveLength(1000), color);
+
+    const QMap<OccModel::Color, QVector<QVector3D>> result = model->tesselate();
+    EXPECT_EQ(result.count(), 1);
```
ubruhin: Better use `ASSERT_EQ()` here, to avoid crash on the `.first()` below if the returned `QSet` is empty (the test aborts if an `ASSERT` fails, in contrast to `EXPECT` which continues the execution):


```suggestion
    ASSERT_EQ(result.count(), 1);
```

Or alternatively, you could use `.value(0)` instead of `.first()`.

ubruhin (commented):
Awesome, thanks! The code actually looks really good 👍 Just a few notes:

- I added two comments regarding the new test, see there.
- CI currently fails as `occmodeltest.cpp` is not formatted properly. Are you able to run clang-format (see details [here](https://developers.librepcb.org/df/d24/doc_code_style_guide.html#doc_code_style_guide_format)) on it? Otherwise I can run it for you and push a formatting commit.
- I noticed that when exporting a board with transparent LEDs as STEP and open it in FreeCAD, the transparency is lost. I didn't investigate yet if the issue is in LibrePCB or in FreeCAD, but I think the export function `addToAssembly()` still accesses the old `tryGetColor()` which only supports RGB, so I expect this is the root cause. In case you want to implement support for RGBA also during export, I'd appreciate it, but I'd merge this PR also without this feature.

Girish4489:
I have addressed the review comments:
 - **RGBA Support**: Updated `OccModel::addToAssembly()` to preserve alpha channels when copying colors. This should improve transparency support when exporting to external tools.
 - **Unit Tests**: Switched to `ASSERT_EQ` for robustness and added `#if` guards for OCCT version compatibility.
 - **Formatting**: Ran the official `./dev/format_code.sh` script on all modified files (C++ and CMake).
 
 Regarding the export behavior in FreeCAD, while I've added RGBA support to the assembly core, I haven't performed deep testing with external tools as it's currently outside the scope of this fix.

ubruhin:
Oh, now something strange happens. In the export, the PCB becomes transparent instead of the LED. Seems like the transparency is applied to the wrong object:

image

> I haven't performed deep testing with external tools as it's currently outside the scope of this fix.

That's fine. I would therefore suggest to revert the export modification to get rid of this new problem, and keep only the unit test changes. I'm happy to merge it then. The transparency of the export is not important for now.

ubruhin (approved):
Thanks!

PR #1756, opened 2026-04-12
3D Viewer: Add 'F' shortcut for Zoom-to-Fit/Flip and improve button tooltips

Girish4489:
This PR enhances the usability and discoverability of the 3D board viewer by introducing a common keyboard shortcut and adding descriptive tooltips to the on-screen controls.

  Key Changes:
   * **New 'F' Keyboard Shortcut:** 
       * Added "F" as a secondary global shortcut for the "Zoom to Fit Contents" command (alongside `Ctrl+Home`).
       * In the 3D viewer, this enables a "Smart Fit/Flip" workflow: the first press fits the board to the screen, and the second press flips the board to show the opposite side.
       * This matches the existing behavior of the toolbar icon and is consistent with other EDA tools.
   * **Enhanced Accessibility (Tooltips):**
       * Added localized tooltips to the 3D viewer's **Zoom to Fit**, **Zoom In**, and **Zoom Out buttons**.
       * Tooltips now automatically display their respective keyboard shortcuts, helping users learn the interface more quickly.
   * **Context-Aware Focus:**
       * Modified `Board3dTab` to ensure that pressing "F" correctly captures focus and triggers the 3D action only when intended, avoiding conflicts with the 2D editor's shortcuts.

### DIFF
--- a/libs/librepcb/ui/library/pkg/packagetab.slint
+++ b/libs/librepcb/ui/library/pkg/packagetab.slint
@@ -607,6 +607,7 @@ component PackageEditorTab inherits TouchArea {
                 checked: tabs[index].tool == EditorTool.arc;
                 enabled: footprint-selected && (!read-only);
                 is-first: false;
+                is-last: false;
 
                 clicked => {
                     tool-triggered(TabAction.tool-arc);
@@ -827,7 +828,7 @@ component PackageEditorTab inherits TouchArea {
 
         zoom-fit-btn := SceneButton {
             x: parent.width - self.width;
-            icon: @image-url("../../../../font-awesome/svgs/solid/expand.svg");
+            command: Cmd.zoom-fit-content;
             icon-scale: 0.9;
             base-color: tabs[index].background-color;
             base-text-color: tabs[index].foreground-color;
@@ -840,7 +841,7 @@ component PackageEditorTab inherits TouchArea {
 
         zoom-in-btn := SceneButton {
             x: parent.width - self.width;
-            icon: @image-url("../../../../bootstrap-icons/icons/zoom-in.svg");
+            command: Cmd.zoom-in;
             base-color: tabs[index].background-color;
             base-text-color: tabs[index].foreground-color;
             enabled: footprint-selected;
@@ -852,7 +853,7 @@ component PackageEditorTab inherits TouchArea {
 
         zoom-out-btn := SceneButton {
             x: parent.width - self.width;
-            icon: @image-url("../../../../bootstrap-icons/icons/zoom-out.svg");
+            command: Cmd.zoom-out;
             base-color: tabs[index].background-color;
             base-text-color: tabs[index].foreground-color;
             enabled: footprint-selected;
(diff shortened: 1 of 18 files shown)


ubruhin:
I agree the "F" shortcut would be nice, but to be honest I'm not happy with the implementation:
- `EditorCommandSet` also exists in Slint, and should be accessed directly from there (not through C++). From this class you can get both the tooltip and the shortcut.
- Showing the shortcut in the tooltip should be a feature of `SceneButton`, and used for *any* scene button, not just in the 3D view. See `ToolButton` which also has a `shortcut` property.
- The displayed shortcut is coming from `EditorCommandSet` (which is user-configurable), but the function itself reacts on the hardcoded shortcut "f". Shortcut actions should not be implemented with `processSceneKeyPressed()`, please implement it in `shortcuts.slint` instead (which respects user-configured shortcuts). I think the easiest would be to just react on the already existing `Cmd.flip-view` command and emit the `TabAction.zoom-fit` action (just in the 3D view, as in the 2D view it is used differently). I think that could be a very simple implementation of this feature.
- The "f" shortcut must not be added to the "Zoom All" `EditorCommand`, since it creates a conflict for the 2D views.

libs/librepcb/ui/library/pkg/packagetab.slint
```diff
@@ -607,6 +607,7 @@ component PackageEditorTab inherits TouchArea {
                 checked: tabs[index].tool == EditorTool.arc;
                 enabled: footprint-selected && (!read-only);
                 is-first: false;
+                is-last: false;
```
ubruhin, line 610: Was this added by accident? It doesn't look correct for me.

libs/librepcb/ui/project/board/board3dtab.slint
```diff
@@ -9,6 +9,7 @@ import {
     Board3dTabData,
     BoardAction,
     Data,
+    EditorCommandSet,
```
ubruhin, line 12: For consistency with all other files, please import as Cmd:


```suggestion
    EditorCommandSet as Cmd,
```

libs/librepcb/ui/widgets/scenebutton.slint
```diff
@@ -7,10 +8,12 @@ export enum SceneButtonStyle {
 }
 
 export component SceneButton inherits Rectangle {
+    in property <EditorCommand> command;
     in property <string> text;
-    in property <image> icon;
+    in property <image> icon: command.icon;
     in property <float> icon-scale: 1.0;
     in property <string> tooltip;
```
ubruhin, line 15: This tooltip property seems now unused, right? Either it should be removed then, or assigned the default value `command.text`.

libs/librepcb/ui/widgets/scenebutton.slint
```diff
@@ -7,10 +8,12 @@ export enum SceneButtonStyle {
...
+    property <string> effective-tooltip: (command.id != "") ? (command.text + (command.shortcut != "" ? " (" + command.shortcut + ")" : "")) : tooltip;
```
ubruhin, line 16: Just nitpicking: I'd prefer `.is-empty` instead of performing a string comparison with "". But I'd accept it as-is.

And in case you keep the `tooltip` property, this assignment should actually use it (currently effective-tooltip does not take tooltip into account).

ubruhin:
@Girish4489 any plan to continue work on this?

PR #1806, opened 2026-05-12
FreeBSD: Implement getProcessNameByPid() with sysctl()

melak:
The current version depends on /proc being mounted which is not necessarily a given. This commit switches the implementation to use a fitting sysctl(3) call to retrieve the executable path (name) from the kernel, uncoupling the implementation from procfs.

### DIFF
--- a/libs/librepcb/core/systeminfo.cpp
+++ b/libs/librepcb/core/systeminfo.cpp
@@ -46,7 +46,7 @@
 #if defined(Q_OS_SOLARIS)
 #include <libproc.h>
 #endif
-#if defined(Q_OS_OPENBSD)
+#if defined(Q_OS_OPENBSD) || defined(Q_OS_FREEBSD)
 #include <sys/sysctl.h>
 #endif
 #elif defined(Q_OS_WIN32) || defined(Q_OS_WIN64)  // Windows
@@ -272,15 +272,30 @@ QString SystemInfo::getProcessNameByPid(qint64 pid) {
                        tr("proc_name() failed with error %1.").arg(errno));
   }
 #elif defined(Q_OS_FREEBSD)
-  char exePath[64];
-  char buf[PATH_MAX + 1];
-  sprintf(exePath, "/proc/%lld/file", pid);
-  size_t len = (size_t)readlink(exePath, buf, sizeof(buf));
-  if (len >= sizeof(buf)) {
-    return QString();  // process not running
+  char pathbuf[PATH_MAX] = { 0, };
+  size_t pathbufsz = PATH_MAX;
+  int mib[4] = { CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, static_cast<pid_t>(pid) };
+  int rc = sysctl(mib, 4, pathbuf, &pathbufsz, NULL, 0);
+  if (rc == -1) {
+    switch (errno) {
+      case ESRCH: // explicitly No such process
+        return QString();
+      break;
+      default: {
+        int saved_errno = errno;
+        char errbuf[512] = { 0, };
+        // Should 512 not be enough, the error string returned by strerror_r() will
+        // be truncated but still be nul-terminated
+        strerror_r(saved_errno, errbuf, sizeof(errbuf));
+        throw RuntimeError(__FILE__, __LINE__,
+                           tr("sysctl() failed with errno=%1 (%2).")
+                               .arg(saved_errno)
+                               .arg(QString::fromLocal8Bit(errbuf)));
+      }
+      break;
+    }
   }
-  buf[len] = 0;
-  processName = QFileInfo(QFile::decodeName(buf)).fileName();
+  processName = QFileInfo(QFile::decodeName(pathbuf)).fileName();
 #elif defined(Q_OS_OPENBSD)
   // https://man.openbsd.org/sysctl.2
   // NOTE: This will return only the first 16 bytes of the process name. If
--- a/libs/librepcb/core/systeminfo.cpp
+++ b/libs/librepcb/core/systeminfo.cpp
@@ -272,27 +272,30 @@ QString SystemInfo::getProcessNameByPid(qint64 pid) {
                        tr("proc_name() failed with error %1.").arg(errno));
   }
 #elif defined(Q_OS_FREEBSD)
-  char pathbuf[PATH_MAX] = { 0, };
+  char pathbuf[PATH_MAX] = {
+      0,
+  };
   size_t pathbufsz = PATH_MAX;
-  int mib[4] = { CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, static_cast<pid_t>(pid) };
+  int mib[4] = {CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME,
+                static_cast<pid_t>(pid)};
   int rc = sysctl(mib, 4, pathbuf, &pathbufsz, NULL, 0);
   if (rc == -1) {
     switch (errno) {
-      case ESRCH: // explicitly No such process
+      case ESRCH:  // explicitly No such process
         return QString();
-      break;
       default: {
         int saved_errno = errno;
-        char errbuf[512] = { 0, };
-        // Should 512 not be enough, the error string returned by strerror_r() will
-        // be truncated but still be nul-terminated
+        char errbuf[512] = {
+            0,
+        };
+        // Should 512 not be enough, the error string returned by strerror_r()
+        // will be truncated but still be nul-terminated.
         strerror_r(saved_errno, errbuf, sizeof(errbuf));
         throw RuntimeError(__FILE__, __LINE__,
                            tr("sysctl() failed with errno=%1 (%2).")
                                .arg(saved_errno)
                                .arg(QString::fromLocal8Bit(errbuf)));
       }
-      break;
     }
   }
   processName = QFileInfo(QFile::decodeName(pathbuf)).fileName();


ubruhin:
Thanks! Two quick questions:
- Did you run the unit test `SystemInfoTest.testGetProcessNameByPid` and it is successful?
- Is it guaranteed that `sysctl()` will add zero-termination to `pathbuf`? I suspect it does, but I did not see a clear statement in the docs. It's probably fine anyway as long as `pathbufsz < PATH_MAX`, but what happens in the edge case of `pathbufsz == PATH_MAX`? If in doubt, better explicitly enforce a final `\0`, e.g. by just increasing the array size to `PATH_MAX+1`

Btw, I'd remove `saved_errno` as accessing `errno` twice shouldn't be harmful. Just my opinion, doesn't need to be changed for merging this.

ubruhin:
Hmm according [docs](https://man.freebsd.org/cgi/man.cgi?query=strerror&sektion=0), `strerror` is not thread-safe. That should be fixed somehow.

melak:
1. I did not, can you give me a quick pointer as to how do I do that please?
2. sysctl will fill pathbuf (or the initial portion of, as needed). pathbuf is initialised to all nulls. PATH_MAX by definition includes terminating null. KERN_PROC_PATHNAME returns a "string". Add all of these together, I don't see how this could even in theory overrun/run out of bounds as a string. I can think of a small handful of ways to completely unnecessarily further "fortify" this, but it's just not necessary; also see for example here: https://github.com/freebsd/freebsd-src/blob/main/usr.bin/limits/limits.c#L724

Saving errno is protecting against pieces of throw and that chain of... stuff there possibly mangling the original (thrown by sysctl()) errno - to be honest I don't have a whole lot of ideas about how C++ works, so if you say it's unnecessary, I'm happy to take it out - this will need touching while I switch to strerror_r() anyway.

ubruhin:
> I did not, can you give me a quick pointer as to how do I do that please?

After you have built LibrePCB, just run the executable `./tests/unittests/librepcb-unittests` in the build folder. This will run all tests, but you can also filter the tests to run, see `--help` for details.

It would actually be even interesting if all the other tests pass too, as we never do this on FreeBSD :see_no_evil: We should do this from time to time...

> sysctl will fill pathbuf (or the initial portion of, as needed). pathbuf is initialised to all nulls. PATH_MAX by definition includes terminating null. KERN_PROC_PATHNAME returns a "string".

I see. I also think it is safe and your comment gives me more confidence too. I just want to be super careful about such low level memory management as it is very easy to accidentally open up a security vulnerability. Therefore I usually try to be safer than necessary (e.g. adding 1 item more to the array doesn't hurt at all, but ensures explicitly that there will be a final zero, no matter how `sysctl()` implements string handling exactly).

> Saving errno is protecting against pieces of throw and that chain of... 

I'm pretty sure it's unnecessary, but anyway, you can keep it as-is. It doesn't hurt.

melak:
So, from what I can tell, the deal is somewhere along the lines of, and this very well may be the case for OpenBSD as well, two of the tests test for non-existent processes, expecting QString() as the answer. In these cases FreeBSD definitely, OpenBSD pretty likely (I don't have one handy to try) will throw an ESRCH (No such process), because, obviously, the process targeted does not exist. This will come out as a RuntimeError from getProcessNameByPid().
Now, the Linux case reads

```
#elif defined(Q_OS_LINUX)

  // From:
  // http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/io/qlockfile_unix.cpp
  if (!FilePath("/proc/version").isExistingFile()) {
    throw RuntimeError(__FILE__, __LINE__,
                       tr("Could not find the file \"/proc/version\"."));
  }
  char exePath[64];
  char buf[PATH_MAX + 1];
  sprintf(exePath, "/proc/%lld/exe", pid);
  size_t len = (size_t)readlink(exePath, buf, sizeof(buf));
  if (len >= sizeof(buf)) {
    return QString();  // process not running
  }
  buf[len] = 0;
  processName = QFileInfo(QFile::decodeName(buf)).fileName();
  // If the executable does no longer exist, the string " (deleted)" is added to
  // the end of the symlink, so we need to remove that to get the naked process
  // name.
  if (processName.endsWith(" (deleted)"))
    processName.chop(strlen(" (deleted)"));
#elif defined(Q_OS_WIN32) || defined(Q_OS_WIN64)  // Windows
```
This is... just very wrong :)
If this is pointed to a non-existent PID, `/proc/pid/exe` wouldn't exist, `readlink` would fail, returning `-1`.
`size_t` is of course unsigned. `(size_t)-1` is a very big positive integer, which is always going to be `>= sizeof(tiny in comparison buf)` thus triggering "QString() on non-existent process".

This is basically by mistake (lack of error checking and expliciteness on an error path) and certainly doesn't feel intentional.

What happens with `buf[len] = 0` if `len` is `-1` is an entirely different can of worms :).

So, I think cleanup needs to begin here. I'm assuming you are most familiar with Linux, so with this in mind you can clean this up properly, laying the foundation for the protocol and fixing the other OSes (I'm happy to do it as well if you specify the desired protocol here).

(Also - this is `ifdef Q_OS_LINUX` - why is explicitly checking for `/proc/version` necessary?)

melak:
Part of the protocol should also be edge cases like "almost certain there's a process with that PID running but can't ascertain any more details (including but not limited to process name)", which may be a common scenario in hardened setups (on pretty much any OS, well maybe save for Windows, I don't know that).

ubruhin:
That code is now 10 years old and was initially copied from Qt. I think the reason for leaving the code as-is was that we simply don't have enough low-level knowledge about operating systems, Qt devs most probably know better. But when looking at the code, I fully agree it looks quite ugly and it would be time to clean up. It seems that also in Qt the code has changed quite a lot in the past 10 years.

> What happens with buf[len] = 0 if len is -1 is an entirely different can of worms :).

`len` cannot be -1 as it is an unsigned type. If it was -1 before the type cast, the `>=` conditional would return from the function before `buf[len]` is executed. Although the code is ugly, I don't see any safety issue with it. Despite the ugly error handling, at least the function is proven to work fine on Linux, MacOS and Windows for the past 10 years ;-)

> (Also - this is ifdef Q_OS_LINUX - why is explicitly checking for /proc/version necessary?)

Probably this was also coming from Qt. But I don't really remember. So far I also don't see a reason for this check.

Regarding intended behavior of the function, its documentation specifies that:

```cpp
/**
   * @brief Get the process name of a given PID
   *
   * @param pid   The process ID (may be a running process or not)
   *
   * @return  The name of the given process ID, or an empty string if no process
   *          with the given PID exists.
   *
   * @throw  Exception    In case of an error.
   */
```

It is true that the function may not work correctly in some sandboxed/hardened systems. So far I have only seen that in this case it returns an empty string, which is not nice, but fine for our use-case. An exception would be much worse I think. In the end, this function is only used to detect stale file locks and is therefore rarely called at all.

ubruhin:
Actually I think we should also eliminate the function `isProcessRunning()` as it has no advantage over `getProcessNameByPid()` - we can just call `getProcessNameByPid()` directly, and detect a non-running process by an empty string as return value.

ubruhin:
> Probably this was also coming from Qt. But I don't really remember. So far I also don't see a reason for this check.

I think I was wrong. Qt [still has this check](https://github.com/qt/qtbase/blob/33468f386f419e8be0b09f76957d667347b41299/src/corelib/kernel/qcore_unix_p.h#L343-L355) and probably it makes sense. It looks like there exist also Linux systems without procfs, in which case we want to be able to distinguish between "we cannot detect the running process" and "the process is not running".

melak:
OK, if this is about stale file locks, then the approach is wrong (or rather underspecified), I think. You can either lock a file (in which case it's not stale) or not, in which case someone else holds it locked possibly having written its PID into the lock file, which you want to use for _informational_ purposes.

Case in point, this `getProcessNameByPid` function has no _material_ effect on the outcome (being able to lock or not). It's a _convenience_ function.

If you _could_ lock the file, the _non-existing process_ case doesn't exist to begin with. If you could not lock the file, chances are (in this context) the conflicting process is also yours, so being able to get its name is all but certain as well. On the very off chance you can't get its name, well the user-facing message says "locked probably by PID x, it's your problem now", and it is also fine.

So in this light I think an approach of "executable name of PID x or QString() in any other case (error or 'there definitely is a process but I can't get to its details')" is perfectly fine, but then this needs to be clarified.

Re `/proc/version`, Qt does still have it, but it's a *case* _inside_ `Q_OS_LINUX`, not _determining whether this is Linux_ or whatever. Where does LibrePCB use "Linux but without procfs", especially in this function? It just doesn't need this information.

Also considering Linux had a long long time ago deprecated and (relatively recently) [removed](https://www.phoronix.com/news/Linux-5.5-Kills-SYSCTL-SYSCALL) sysctl(2) and the canonical way to do this kind of thing is now through `/proc/sys`, procfs-less Linux has basically been an oxymoron for longer than I care to remember. And further, that bit is the only place in Qt where this define is present, so I think it's safe to say to call it an accidental leftover.

ubruhin:
> Case in point, this getProcessNameByPid function has no material effect on the outcome (being able to lock or not). It's a convenience function.

That's not exactly true. If we can detect that the lock is hold by another, still running process, we tell the user that the file is already locked. But if the lock was obtained by an application that doesn't run anymore (i.e. it crashed and therefore didn't release the lock - the PID may even be reused by another application, so we have to check the process name), we detect it as a stale lock and silently override it. And there are even other cases where we warn the user about an existing lock but allow him to override the lock on his own risk. In fact, we distinguish between several different kinds of locks:

```cpp
  enum class LockStatus {
    /// The directory is not locked (lock file does not exist).
    Unlocked,
    /// The directory is locked by a crashed application instance.
    StaleLock,
    /// The directory is locked by this application instance.
    LockedByThisApp,
    /// The directory is locked by another application instance on this machine.
    LockedByOtherApp,
    /// The directory is locked by another user or machine.
    LockedByOtherUser,
    /// The directory is locked by an unknown application (may be stale).
    LockedByUnknownApp,
  };
```

Anyway, somehow I don't see why we should spend time on something that works fine since 10 years (lock detection on Windows, MacOS and Linux). Refactoring this code without a good reason is just an unnecessary risk. I see only three good reasons for modifying it:

- Improving the behavior for operating systems on which we don't have this confidence (e.g. FreeBSD, so your change is welcome). But this should not touch the other, known-to-be-working systems.
- If there are any security concerns with the implementation. That's not the case as far as I can see.
- Re-implementing this whole function in Rust. We will need it in Rust anyway sooner or later, so that would not be an unnecessary risk, it has to be done anyway.

melak:
> (i.e. it crashed and therefore didn't release the lock

That's where this whole scheme goes horribly sideways. In a proper system this case does not exist - if a process goes away, it's held resources are cleaned up, including any locks held.

Sidestepping this semantic with PIDs in the lockfile and whatnot is trying make locks out of building blocks inherently unsuitable for proper locking.

On Linux, you can for certain figure out (usual race conditions notwithstanding) what holds a lock by parsing `/proc/locks` (or scraping `lslocks`) and getting the information from there. Other systems have their own intricacies (or you can fall back to `lsof` for a portable interface and scraping its output).

_Everything else is handwaving_. This is _highly_ platform-dependant full of landmines.

There used to be a `QtSingleApplication` class but apparently it's not in Qt anymore, although there seem to be replacements still actively developed. If the goal is single instance, maybe those should be explored, or, as you hinted, use [Rust single instance](https://docs.rs/single-instance/latest/single_instance/) for example.

This should be a framework problem, not an application problem.

ubruhin:
There is a reason why we do it the way we do it. Our lock system has several advantages over the existing lock mechanisms:
- It is simple. No rocket science.
- The concept is platform independent (only the implementation requires a little platform-specific code, but the concept is generic)
- It works on every file system
- It works across file system boundaries, e.g. even in mounted network shares. If you put a project on a network drive and open it from 2 different computers, LibrePCB will tell you which other user in the network has this project already open.

I haven't seen a ready-to-use implementation which does all what we need. We don't want to restrict the user to run only a single instance of the application. It is perfectly valid to run multiple LibrePCB instances at the same time as long as they don't access the same workspace/project/library. So we need a lock mechanism on files, not on processes.

Also, the mechanism does not need to be bullet-proof. There is no harm at all if a lock file is not cleaned up when the process crashes. A crash is an extremely rare situation, and the resulting "harm" is that on the next application start the application detects the lock as stale and overrides it. No harm at all. In worst case, if the stale detection fails too, the user will get a "Override the lock?" dialog where he can press "Yes". No problem at all to get this message box once in 2 years. No files will be held locked, no files will be corrupted. Why should we spend days of our time to implement a highly sophisticated, using-native-technologies-on-every-system, need-to-study-documentations technologies if there is no problem with our simple concept?

Don't get me wrong - I am not against moving to a better system **if there is any**, and it provides clear advantages. But I haven't seen any. Also you just complain about our concept, but didn't tell us what would be the better solution. It's easy to complain, but it's not easy to find a better solution (and to actually spend the time to *implement* the better solution).

melak:
Right, I see where I've gone wrong here. It's not a _lock_ in the systems sense, it's a post-it note on the workspace saying "please don't touch it". For that, it's perfectly fine.

I'll update the PR so that tests pass.

melak:
```
[----------] 5 tests from SystemInfoTest
[ RUN      ] SystemInfoTest.testGetUsername
Username: ice
[       OK ] SystemInfoTest.testGetUsername (0 ms)
[ RUN      ] SystemInfoTest.testGetFullUsername
Full username: User &
[       OK ] SystemInfoTest.testGetFullUsername (0 ms)
[ RUN      ] SystemInfoTest.testGetHostname
Hostname: freebsd15
[       OK ] SystemInfoTest.testGetHostname (0 ms)
[ RUN      ] SystemInfoTest.testIsProcessRunning
[       OK ] SystemInfoTest.testIsProcessRunning (24 ms)
[ RUN      ] SystemInfoTest.testGetProcessNameByPid
[       OK ] SystemInfoTest.testGetProcessNameByPid (430 ms)
[----------] 5 tests from SystemInfoTest (454 ms total)
```

I'll send you the full unittest output in a private email for reference.

PR #1813, opened 2026-05-19
Schematic editor: Keep netlines orthogonal when dragging symbols

### LINKED ISSUE #1701: Schematic Editor: Wires should stay orthogonal when you move components
When you move a component the wires become angled as shown here:

https://github.com/user-attachments/assets/ffe61b36-1462-4019-8c2f-35c33b0f5d6a

You have to tediously re-align them. I don't think I've ever seen a schematic with non-orthogonal wires, so it makes waaay more sense to do this automatically. Even if it might mess them up a bit sometimes it's better than not trying at all.

Maybe there is a setting for this (I looked but couldn't find it), in which case the bug is that it should be enabled by default.

Btw this is the first time I've tried LibrePCB after trying almost every other free PCB software. I'm quite impressed - lots of nice touches that show you care about usability. Even with this one which is the first rough edge I found... at least you didn't do the idiotic thing that KiCAD does - just leave the wires behind and disconnect them!

SiboVG:
This changes schematic component dragging so connected wires stay orthogonal automatically where possible.

The new behavior kicks in when moving selected symbols with connected schematic wires. In particular, it handles:

- pin-to-pin wires, by inserting an L-bend instead of leaving a diagonal wire
- pin-to-netpoint-to-pin chains, by moving the adjacent netpoint and adding a bend at the next segment when needed
- horizontal and vertical multi-segment runs, while only adjusting the nearest affected vertex instead of pulling an entire collinear chain along
- undo/redo and live preview during the drag

The routing decision itself is implemented in Rust in `librepcb-rust-core`. The Rust side works on a small generic graph of anchors and wires, so it is not tied to schematic editor classes and can be unit-tested directly. The C++ schematic editor code builds this graph from `SI_SymbolPin`, `SI_NetPoint`, and `SI_NetLine`, calls the Rust FFI, then applies the returned movements using the existing preview and undo command infrastructure.

I kept the actual schematic mutation in C++ because the editor object model, live drag preview, and undo/redo ownership are all still C++/Qt-based. The new C++ code is therefore mostly adapter and command plumbing, while the generic wire movement calculation lives in Rust.

Tests added:

- Rust unit tests for the generic orthogonal wire movement algorithm
- C++ integration tests for schematic dragging behavior, including pin-to-pin, pin-netpoint-pin, vertical multi-segment, no-op drag, rotate-during-drag cleanup, and return-to-axis preview cleanup cases

Fixes #1701

Demo:

https://github.com/user-attachments/assets/441614a9-acff-4f3a-b375-1fcef2c6477d

### DIFF
--- a/libs/librepcb/editor/project/cmd/cmddragselectedschematicitems.cpp
+++ b/libs/librepcb/editor/project/cmd/cmddragselectedschematicitems.cpp
@@ -41,19 +41,127 @@
 #include <librepcb/core/project/schematic/items/si_netlabel.h>
 #include <librepcb/core/project/schematic/items/si_netline.h>
 #include <librepcb/core/project/schematic/items/si_netpoint.h>
+#include <librepcb/core/project/schematic/items/si_netsegment.h>
 #include <librepcb/core/project/schematic/items/si_polygon.h>
 #include <librepcb/core/project/schematic/items/si_symbol.h>
 #include <librepcb/core/project/schematic/items/si_symbolpin.h>
 #include <librepcb/core/project/schematic/items/si_text.h>
 #include <librepcb/core/project/schematic/schematic.h>
+#include <librepcb/rust-core/ffi.h>
 
 #include <QtCore>
+
+#include <memory>
+#include <vector>
 
 /*******************************************************************************
  *  Namespace
  ******************************************************************************/
 namespace librepcb {
 namespace editor {
+
+/*******************************************************************************
+ *  Local Classes
+ ******************************************************************************/
+namespace {
+
+// Preview helper for a boundary netline which cannot be kept orthogonal by
+// moving an existing netpoint. It replaces one line with an L-bend at the
+// moving side, then either gets discarded during preview or appended to the
+// final undo command.
+class CmdSchematicNetLineSplit final : public UndoCommand {
+public:
+  CmdSchematicNetLineSplit(SI_NetLine& line, SI_NetLineAnchor& movingAnchor,
+                           SI_NetLineAnchor& fixedAnchor)
+    : UndoCommand(QObject::tr("Split netline")),
+      mNetSegment(line.getNetSegment()),
+      mOldLine(line),
+      mBendNetPoint(new SI_NetPoint(mNetSegment, Uuid::createRandom(),
+                                    movingAnchor.getPosition())),
+      mLineToMoving(new SI_NetLine(mNetSegment, Uuid::createRandom(),
+                                   movingAnchor, *mBendNetPoint,
+                                   line.getWidth())),
+      mLineToFixed(new SI_NetLine(mNetSegment, Uuid::createRandom(),
+                                  *mBendNetPoint, fixedAnchor,
+                                  line.getWidth())),
+      mApplied(false) {}
+
+  ~CmdSchematicNetLineSplit() noexcept override {
+    if (mApplied && (!wasEverExecuted())) {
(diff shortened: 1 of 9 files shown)


ubruhin:
Awesome! The architecture with the generic logic in Rust and the integration in C++ is exactly how I wished it to have :muscle: 

I didn't look at the code though, as I am a bit busy these days. This may take a few days. But I quickly checked it out and played around to get a feeling about its capabilities. So far I have two comments:

Probably it should also move net labels to avoid leaving a mess behind (also see on the right side of the symbol, where the situation is even worse):

[Peek 2026-05-19 22-05.webm](https://github.com/user-attachments/assets/8e81351d-3141-4ca6-ac4e-5f172398212f)

And it happens very easily that "dangerous"/misleading situations are the result, like this:

[Peek 2026-05-19 22-10.webm](https://github.com/user-attachments/assets/82a7f891-96c5-404f-ab31-3127f1aec3ff)

I mean, it is clear that generally this can happen on every move operation and the ERC should be responsible to detect and report such situations. However, the old behavior without any "smartness" at least was much less dangerous as the problems were easily visible: 

image

I guess in the end, the orthogonal wire feature should also take some responsibility about not leaving a dangerous situation behind. Maybe as a last resort, just disable orthogonal movement of wires which would end up in a visual short-circuit between different nets (i.e. falling back to the old direct-wire approach)? Or any other ideas?

SiboVG:
> Probably it should also move net labels to avoid leaving a mess behind (also see on the right side of the symbol, where the situation is even worse):

Good suggestion, thanks!

> I guess in the end, the orthogonal wire feature should also take some responsibility about not leaving a dangerous situation behind. Maybe as a last resort, just disable orthogonal movement of wires which would end up in a visual short-circuit between different nets (i.e. falling back to the old direct-wire approach)? Or any other ideas?

Hm, that's tricky. I'm thinking an easy solution for now would be to always use diagonal lines when you're moving an IC with more than one pin on the side that is being moved.

SiboVG:
> > I guess in the end, the orthogonal wire feature should also take some responsibility about not leaving a dangerous situation behind. Maybe as a last resort, just disable orthogonal movement of wires which would end up in a visual short-circuit between different nets (i.e. falling back to the old direct-wire approach)? Or any other ideas?
> 
> Hm, that's tricky. I'm thinking an easy solution for now would be to always use diagonal lines when you're moving an IC with more than one pin on the side that is being moved.

This is what it would look like (I have a commit ready for it, but will wait for your approval first):

https://github.com/user-attachments/assets/5908fc4f-baaf-49f8-a17f-f9fe94d0f84e

SiboVG:
As a sidenote, I want to give kudos to you @ubruhin and the other LibrePCB devs! This is a great tool. I've been using many EDA tools over the years, but this is one of the most intuitive ones. Thank you for all your hard work!

ubruhin:
> As a sidenote, I want to give kudos to you @ubruhin and the other LibrePCB devs! 

Thank you, much appreciated 😊

Regarding the PR, please be patient a bit - I'm quite busy with many things these days, but I'll come back to this PR as soon as possible. It's also quite a complex thing with many things to consider, so it's not done in a few minutes...

SiboVG:
> Regarding the PR, please be patient a bit - I'm quite busy with many things these days, but I'll come back to this PR as soon as possible. It's also quite a complex thing with many things to consider, so it's not done in a few minutes...

Yeah sure, no rush, take your time.

PR #1815, opened 2026-05-22
Fix project migration log not opening in Snap/Flatpak

ubruhin:
Fixes https://librepcb.discourse.group/t/unable-to-see-messages-from-file-format-upgrade/1036


ubruhin:
Changed the implementation to store the temporary migration log in the project's logs directory instead of `~/.cache`, since the Firefox and Chromium Snaps reject to open local HTML files from hidden directories :sob: See https://bugs.launchpad.net/snapd/+bug/1972762

Issue #1524, opened 2025-05-15
Design SVG icons for the new UI


ubruhin:
##### SUMMARY

With the new UI, we're now using monochrome SVG icons from Font Awesome and Bootstrap Icons. These are great libraries, but there are still some specific icons missing for LibrePCB. It would be nice to get contributions from people who are talented in icon design as I am absolutely not.

If anyone is interested to help, please let me know so I'll make a list then which icons we need.

##### SOLUTION

There are some technical requirements on the icons:

- Must be SVG
- Must have no color set. They will be colorized as needed at runtime.
- They should be as small as possible in file size. No unnecessary XML nodes/attributes should be present. Whitespaces/linebreaks should/could be minimized as well, and of course the actual SVG content should be as simple as possible. Check out the Font Awesome / Bootstrap icons to get an idea how they can be really small :)
- They must be released under CC0-1.0 with the copyright specified as "LibrePCB Developers" to keep license handling as simple as possible. Of course we're happy to list icon contributors in our AUTHORS.md.

z3ugma:
@ubruhin I can work on this one as well if you've got a list of icons which need an SVG replacement that matches the icon style of the rest of the icon set you're using

ubruhin:
Great! Here are some icons to be created:

![Image](https://github.com/user-attachments/assets/f0ff964f-6ba7-4d45-91ea-c614f7bb83ae)

![Image](https://github.com/user-attachments/assets/25b6b0d0-618b-42d8-96a3-6e452fe0f90a)

More will be needed for the new library editor which is currently work in progress.

Note that I think designing icons can only be done if you can review them in the real UI because for standalone icons one cannot see if the look is consistent with the rest of the UI (e.g. similar line thickness etc.). The best solution for that is to run the [appwindow.slint](https://github.com/LibrePCB/LibrePCB/blob/master/libs/librepcb/editor/ui/appwindow.slint) file in the Slint live previewer. In VSCode you could install the Slint extension (or Slint Nightly) which provides syntax highlighting and a live previewer. Of course alternatively you can compile LibrePCB to see the result, but the feedback loop is much shorter when using the Slint preview.

And of course feel free to also look for [Font Awesome](https://fontawesome.com/icons) or [Bootstrap Icons](https://icons.getbootstrap.com/) first, maybe I just didn't find what I was looking for.

[...]

PR #1826, opened 2026-05-27
Replace some icons with new, custom SVG icons

SiboVG:
## Summary
This PR starts addressing #1524 by replacing several app UI icons with custom SVG icons.

Included in this first batch:
- Add custom SVG icons for schematic documents/views
- Add separate icons for schematic wire drawing and PCB trace drawing
- Add custom icons for output jobs, pin numbering, holes/drills, and vias
- Replace output jobs launcher/dialog/list icons, including the sidebar play icon
- Register the new SVG assets in `img/images.qrc`
- Add REUSE metadata for the new LibrePCB-owned CC0-1.0 icons

image
image

## Notes

This is only a first step toward #1524. It does not implement all requested icon updates.

In particular, the monochrome icon requests are not included in this PR and still need follow-up work.

## Icon design

Most icons were designed in Affinity Designer. To ensure small file size, the following precautions were taken:

- Artbaord was kept small (16x16, 20x20, 24x24)
- Boolean-combined shapes (Layer > Geometry > Add)
- Converted text to curves 'Layer > Convert to Curves)
- Deleted hidden/unused layers

Then I exported the file as SVG, preset 'SVG (for export)'. Affinity does add a lot of junk to the SVG, so I used the [SVGO tool](https://github.com/svg/svgo) to make the file even smaller by running `npx svgo input.svg -o output.svg`

A few icons were also generated by Gemini Pro 3.1, and then often modified in Affinity to match my needs.


ubruhin:
Nice! Generally I like the look of the new icons, and how compact the SVG files are :+1: Just a few comments:

- The schematic icon is the only one which doesn't fully convince me due to 2 reasons:
  - It is square, but schematics are usually landscape. I suspect the icon would be more intuitive if its aspect ratio is (slightly) landscape?
  - Somehow the icon sometimes looks a bit blurry/bold, at least compared to the board icon. Maybe it is only a rendering issue, but maybe it also has a thicker line width than the board icon. A slightly thinner line width (same as the board icon) may lead to a more consistent look?
  image
- I prefer the new output jobs icon over the old one, but not at every location in the UI. At the right side of the output jobs list (marked in red), the old icons were better IMO because they clearly indicated the "run" action. So I'd just restore the old icons here:
  image
- In the output jobs dialog, the icons color does not change with the selected theme. Instead of constructing a `QIcon` directly from the SVGs, load them with `EditorToolbox::svgIcon()` which will apply a theme-dependent color.
  image
- Could you move all the new icons from `img/actions/` one level up, i.e. into `img/`? I'm sorry this is not documented anywhere. The subdirectories `img/actions` and `img/command_toolbars` are considered legacy since we started to use monochrome icons, and with the subdirectories we can easily see which icons are already replaced and which are still legacy. The idea is to replace icons step by step until these two directories are empty.
- Same as in the other PR: The entries in `.reuse/dep5` should be merged & sorted to make CI pass.

SiboVG:
Thanks for the review @ubruhin! Your comments should be fixed in the latest commit.

image

libs/librepcb/editor/project/schematic/fsm/schematiceditorstate_drawwire.cpp
```diff
@@ -849,14 +849,14 @@ std::optional<NetSignal*>
                        });
   QMenu menu;
   menu.setDefaultAction(
-      menu.addAction(QIcon(":/img/actions/draw_wire.png"),
+      menu.addAction(QIcon(":/img/actions/draw_wire.svg"),
```
ubruhin: It's not in the `/actions/` subdirectory anymore.

libs/librepcb/editor/project/schematic/fsm/schematiceditorstate_drawwire.cpp
```diff
@@ -849,14 +849,14 @@ std::optional<NetSignal*>
...
                      tr("Add New Bus Member") +
                          QString(" (%1)").arg(QCoreApplication::translate(
                              "QShortcut", "Ctrl"))));
   NetSignal* selectedNet = nullptr;
   for (NetSignal* net : std::as_const(nets)) {
     QAction* a =
-        menu.addAction(QIcon(":/img/actions/draw_wire.png"), *net->getName(),
+        menu.addAction(QIcon(":/img/actions/draw_wire.svg"), *net->getName(),
```
ubruhin: It's not in the `/actions/` subdirectory anymore.

libs/librepcb/ui/library/pkg/packagetab.slint
```diff
@@ -557,7 +557,7 @@ component PackageEditorTab inherits TouchArea {
         ToolButtonGroup {
             tool-line-btn := ToolButton {
                 cmd: Cmd.tool-line;
-                icon: @image-url("../../../../../img/actions/draw_wire.png");
+                icon: @image-url("../../../../../img/draw_wire.svg");
```
ubruhin: This line can/should now be removed completely, since the icon is already contained in `cmd`.

Same applies to several other places where the images are used in `ToolButton` elements (symbol editor, schematic editor, board editor).

ubruhin (commented):
Since 2.1.1 is now released, we can get this finished, so I reviewed again. I found some final things:

- See comments.
- The new image files are named snake_case, but for new images we have switched to kebab-case. Could you please rename the files accordingly? Sorry that I didn't see this on my first review...
- As there is a merge conflict now, the branch should be rebased.

SiboVG:
Right, thanks for the review @ubruhin. I think it's ok now.

ubruhin (approved):
Thanks! Just found one theming problem during testing, fixed by myself by adding missing `EditorToolbox::svgIcon()`.

This change is really appreciated, the new icons look very nice! If you ever have more ideas about possible icon improvements (even such not listed in #1524), I'm happy about suggestions 🙂

PR #1828, opened 2026-05-27
macOS: Add Applications shortcut to DMG installer

### LINKED ISSUE #1824: Add Applications folder shortcut to macOS DMG installer
##### SUMMARY
The macOS `.dmg` installer currently only shows the LibrePCB application icon. There is no visible shortcut to the Applications folder, so users have to manually open Applications in Finder and drag LibrePCB there.

This is less convenient than the usual macOS DMG installation flow, where the installer window shows both the app icon and an Applications folder shortcut, often with an arrow indicating that the app should be dragged there.

The attached screenshot shows the current DMG window, where only the LibrePCB app is visible.

Image

##### SOLUTION
The LibrePCB macOS `.dmg` installer should include an Applications folder shortcut next to the LibrePCB app icon.

Expected behavior:

- Opening the .dmg shows the LibrePCB app icon.
- The same window also shows a shortcut/alias to the Applications folder.
- Users can install LibrePCB by dragging the app icon onto the Applications folder shortcut.
- Optionally, the DMG background could include a simple arrow or visual hint from the app icon to the Applications folder.

This would make installation more intuitive and consistent with common macOS application installers.

Example:

Image

SiboVG:
Fixes #1824 by using an Applications shortcut in the .dmg installer Volume. I created a custom background for the installer:

Screenshot 2026-05-27 at 23 05 23


ubruhin:
Awesome! Even with some nice artwork :smiley: :tada: 

Just two comments:
- Did you remove `--skip-jenkins` by intention? It seems some people report issues with running `create-dmg` on CI (not only Jenkins AFAIK) and that flag seems to fix those issues. So generally I'd feel better to keep it unless there is a good reason to remove.
- For consistency with other deployment resources, could you move the images into the `dist` folder in the repo root, resp. into a new subdirectory of it (called maybe `dmg` or `macos` or so). Unfortunately this would implicitly change its license to GPLv3 due to the glob in `.reuse/dep5`, but I still think this is the right place for those images so let's just list them explicitly in `.reuse/dep5` to assign CC0 to them (if you agree to that).
- Just to confirm: I assume the SVG is the source file, and the PNG is an export of it?

Btw, note that since v2.1.0 has just been released recently, I am currently only merging bugfix PRs which shall land in 2.1.1. I'll merge this after 2.1.1 has been released, to avoid the risk of introducing new issues in a bugfix release. But of course PRs can already be prepared, I'll just hold them back for a short moment.

SiboVG:
> * Did you remove `--skip-jenkins` by intention? It seems some people report issues with running `create-dmg` on CI (not only Jenkins AFAIK) and that flag seems to fix those issues. So generally I'd feel better to keep it unless there is a good reason to remove.

Unfortunately, I had to remove it, because `create-dmg` skips the Finder AppleScript when `--skip-jenkins` is set. Without that AppleScript, the background image and icon positions are not applied, so the DMG would contain the Applications link but still not show the custom styled drag-and-drop layout.

> * For consistency with other deployment resources, could you move the images into the `dist` folder in the repo root, resp. into a new subdirectory of it (called maybe `dmg` or `macos` or so). Unfortunately this would implicitly change its license to GPLv3 due to the glob in `.reuse/dep5`, but I still think this is the right place for those images so let's just list them explicitly in `.reuse/dep5` to assign CC0 to them (if you agree to that).

Right, fixed now!

> * Just to confirm: I assume the SVG is the source file, and the PNG is an export of it?

That's right. Well, I actually created it in Affinity Designer, but it wouldn't make sense to track that file. An SVG is universal for any vector editor.

ubruhin:
> Unfortunately, I had to remove it, because create-dmg skips the Finder AppleScript when --skip-jenkins is set. 

Ah, I see. Let's hope then it doesn't cause troubles on CI :see_no_evil: 

The PR looks good to me now. The only thing - why CI fails - is that the entry in `.reuse/dep5` needs to be merged into the already existing CC0 section (in alphabetical order). Afterwards I'm happy and I'll merge once the bugfix release 2.1.1 is out :muscle:

SiboVG:
> The PR looks good to me now. The only thing - why CI fails - is that the entry in `.reuse/dep5` needs to be merged into the already existing CC0 section (in alphabetical order). Afterwards I'm happy and I'll merge once the bugfix release 2.1.1 is out 💪

Great. I think it's ok now. I also rebased to the latest master branch.

ubruhin (approved):
Thanks! 😃

ubruhin:
Finally I was able to actually test this. On macOS 10.14 Intel it looks good. On a macOS ARM64 (not sure which macOS version it was), there is a scrollbar at the bottom:

image

Not sure why this happens and how it is related to the macOS version / CPU architecture.

It's also nothing critical, just a bit ugly. Maybe we could try to increase the window size by a few pixels to get rid of the scrollbar, though it might not be worth the effort.

@SiboVG what do you think, do you want to give it a try or should we keep it as-is?

EDIT: Here are the builds from CI: https://download.librepcb.org/nightly_builds/macos-dmg/

SiboVG:
> Finally I was able to actually test this. On macOS 10.14 Intel it looks good. On a macOS ARM64 (not sure which macOS version it was), there is a scrollbar at the bottom:
> image
> 
> Not sure why this happens and how it is related to the macOS version / CPU architecture.
> 
> It's also nothing critical, just a bit ugly. Maybe we could try to increase the window size by a few pixels to get rid of the scrollbar, though it might not be worth the effort.
> 
> @SiboVG what do you think, do you want to give it a try or should we keep it as-is?
> 
> EDIT: Here are the builds from CI: https://download.librepcb.org/nightly_builds/macos-dmg/

Huh, that's odd, since I also compile on ARM64. But I agree that it's ugly. I'll increase the window size.

SiboVG:
@ubruhin since I cannot recreate the issue, could you verify that the scroll bar is gone now?

ubruhin (approved):
Thanks, now the scrollbar is gone! 🎉

PR #1832, opened 2026-05-31
Improve macOS keyboard shortcuts look-and-feel and input fallbacks

### LINKED ISSUE #1817: Mac-Specific User Interface Tweaks (and an introduction)
##### SUMMARY

(Running MacOS Sonoma 14.8.2 Intel)

Item 1: Mouse-hover tooltips in LibrePCB show keyboard control-key combinations labeled with "CTRL", while the actual key required to perform the action on a mac is the Command or Open-Apple key. Further, the actual CTRL key on the Mac keyboard does not perform the function.

I don't know if it's an easy fix in your localization of the tooltip phrases, but ideally when running in MacOS tooltips should display the Command-key icon "⌘" (Unicode "Place of Interest" symbol, see: https://www.compart.com/en/unicode/U+2318 ) rather than "CTRL".

In addition, most Mac keyboards have a separate CTRL key. Would it be feasible for LibrePCB to accept both CTRL or Command for the same action? This might help people who migrate between Windows PCs and Macs regularly. (On the other hand, it may be useful to reserve that extra key for future functionality.)

Item 2: Delete key. On Macs (not sure about Windows), delete and backspace are distinct. But most MacOS programs accept Backspace as a "delete selected object" function where appropriate. It would be nice if LibrePCB would accept either Backspace or Delete for these actions. (No need to update the tooltips.)

##### MISC

Now a brief intro: I selected LibrePCB this weekend to design my first PCB. I'm not a complete newbie, but close enough - when I was a kid, I would help my father with hand-drawn, hand photo-etched project boards, and played around with Radio Shack project kits and various breadboard tinkering. In the 90s at a tech startup, I gave input to the occasional PCB design but never directly did any design work. So while I know what a via is, there are a lot of the fundamentals that I do not know.

I have some hobby projects that I've been meaning to realize, that require custom PCBs. So it's time to learn.

I realize LibrePCB is new and somewhat incomplete, and not fully documented. But I didn't want to start out on a large, possibly bloated, legacy system, even if better documented. Based on reviews and forum posts, I like the stated ideals of LibrePCB, and features such as its library structure. (Although I'm immediately finding the incompleteness of the libraries to be a problem - I've had to design my own components, so I'm definitely jumping into this with both feet.)

I mention all this because I will be posting more issues and feature requests - I hope that's not presumptuous of me, and that having the perspective of a newbie (not just to LibrePCB but to PCB design in general) will be of some value to the project.

Thanks much for dedicating yourself to this project.

SiboVG:
This PR addresses the macOS keyboard shortcut and tooltip feedback from #1817. It improves the native look-and-feel on macOS and ensures that shortcuts work seamlessly with common Mac keyboard layouts.

### Key Changes
* **Show Native macOS Symbols in UI (#1817 Item 1, Part 1)**:
  * Updated tooltips, menus, and status messages to display native symbols (⌘, ⇧, ⌥, ⌃) instead of text labels like "Ctrl" on macOS.
  * Added `EditorToolbox::modifierKeyText()` to handle bare modifier keys in info box and status bar hints.

image

Note that macOS typically doesn't use `+` for key combinations.

* **Support interchangeable Control & Command Modifiers (#1817 Item 1, Part 2)**:
  * On macOS, Qt maps the Command key to `Qt::ControlModifier`. We now dynamically register fallbacks using the physical Control key (`Qt::MetaModifier`) for all default/custom Command key shortcuts.
   
* **Support interchangeable Delete & Backspace Keys (#1817 Item 2)**:
  * Mac keyboards typically feature a key labeled "delete" in place of Backspace, which produces `Qt::Key_Backspace`.
  * Added `Qt::Key_Backspace` as a dynamic runtime fallback for all shortcuts using `Qt::Key_Delete` (such as object removal, sheet removal, etc.) on macOS.
*Note: All fallback mappings are computed dynamically at runtime using `EditorCommand::getShortcutsWithMacFallbacks()`. This ensures the UI settings dialog, exported shortcut reference PDF, and tooltips remain clean and show only the primary shortcut.*

### Verification & Testing
* Built the project successfully with CMake (Qt 6).
* All 2591 unit tests build and pass successfully.
* Verified that on macOS:
  * Both `⌘ + Z` and `⌃ + Z` trigger Undo.
  * Both Backspace (labeled delete) and Fn + Delete keys successfully delete selected objects in the editors.
  * Tooltips and menus display clean modifier symbols (e.g. `⌘Z`).

Fixes #1817.

### DIFF
--- a/libs/librepcb/ui/mainmenubar.slint
+++ b/libs/librepcb/ui/mainmenubar.slint
@@ -434,7 +434,7 @@ export component MainMenuBar {
                     text: @tr("Cut");
                     icon: @image-url("../../font-awesome/svgs/solid/scissors.svg");
                     status-tip: @tr("Cut the selected object(s) to clipboard");
-                    shortcuts: "Ctrl+X";
+                    shortcuts: Cmd.clipboard-cut.shortcut;
                     enabled: Data.current-tab.features.cut == FeatureState.enabled;
 
                     clicked => {
@@ -447,7 +447,7 @@ export component MainMenuBar {
                     text: @tr("Copy");
                     icon: @image-url("../../font-awesome/svgs/regular/copy.svg");
                     status-tip: @tr("Copy the selected object(s) to clipboard");
-                    shortcuts: "Ctrl+C";
+                    shortcuts: Cmd.clipboard-copy.shortcut;
                     enabled: Data.current-tab.features.copy == FeatureState.enabled;
 
                     clicked => {
@@ -460,7 +460,7 @@ export component MainMenuBar {
                     text: @tr("Paste");
                     icon: @image-url("../../font-awesome/svgs/solid/paste.svg");
                     status-tip: @tr("Paste object(s) from the clipboard");
-                    shortcuts: "Ctrl+V";
+                    shortcuts: Cmd.clipboard-paste.shortcut;
                     enabled: Data.current-tab.features.paste == FeatureState.enabled;
 
                     clicked => {
--- a/libs/librepcb/editor/editorcommand.cpp
+++ b/libs/librepcb/editor/editorcommand.cpp
@@ -60,6 +60,53 @@ EditorCommand::~EditorCommand() noexcept {
 }
 
 /*******************************************************************************
+ *  Static Methods
+ ******************************************************************************/
+
+QList<QKeySequence> EditorCommand::getShortcutsWithMacFallbacks(
+    const QList<QKeySequence>& sequences) noexcept {
+#if defined(Q_OS_MACOS)
+  QList<QKeySequence> result = sequences;
+  for (const QKeySequence& seq : sequences) {
+    if (seq.isEmpty()) continue;
+    if (seq.count() == 1) {
+      QKeyCombination comb = seq[0];
+      Qt::KeyboardModifiers modifiers = comb.keyboardModifiers();
+      Qt::Key key = comb.key();
+
+      // Fallback 1: If it has Command (Qt::ControlModifier), add a fallback with Control (Qt::MetaModifier)
+      if (modifiers.testFlag(Qt::ControlModifier)) {
+        Qt::KeyboardModifiers fallbackModifiers = (modifiers & ~Qt::ControlModifier) | Qt::MetaModifier;
+        QKeySequence fallback(QKeyCombination(fallbackModifiers, key));
+        if (!result.contains(fallback)) {
+          result.append(fallback);
+        }
+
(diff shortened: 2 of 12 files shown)


libs/librepcb/ui/mainmenubar.slint
```diff
@@ -434,7 +434,7 @@ export component MainMenuBar {
                     text: @tr("Cut");
                     icon: @image-url("../../font-awesome/svgs/solid/scissors.svg");
                     status-tip: @tr("Cut the selected object(s) to clipboard");
-                    shortcuts: "Ctrl+X";
+                    shortcuts: Cmd.clipboard-cut.shortcut;
```
ubruhin, line 437: Oh strange that this was hardcoded here :sob: But actually the properties `text` and `status-tip` are duplicated too. Could you remove `text`, `status-tip` and `shortcuts` from this element, and add this line instead?

```slint
cmd: Cmd.clipboard-cut;
```

As far as I can see, that should lead to identical properties as hardcoded before but avoids duplicate translation strings.

The `icon` property is still needed because the command does not have an icon set (we cannot do this yet, as long as the command is also still used in legacy Qt widgets where we need PNG instead of SVG).

Same applies to the other two commands.

ubruhin (commented):
Nice! I think the `EditorToolbox::modifierKeyText()` is a no-brainer and can be even included in the next bugfix release. But for the shortcut fallbacks I see a bit of a risk of unintended side effects, thus not ideal to include in a bugfix release. I'll need to think about it if this is really a desired and safe change.

Therefore, would it be possible to move the changes for the shortcut fallbacks into a separate PR? That would be great.

ubruhin:
@SiboVG Would you be able to make the suggested changes in the next few days? I think we should release 2.1.1 soon and it would be very nice to include the mac-specific modifier texts (as mentioned above, only the displayed texts for now, no changed behavior).

But no rush, I'm not sure yet when will be the release date. And if you are too busy, just let me know, I could then split the PR by myself.

SiboVG:
> @SiboVG Would you be able to make the suggested changes in the next few days? I think we should release 2.1.1 soon and it would be very nice to include the mac-specific modifier texts (as mentioned above, only the displayed texts for now, no changed behavior).
> 
> But no rush, I'm not sure yet when will be the release date. And if you are too busy, just let me know, I could then split the PR by myself.

Sure, does #1846 match what you meant? Do you want me to close this PR?

ubruhin:
Perfect, thanks! We can keep this PR open as it still contains a useful change. We can rebase it after the 2.1.1 release.

PR #1846, opened 2026-06-08
Show native modifier key text on macOS

SiboVG:
This is a subset of PR #1832 which adds `EditorToolbox::modifierKeyText()` to display platform-specific keyboard modifier text in editor messages.

On macOS, modifier keys are shown with native symbols such as `⌘` and `⇧`, matching the actual keys used by Qt shortcuts. On other platforms, the existing translated text such as `Ctrl` and `Shift` is kept.

Also removes duplicated Cut/Copy/Paste menu item text, status tips and shortcut strings from `mainmenubar.slint` by binding these items to their existing editor commands while keeping their SVG icons.

This intentionally does not include the macOS shortcut fallback changes from the original branch.

### DIFF
--- a/libs/librepcb/editor/library/pkg/fsm/packageeditorstate_drawpolygonbase.cpp
+++ b/libs/librepcb/editor/library/pkg/fsm/packageeditorstate_drawpolygonbase.cpp
@@ -537,8 +537,8 @@ void PackageEditorState_DrawPolygonBase::updateOverlayText() noexcept {
 void PackageEditorState_DrawPolygonBase::updateStatusBarMessage() noexcept {
   QString note = " " %
       tr("(press %1 to disable snap, %2 to abort)")
-          .arg(QCoreApplication::translate("QShortcut", "Shift") % "/" %
-               QCoreApplication::translate("QShortcut", "Ctrl"))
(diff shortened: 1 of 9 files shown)


ubruhin (approved):
Thanks a lot! Very nice improvement.

I just pushed a commit to fix the formatting.

ubruhin:
Hmm I was too fast with merging this :sob: I found two issues:

- The menus still show Ctrl/Shift etc. instead of the macOS symbols. That change should have been applied too from #1832.
- On Intel macOS, those macOS symbols are rendered correctly in the workspace settings dialog (Qt), but not in the main window (Slint): 
  image
  On ARM64 it seems to work properly.

Not sure if there will be an easy fix for this. If not, we will have to revert this change completely :-/

One thing I could imagine is that our bundled font doesn't contain those glyphs and ARM64 somehow silently falls back to some other font. I'll check if we will have to include those glyphs in our font.

ubruhin:
> One thing I could imagine is that our bundled font doesn't contain those glyphs and ARM64 somehow silently falls back to some other font. I'll check if we will have to include those glyphs in our font.

I think that was the problem. Fixed in #1848, will test when the nightly build is ready.

PR #1859, opened 2026-06-26
docs: replace requirements list with summary table

nizarbadriali:
Replaced the bullet list in the requirements section with a summary table in order to make it easier to scan dependencies, versions, and whether each is required or optional at a glance.

No content has been removed — all information from the original list is preserved in the table.
No code has been created or removed

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,17 +50,15 @@ productive use, please install an official release as described in the
 
 To compile and run LibrePCB, you need the following software components:
 
-- C++ compiler: g++, MinGW or Clang (any version with C++20 support should work)
-- [Rust](https://www.rust-lang.org/) toolchain >= 1.92 (GNU, not MSVC)
-- [Qt](http://www.qt.io/download-open-source/) >= 6.2 (make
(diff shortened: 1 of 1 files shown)


ubruhin:
Hi,

Thanks for the contribution. However, to be honest I prefer the old list for two reasons:

- The information of the list is more accurate than the table. For example the table does not mention anymore that the imageformats plugin will be needed at runtime. Without this hint, people send us feedback that LibrePCB compiled without this plugin, claiming that the docs are wrong (this happened in the past, so we added that hint). Also for other dependencies IMHO the old list was more accurate and more informative than the new table which lost several of the (important) details.
- The list is very easy to read in a simple text editor. The table looks nice only with a markdown renderer, but is difficult to read in a text editor. Since the readme is intended to be read locally in a simple text editor (remember it is also part of our source distributions, for people not downloading from github), I prefer the easy-to-read variant.

Thus closing this PR - sorry about that.

For a general recommendation how to contribute to LibrePCB, please have a look at https://librepcb.org/contribute/.

PR #1865, opened 2026-07-19
 Schematic editor: Merge collinear netlines after dragging 

### LINKED ISSUE #1823: Schematic Editor: Merge collinear wire segments after moving components or wires
##### SUMMARY
When dragging and releasing a component or wire in the schematic editor, LibrePCB should automatically merge connected collinear wire segments into a single wire segment.

Currently, moving a component or wire can result in two separate wire segments which are directly connected and collinear. Visually, these look like one continuous wire, but internally they remain separate segments.

This can cause annoying editing behavior afterwards. For example, as shown in the attached video, after moving a component it can look like the connected wire is straightened correctly. However, because the wire is still split into multiple collinear segments, moving the component up afterwards can cause one of the wire segments to overlap with the component instead of behaving like a single clean wire.

https://github.com/user-attachments/assets/92351c60-9419-4a89-9a6f-33dbe3f1c44d

##### SOLUTION
After finishing a drag operation, LibrePCB should check whether any affected connected wire segments are collinear and directly connected. If so, they should be merged into a single wire segment.

This should apply when:

- A component is dragged and released, changing the attached net geometry.
- A wire segment is dragged and released.
- The operation results in two or more connected wire segments being collinear.

Expected behavior:

- Connected horizontal wire segments on the same line should be merged.
- Connected vertical wire segments on the same line should be merged.
- The merge should happen after releasing the mouse button.
- The resulting schematic should contain one clean wire segment instead of multiple overlapping or directly adjacent collinear segments.

This would keep the schematic geometry cleaner and avoid surprising behavior when editing the same component or net again afterwards.

SiboVG:
## Description

Automatically simplify affected schematic net and bus segments after dragging components, wires, or junctions. LibrePCB already had code capable of merging collinear wire segments: `CmdSimplifySchematicSegments`. The missing part was calling it after dragging.

Moving a component or wire could previously leave directly connected collinear netlines as separate internal objects. Although these lines appeared as one continuous wire, later editing operations could produce overlapping or unexpected wire geometry.

The drag command now collects the affected segments before movement. Once the drag has been committed on mouse release, the existing schematic segment simplification command is pushed onto the undo stack as a separate action.

Segments connected to moved component pins are included explicitly because their netlines are not part of the selection query. Label-only moves do not trigger simplification.

Keeping both operations separate provides the following undo/redo behavior:

- The first undo reverses simplification while keeping the dragged geometry. 
- The second undo reverses the drag.
- Redo reapplies the drag first and simplification second.

Fixes #1823.

https://github.com/user-attachments/assets/6fd49f87-8a71-420c-9e6a-2b5fb4741bfc

## Testing

- Added a regression test covering collinear netline merging after dragging.
- Verified that the first undo restores the moved, split netlines.
- Verified that the second undo restores the original drag position.
- Verified the corresponding two-step redo behavior.
- Passed all executed unit tests.
- One unrelated UI tab-order test was skipped.
- Passed the project formatting check.
- Successfully built the LibrePCB application.

### DIFF
--- a/libs/librepcb/editor/project/cmd/cmddragselectedschematicitems.cpp
+++ b/libs/librepcb/editor/project/cmd/cmddragselectedschematicitems.cpp
@@ -29,6 +29,7 @@
 #include "../../project/cmd/cmdschematicbuslabeledit.h"
 #include "../../project/cmd/cmdschematicnetlabeledit.h"
 #include "../../project/cmd/cmdschematicnetpointedit.h"
+#include "../../project/cmd/cmdsimplifyschematicsegments.h"
 #include "../../project/cmd/cmdsymbolinstanceedit.h"
 #include "../../project/cmd/cmdsymbolinstancetextsreset.h"
 #include "../schematic/schematicgraphicsscene.h"
@@ -37,10 +38,13 @@
 #include <librepcb/core/project/project.h>
 #include <librepcb/core/project/schematic/items/si_busjunction.h>
 #include <librepcb/core/project/schematic/items/si_buslabel.h>
+#include <librepcb/core/project/schematic/items/si_busline.h>
+#include <librepcb/core/project/schematic/items/si_bussegment.h>
 #include <librepcb/core/project/schematic/items/si_image.h>
 #include <librepcb/core/project/schematic/items/si_netlabel.h>
 #include <librepcb/core/project/schematic/items/si_netline.h>
 #include <librepcb/core/project/schematic/items/si_netpoint.h>
+#include <librepcb/core/project/schematic/items/si_netsegment.h>
 #include <librepcb/core/project/schematic/items/si_polygon.h>
 #include <librepcb/core/project/schematic/items/si_symbol.h>
 #include <librepcb/core/project/schematic/items/si_symbolpin.h>
@@ -86,6 +90,29 @@ CmdDragSelectedSchematicItems::CmdDragSelectedSchematicItems(
   query.addSelectedImages();
   query.addJunctionsOfBusLines();
   query.addNetPointsOfNetLines();
+
+  // Remember all segments whose geometry is affected by the drag. For symbols,
+  // the connected net lines are not part of the selection query, so collect
+  // their segments explicitly through the pins.
+  for (SI_BusLine* line : query.getBusLines()) {
+    mBusSegmentsToSimplify.insert(&line->getBusSegment());
+  }
+  for (SI_BusJunction* junction : query.getBusJunctions()) {
+    mBusSegmentsToSimplify.insert(&junction->getBusSegment());
+  }
+  for (SI_NetLine* line : query.getNetLines()) {
+    mNetSegmentsToSimplify.insert(&line->getNetSegment());
+  }
+  for (SI_NetPoint* point : query.getNetPoints()) {
+    mNetSegmentsToSimplify.insert(&point->getNetSegment());
+  }
+  for (SI_Symbol* symbol : query.getSymbols()) {
+    for (SI_SymbolPin* pin : symbol->getPins()) {
+      for (SI_NetLine* line : pin->getNetLines()) {
+        mNetSegmentsToSimplify.insert(&line->getNetSegment());
+      }
+    }
+  }
 
   // Find the center of all elements and create undo commands.
   foreach (SI_Symbol* symbol, query.getSymbols()) {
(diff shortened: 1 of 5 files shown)


libs/librepcb/editor/project/cmd/cmddragselectedschematicitems.cpp
```diff
@@ -87,6 +90,29 @@ CmdDragSelectedSchematicItems::CmdDragSelectedSchematicItems(
...
+    mNetSegmentsToSimplify.insert(&line->getNetSegment());
+  }
+  for (SI_NetPoint* point : query.getNetPoints()) {
+    mNetSegmentsToSimplify.insert(&point->getNetSegment());
+  }
+  for (SI_Symbol* symbol : query.getSymbols()) {
+    for (SI_SymbolPin* pin : symbol->getPins()) {
+      for (SI_NetLine* line : pin->getNetLines()) {
+        mNetSegmentsToSimplify.insert(&line->getNetSegment());
+      }
```
ubruhin: Remove one for loop:


```suggestion
      if (SI_NetSegment* segment : pin->getNetSegmentOfLines()) {
        mNetSegmentsToSimplify.insert(segment);
      }
```

libs/librepcb/editor/project/cmd/cmddragselectedschematicitems.h
```diff
@@ -70,6 +72,12 @@ class CmdDragSelectedSchematicItems final : public UndoCommandGroup {
   void setCurrentPosition(const Point& pos) noexcept;
   void rotate(const Angle& angle, bool aroundCurrentPosition) noexcept;
   void mirror(Qt::Orientation orientation, bool aroundCurrentPosition) noexcept;
+  const QSet<SI_NetSegment*>& getModifiedNetSegments() const noexcept {
+    return mNetSegmentsToSimplify;
+  }
+  const QSet<SI_BusSegment*>& getModifiedBusSegments() const noexcept {
+    return mBusSegmentsToSimplify;
+  }
```
ubruhin, line 80: It would be nice to have the same terminology for both the methods and the member variables ("modified segments" vs. "segments to simplify").

libs/librepcb/editor/project/schematic/fsm/schematiceditorstate_select.cpp
```diff
@@ -520,7 +520,15 @@ bool SchematicEditorState_Select::processGraphicsSceneLeftMouseButtonReleased(
...
-      execCmd(mSelectedItemsDragCommand.release());  // can throw
+      const QSet<SI_NetSegment*> netSegments =
+          mSelectedItemsDragCommand->getModifiedNetSegments();
+      const QSet<SI_BusSegment*> busSegments =
+          mSelectedItemsDragCommand->getModifiedBusSegments();
+      if (execCmd(mSelectedItemsDragCommand.release()) &&
+          ((!netSegments.isEmpty()) || (!busSegments.isEmpty()))) {
+        execCmd(new CmdSimplifySchematicSegments(netSegments,
+                                                 busSegments));  // can throw
+      }
```
ubruhin, line 535: I would prefer to wrap the execution of CmdSimplifySchematicSegments in an additional try/catch block. Somehow I don't have 100% trust in this undo command, i.e. there might be a small risk that it fails in rare cases and I don't want to show an error message box in that case as it would be confusing/cumbersome for the user. It's not critical if CmdSimplifySchematicSegments fails, so we can just ignore such an error.

Some day, once we are confident that CmdSimplifySchematicSegments is really reliable, we can still remove the additional try/catch block.

ubruhin (commented):
Very nice! 👍 I just wonder why CI fails, it looks like there is a bug in the code which makes the unit tests crash.

tests/unittests/editor/project/schematic/cmddragselectedschematicitemstest.cpp
```diff
@@ -0,0 +1,150 @@
...
+  ProjectLoader loader;
+  std::unique_ptr<Project> project =
+      loader.open(std::make_unique<TransactionalDirectory>(fileSystem),
+                  projectFile.getFilename());
+  Schematic& schematic = *project->getSchematics().first();
+  Circuit& circuit = project->getCircuit();
+  auto netSignal = new NetSignal(circuit, Uuid::createRandom(),
+                                 *circuit.getNetClasses().first(),
+                                 CircuitIdentifier("N"), false);
+  netSignal->addToCircuit();
```
ubruhin: This must not be called directly, call `Circuit::addNetSignal()` instead.

Maybe this was the reason for the unit test crashes(?)

SiboVG:
Got it, thanks @ubruhin! I pushed a fix to the comments and test failure.

ubruhin (approved):
Seems to work fine, thanks! Now let's hope this doesn't introduce too much of unintended behavior 🙈

PR #1866, opened 2026-07-22
CLI: Run headless without requiring an X server

### LINKED ISSUE #1793: CLI: Support running headless without requiring xvfb
Currently the `librepcb-cli` requires an X server on Linux to run, which is cumbersome for headless systems. In our `librepcb-cli` Docker image we use a wrapper that calls `xvfb-run`:

https://github.com/LibrePCB/docker-librepcb-cli/blob/master/librepcb-cli

But I think it should be easily possible to get rid of this workaround. When I run `librepcb-cli -platform minimal`, it seems to work fine even in a Docker container. Also `-platform offscreen` works, and the env var `QT_QPA_PLATFORM` works as well.

So if such a command line argument or even env var works, I guess it should be possible to adjust our `main()` somehow to enforce the minimal platform plugin, to avoid requiring any workarounds on user side. Or, if that is really not possible, at least we should replace `xvfb-run` in our Docker image with the env var `QT_QPA_PLATFORM`.

- [ ] Research which platform makes more sense - `minimal` or `offscreen`?
- [ ] Investigate how we can enforce the CLI to run with a specific platform plugin (with only C++ code, no wrappers or so). Maybe just call `qputenv()` to set `QT_QPA_PLATFORM=minimal`?
- [ ] Implement & test the solution (also verify that the fix doesn't break MacOS or Windows!)
- [ ] Update https://github.com/LibrePCB/docker-librepcb-cli accordingly (on the next release)

firuzakhmad:
Fixes #1793.

librepcb-cli needs an X server on Linux right now because
QGuiApplication defaults to the "xcb" platform plugin. This forces
"offscreen" instead (unless QT_QPA_PLATFORM is already set), so the
CLI can run in Docker/CI without xvfb-run.

Went with "offscreen" rather than "minimal" as both were mentioned in
the issue, because the export commands do real rendering
(QPainter/QImage/QPrinter/QPdfWriter), and offscreen is the backend
meant for that. minimal is just a stub.

## Testing

Ubuntu 24.04, Qt 6.4.2.

Before, with no DISPLAY/WAYLAND_DISPLAY:

    ./librepcb-cli --version
    [FATAL] This application failed to start because no Qt platform
    plugin could be initialized.

After:

    ./librepcb-cli --version
    LibrePCB CLI Version ...

Also checked the export path itself, not just startup:

    ./librepcb-cli open-project tests/data/projects/DRC/project.lpp \
      --export-schematics /tmp/test.pdf

Produces a correct, non-empty PDF headlessly.

Also confirmed setting QT_QPA_PLATFORM manually still works (not
overridden), and normal desktop use with a display is unaffected.

The change is behind #ifdef Q_OS_LINUX, so it's a no-op on
macOS/Windows, CI should confirm nothing there changes.

Haven't touched the docker-librepcb-cli image yet, can follow up
with that separately once this lands, as the issue suggests.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -123,7 +123,7 @@ jobs:
         run: xvfb-run -a ./build/tests/unittests/librepcb-unittests
         if: ${{ (steps.build.outcome == 'success') && (!cancelled()) }}
       - name: Run LibrePCB-CLI Functional Tests
-        run: xvfb-run -a --server-args="-screen 0 1024x768x24" uv --directory ./tests/cli run --no-dev pytest -vvv --librepcb-executable="../../build/install/bin/librepcb-cli"
+        run: uv --directory ./tests/cli run --no-dev pytest -vvv --librepcb-executable="../../build/install/bin/librepcb-cli"
         if: ${{ (steps.build.outcome == 'success') && (!cancelled()) }}
       - name: Run LibrePCB Functional Tests
         run: xvfb-run -a --server-args="-screen 0 1024x768x24" uv --directory ./tests/funq run --no-dev pytest -vvv --librepcb-executable="../../build/install/bin/librepcb" --reruns 5 --reruns-delay 10
--- a/apps/librepcb-cli/main.cpp
+++ b/apps/librepcb-cli/main.cpp
@@ -47,6 +47,16 @@ int main(int argc, char* argv[]) {
   // Creates the Debug object which installs the message handler. This must be
   // done as early as possible.
   Debug::instance();
+  
+#ifdef Q_OS_LINUX
+  // Avoid requiring a real X server / Xvfb just to run the CLI headlessly.
+  // "offscreen" works fine for our PDF/image/gerber export (unlike
+  // "minimal", which doesn't actually render anything). Don't override if
+  // the user already set QT_QPA_PLATFORM themselves.
+  if (qEnvironmentVariableIsEmpty("QT_QPA_PLATFORM")) {
+    qputenv("QT_QPA_PLATFORM", "offscreen");
+  }
+#endif
 
   // Silence logging output, it's a command line tool where logging messages
   // could lead to issues when parsing the CLI output. Real errors will be
--- a/apps/librepcb-cli/main.cpp
+++ b/apps/librepcb-cli/main.cpp
@@ -47,13 +47,15 @@ int main(int argc, char* argv[]) {
   // Creates the Debug object which installs the message handler. This must be
   // done as early as possible.
   Debug::instance();
-  
+
 #ifdef Q_OS_LINUX
-  // Avoid requiring a real X server / Xvfb just to run the CLI headlessly.
-  // "offscreen" works fine for our PDF/image/gerber export (unlike
-  // "minimal", which doesn't actually render anything). Don't override if
-  // the user already set QT_QPA_PLATFORM themselves.
-  if (qEnvironmentVariableIsEmpty("QT_QPA_PLATFORM")) {
+  // Only force offscreen rendering if there's truly no display available
+  // and the user hasn't picked a platform themselves. This keeps existing
+  // setups working (e.g. CI running under xvfb-run) and only kicks in for
+  // genuinely headless environments.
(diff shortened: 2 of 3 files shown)


firuzakhmad:
Pushed a follow-up fix after realizing the first version broke the funq test suite in CI.

After digging through the CI logs, I found that the packaged install build only includes the `xcb` platform plugin, not `offscreen`. Those tests were already running successfully under `xvfb-run`, since it provides a virtual display that `xcb` can use. Forcing `offscreen` ended up replacing a working setup with one that failed.

I've updated the logic so `QT_QPA_PLATFORM` is only set when neither `DISPLAY` nor `WAYLAND_DISPLAY` is available, which is the actual headless scenario this issue is meant to handle. If a real or virtual display (like Xvfb) is already present, it now leaves the environment unchanged.

ubruhin:
Thanks for working on this!

> After digging through the CI logs, I found that the packaged install build only includes the xcb platform plugin, not offscreen. 

Hmm so then the binary releases we build on CI will still not work without xvfb? Then IMO the real problem is still not solved.

I see that our binary releases (e.g. [librepcb-nightly-linux-x86_64.tar.gz](https://download.librepcb.org/nightly_builds/master/librepcb-nightly-linux-x86_64.tar.gz)) contain a file `plugins/platforms/libqxcb.so`. Do I suspect correctly that for `offscreen` to work, there will be a separate plugin needed? In that case, we should adjust our CI to include that plugin. Otherwise people will still run into the problem that our CLI doesn't run without xvfb.

Generally I feel like our CI should not use xvfb-run anymore to run the CLI tests - the CI should actually *prove* that the CLI runs without xvfb. When we keep xvfb-run, we will not notice if we (accidentally) re-introduce the X server dependency some day.

Maybe we just need to install the offscreen plugin in our CI Docker images (https://github.com/LibrePCB/docker-librepcb-dev)?

firuzakhmad:
Ahh! Good questions, dug into both.

You're right that offscreen needs its own plugin - on Debian/Ubuntu,
libqt6gui6 only ships libqxcb.so, offscreen/minimal/vnc are in a
separate package, qt6-qpa-plugins. So the CI/dev image would need
that installed for a genuinely headless test to work.

Same story for the release binaries. ci/build_linux_portables.sh
bundles Qt plugins via `linuxdeploy-x86_64.AppImage --plugin qt`, and
linuxdeploy-plugin-qt only bundles libqxcb.so by default. There's an
EXTRA_PLATFORM_PLUGINS env var meant exactly for this - adding
EXTRA_PLATFORM_PLUGINS=libqoffscreen.so to that call should get
offscreen into the actual release tarball/AppImage.

Also noticed the release script's own sanity check further down
still wraps the CLI in xvfb-run, so that'd want updating too once
offscreen is actually bundled.

Agreed on dropping xvfb-run from the CLI test job once the runner
has the plugin - only way to actually prove it rather than assume
it.

Happy to send a follow-up PR for the build_linux_portables.sh change
and the CI workflow change. docker-librepcb-dev is your call, or let
me know if you'd rather I open a PR there too.

Enjoying working through this one, happy to keep going.

ubruhin:
I just checked the Docker image `librepcb/librepcb-dev:ubuntu-22.04-qt6.11-1`, and the file `/opt/qt/plugins/platforms/libqoffscreen.so` already exists there. So I think the Docker images don't need any change.

> Happy to send a follow-up PR for the build_linux_portables.sh change and the CI workflow change.

Please just push those changes to this PR branch. I like to review & test the whole change at once before merging anything.

firuzakhmad:
Pushed both. Left the funq/ui test steps as is since those drive the
actual GUI editor and need a real display.

ubruhin (approved):
LGTM, thanks!

firuzakhmad:
Thanks for the review!

Regarding docker-librepcb-cli, I'll hold off on touching that repo
until 2.2.0 actually ships, then follow up with a PR there to
simplify the wrapper script.

ubruhin:
> Regarding docker-librepcb-cli, I'll hold off on touching that repo until 2.2.0 actually ships, then follow up with a PR there to simplify the wrapper script.

Alright :+1: You could also open a draft PR if you like, which we just keep open until 2.2.0 is ready. Up to you.

PR #1867, opened 2026-07-22
BoardPlaneFragmentsBuilder: Respect netclass clearances

### LINKED ISSUE #1863: Per-netclass clearance not respected by planes
Not sure if that was intentionally or a mistake, but planes do not seem to respect the clearance value defined for net classes. This makes it difficult to create e.g. planes that keep a small clearance to low voltage nets, and a higher clearance to high voltage nets.

To be investigated how to fix this.

LibrePCB 2.1.0

giri256:
## Summary

- Respect each net's `min_copper_copper_clearance` when rebuilding board plane fragments.
- Apply the effective maximum clearance symmetrically between planes and foreign-net planes, vias, traces, polygons, and pads.
- Preserve same-net thermal behavior and each plane's explicit manual clearance.
- Invalidate plane fragments when netclass design rules change.
- Add focused coverage for geometry and rebuild scheduling.

## Implementation notes

Netclass clearances are captured on the main thread before the threaded geometry jobs begin, then stored in immutable job data. This avoids accessing project model objects from worker threads.

The existing plane-fragment and Gerber golden snapshots predate netclass-specific plane geometry. They explicitly neutralize the fixture's special rule so those tests remain focused on their original scenarios; the new dedicated regression validates the real 3.33 mm rule.

## Verification

- The first CI run compiled successfully on Linux, macOS, and Windows.
- `Check / Clang Tidy` and `Check / Doxygen` passed.
- The exact `librepcb/librepcb-dev:devtools-7` formatter was run on every touched C++ file.
- `git diff --check` passed.
- The follow-up isolates the legacy Gerber golden comparison that failed after the corrected geometry changed its shared fixture.

Fixes #1863

### DIFF
--- a/libs/librepcb/core/project/board/boardplanefragmentsbuilder.h
+++ b/libs/librepcb/core/project/board/boardplanefragmentsbuilder.h
@@ -215,12 +215,13 @@ private:  // Methods
     QList<std::tuple<Transform, PositiveLength, NonEmptyPath>> holes;
     QList<TraceData> traces;  // Converted to polygons after preprocessing.
     std::shared_ptr<ClipperLib::Paths> boardArea;  // Populated in preprocessing
-    QHash<Uuid, UnsignedLength> netClassClearances;
+    QHash<Uuid, UnsignedLength> netSignalClearances;
 
     UnsignedLength getNetClassClearance(
-        const std::optional<Uuid>& netSignal) const noexcept {
-      return netSignal ? netClassClearances.value(*netSignal, UnsignedLength(0))
-                       : UnsignedLength(0);
+        const std::optional<Uuid>& netSignalUuid) const noexcept {
+      return netSignalUuid
+          ? netSignalClearances.value(*netSignalUuid, UnsignedLength(0))
+          : UnsignedLength(0);
     }
   };
 
--- a/libs/librepcb/core/project/board/boardplanefragmentsbuilder.h
+++ b/libs/librepcb/core/project/board/boardplanefragmentsbuilder.h
@@ -215,6 +215,14 @@ private:  // Methods
     QList<std::tuple<Transform, PositiveLength, NonEmptyPath>> holes;
     QList<TraceData> traces;  // Converted to polygons after preprocessing.
     std::shared_ptr<ClipperLib::Paths> boardArea;  // Populated in preprocessing
+    QHash<Uuid, UnsignedLength> netClassClearances;
+
+    UnsignedLength getNetClassClearance(
+        const std::optional<Uuid>& netSignal) const noexcept {
+      return netSignal ? netClassClearances.value(*netSignal,
+                                                  UnsignedLength(0))
+                       : UnsignedLength(0);
+    }
   };
 
   struct LayerJobResult {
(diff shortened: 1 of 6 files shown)


giri256:
I pushed `e44766137` to fix the exact CI findings: the touched C++ files are now formatted with the repository's `devtools-7` image, and the legacy Gerber golden test neutralizes the fixture's special netclass rule while the dedicated regression still tests the corrected clearance behavior. The fresh CI run is currently marked `action_required` with no jobs, so it appears to need maintainer approval before it can start.

libs/librepcb/core/project/board/boardplanefragmentsbuilder.h
```diff
@@ -215,6 +215,13 @@ class BoardPlaneFragmentsBuilder final : public QObject {
     QList<std::tuple<Transform, PositiveLength, NonEmptyPath>> holes;
     QList<TraceData> traces;  // Converted to polygons after preprocessing.
     std::shared_ptr<ClipperLib::Paths> boardArea;  // Populated in preprocessing
+    QHash<Uuid, UnsignedLength> netClassClearances;
+
+    UnsignedLength getNetClassClearance(
+        const std::optional<Uuid>& netSignal) const noexcept {
+      return netSignal ? netClassClearances.value(*netSignal, UnsignedLength(0))
```
ubruhin: Hmm I think here you mix net class UUIDs with net signal UUIDs. The hash table `netClassClearances` uses net class UUIDs as key (which makes sense), but for lookup you use net signal UUIDs.

ubruhin (commented):
Thank you for working on this. Regarding unit test, generally I prefer to keep using the golden sample approach, i.e. just updating the Gerber files and reviewing them manually if everything looks correct. I can do this update later.

Let me also first think about whether this is really the desired solution, and whether we really want this in a minor release or this should be considered a breaking change.

I am especially wondering about how we will allow in future to specify different clearances for different combinations of nets. For example if the net class for a 100V signal has a 2mm clearance, but there are several nets with 100 volts, they don't need 2mm clearance between each other, but only to nets with different voltage. Currently that use-case would not be possible since the clearance is applied to every net.

libs/librepcb/core/project/board/boardplanefragmentsbuilder.h
giri256: Thanks for flagging this. The current job data is intentionally keyed by net-signal UUID: all plane/via/trace/polygon/pad records carry the signal UUID for same-net comparisons, and `createJob()` stores that signal's netclass-derived clearance under the same signal UUID. The old `netClassClearances` name made it look class-keyed. In `b45c9fa89` I renamed it to `netSignalClearances` and made the lookup parameter `netSignalUuid`, so the key contract is explicit. I left the broader clearance semantics unchanged pending your design/release decision.

giri256:
Pushed `b45c9fa89` to address the UUID-domain concern without pre-empting the broader design discussion. The immutable job map and all geometry records are consistently keyed by net-signal UUID; the map and lookup names now state that explicitly. `git diff --check` passes. I have intentionally not changed the clearance model or golden-test strategy while those remain under maintainer consideration.

PR #1868, opened 2026-07-25
CMake: Fix duplicate target errors when building with Qt Creator

### LINKED ISSUE #1812: Cant build with new QT Creator version
##### SUMMARY
I attempted to build LibrePCB using an older version of QT to get Windows 10 LTSB support. When i try Qt Creator 19.0.1 I got many Errors while opening cmakelist.txt. The first is:
```
\LibrePCB\CMakeLists.txt:162: error: add_library cannot create target "delaunay_triangulation" because another
target with the same name already exists.  The existing target is an
interface library created in source directory "/LibrePCB".
See documentation for policy CMP0002 for more details.

Call stack:
  CMakeLists.txt:162 (find_package)
  cmake/FindDelaunayTriangulation.cmake:6 (add_library)
```
Qt Creator 14.0.2 works fine and I can succesfully build LibrePCB. Google suggestion helps with first error, so I think there is a chance that it is a cmake file problem and not my QT setup.

firuzakhmad:
Fixes #1812.

As you mentioned QtCreator injects
-DCMAKE_PROJECT_INCLUDE_BEFORE:FILEPATH=.../qtcreator-project.cmake,
and that unsetting it makes the problem go away. Looked into why -
CMAKE_PROJECT_INCLUDE_BEFORE gets re-included before every project()
call in the whole tree, not just the top-level one. libs/corrosion
declares its own project(Corrosion...) right before
find_package(DelaunayTriangulation) runs in the main CMakeLists.txt,
so with that hook set, the module effectively gets processed twice
in one configure. add_library() then fails because the target from
the first pass is still there.

Fix is just to guard the target creation with if(NOT TARGET ...), so
a second pass is a no-op instead of a fatal error.

Tested this two ways:

1) Isolated repro: a small CMakeLists.txt that calls
  find_package(DelaunayTriangulation) twice in a row. On the
  original module this fails with the exact same error and CMP0002
  reference from the issue. With the fix, both calls succeed and the
  vendored library message only prints once, so the normal
  single-invocation path is unaffected.
2) Same repro opened directly in Visual Studio on Windows using its
  built-in CMake support. With the original module, this reproduced
  the exact same CMP0002 error. With the fix applied, it configured
  and built successfully.

Also did a normal configure of the full project with this change to
make sure nothing else broke.


ubruhin:
Hmm I'm a bit confused. It sounds like the problem is in libs/corrosion, but the fix is in FindDelaunayTriangulation.cmake? Somehow that sounds not right to me. Shouldn't the fix be then to wrap corrosion in a `if(NOT TARGET ...)` conditional? Though that would be cumbersome as we use the upstream corrosion as-is...

I just don't understand how DelaunayTriangulation is related to this problem.

firuzakhmad:
Went back and actually tested this properly instead of guessing and you were right to push back.

To directly answer your question: DelaunayTriangulation isn't
actually related to this at all. It just happens to be the first
find_package() call after Qt6 in the real CMakeLists.txt, so it's
the first one to collide.

Confirmed corrosion isn't involved either: a minimal repro with just
find_package(Qt6) followed by find_package(DelaunayTriangulation),
no corrosion anywhere, reproduces the exact same CMP0002 error under
the real qtcreator-project.cmake hook. Adding corrosion back in
doesn't change anything either way.

Also tried disabling it properly by setting
QT_CREATOR_ENABLE_PACKAGE_MANAGER_SETUP to OFF via CACHE...FORCE,
both after and before project(). Neither works, since
CMAKE_PROJECT_INCLUDE_BEFORE fires as the first step of project()
itself, after any set() we could put before it, and the hook seems
to reassert the IDE's live setting with FORCE regardless. So this
isn't something fixable from CMakeLists.txt, matches what you found
too, that unsetting it is a Qt Creator build-config setting, not a
project-level one.

More importantly: it's not specific to DelaunayTriangulation at all.
Testing find_package(Qt6) -> DelaunayTriangulation -> Dxflib in
sequence, Dxflib fails too (with a different symptom - "binary
directory already used" from its add_subdirectory, rather than
add_library's CMP0002). Looking through the rest of cmake/Find*.cmake,
the same unguarded pattern (add_library/add_subdirectory with no
NOT TARGET check) also exists in FontoBeneQt, MuParser, Polyclipping,
TypeSafe, GTest, GLU, OpenCascade, and Slint, therefore, in principle any of
these could hit the same issue depending on what order Qt Creator's
auto-setup happens to touch them in.

Given that, I think the fix needs to be broader than just
FindDelaunayTriangulation.cmake, probably guarding each of these the
same way (NOT TARGET for the add_library cases, similar idea for the
add_subdirectory ones). Wanted to check with you before expanding the
PR that much though, is that the scope you'd want, or would you
rather handle this differently such as a single shared helper macro
these Find modules call into, rather than repeating the guard in
each file?

ubruhin:
It doesn't sound right to me to add such a conditional to every library. What I wonder:

- Does the problem actually still exist with the latest qtcreator release?
- Is the real problem really in our CMake files, or is qtcreator doing something nasty? If the real problem is actually located in qtcreator, I would propose to fix the problem there instead of applying some ugly workaround.

firuzakhmad:
I read through the actual maintenance_tool_provider.cmake file Qt Creator writes into the
build dir that hooks every
find_package() call. If a module doesn't set <PackageName>_FOUND, it assumes the first call failed and runs find_package() again so that's what collides with the target already created.

Yes, it still happens on 20.0.0, current release, so not something already
fixed upstream.

None of the modules ever set _FOUND. CMake's own bundled ones do (https://github.com/Kitware/CMake/blob/master/Modules/FindZLIB.cmake) guards the exact same way (if(NOT TARGET ZLIB::ZLIB) around its add_library). So this isn't really a Qt Creator workaround.

Set _FOUND properly in all the affected modules, kept NOT TARGET as a backstop. Tested with the real provider in Qt Creator 20.0.0 and full project configures clean now, same case that failed before.

ubruhin:
Thanks for clarifying. So then I agree with the approach. I just saw that the _FOUND variables were not set for some cases when system libraries were used, which I quickly fixed by myself. Also I refactored the if(NOT TARGET) pattern to an early return in every cmake file to make it also work for system libraries, not only for bundled libraries.

firuzakhmad:
Great, thank you for reviewing everything!

PR #1873, opened 2026-08-16
Add KiCad/EAGLE style trace dragging and copper pad expansion DR

codedrifter22:
## Description
- Added support for KiCad/EAGLE style trace dragging after drawing.
- Added copper pad expansion settings to Design Rules (DRC).

## Testing
- Tested functionality locally in the editor.


ubruhin:
Hi,

> KiCad/EAGLE style trace dragging after drawing

Can you elaborate what you mean exactly with this?

> Added copper pad expansion settings to Design Rules (DRC).

Same here, what is the changed behavior? And is this related to the trace dragging at all?

Also it looks like you somehow messed up rebasing or so, since your commit reverts tons of changes from the past few weeks (or even months). This needs to be cleaned up before I can start to review the code.

EDIT: Hmm it seems you made changes to the source ZIP of release 2.1.1 and then committed the whole folder. Unfortunately this is not how it works, you will have to commit only the *diff* to the current master branch.

PR #1876, opened 2026-08-22
Clarify usage of LLM in contribution guidelines & add PR template

ubruhin:
I don't want to forbid usage of LLMs for contributions, but I am tired of chatting with LLMs through GitHub issues/PRs and of AI slop in the diffs. Therefore I specified some rules about the usage of LLMs in the contribution guidelines. In addition, I added a PR template which (beside other things) contains a section to declare the usage of LLMs, as now required by those new rules.

Rendered previews:
- [CONTRIBUTING.md](https://github.com/LibrePCB/LibrePCB/blob/pr-template/CONTRIBUTING.md#use-of-aillmagents)
- [pull_request_template.md](https://github.com/LibrePCB/LibrePCB/blob/pr-template/.github/pull_request_template.md)

Any feedback welcome.

PR #1877, opened 2026-08-23
Add smart trace/angle snapping and global pad size extension

codedrifter22:
## Summary

This PR adds two related usability improvements to the board editor:

### 1. Intelligent angle/line snapping while dragging
When moving a trace or footprint, connected line segments and angles
now automatically adjust to clean, sensible angles — similar to how
KiCad handles interactive routing adjustments. This removes the need
to manually clean up angles after every move.

Holding **Ctrl** while dragging disables this behavior and falls back
to the original free-form drag, for situations where the automatic
adjustment is not desired.

### 2. Global pad size extension (design rules)
Added a new design rule option that lets the user globally increase
the size of all pads on a board by a configurable amount, instead of
having to edit pad sizes footprint by footprint. This is useful e.g.
when a PCB fabricator requires larger pads/clearances than what the
footprint library defines by default.

## Why this is useful
- Faster, less fiddly routing adjustments (angle snapping matches a
  workflow many users already know from KiCad).
- Avoids duplicating footprints just to get bigger pads for a specific
  fab/process — one global setting instead of dozens of manual edits.

## Testing
- Manually tested moving traces/footprints with and without Ctrl held.
- Manually tested setting a global pad extension value and verifying
  it applies to all pads on the board.

## Notes for maintainers
I'd love feedback on the general approach, especially regarding
whether the pad extension setting should affect the saved project
file format. Happy to adjust based on your review.

### DIFF
--- a/libs/librepcb/core/project/board/boarddesignrules.cpp
+++ b/libs/librepcb/core/project/board/boarddesignrules.cpp
@@ -55,6 +55,10 @@ BoardDesignRules::BoardDesignRules() noexcept
     mPadAnnularRing(UnsignedRatio(Ratio::fromPercent(25)),  // 25%
                     UnsignedLength(250000),  // 0.25mm
                     UnsignedLength(2000000)),  // 2mm
+    // pad copper expansion
+    mPadCopperExpansion(UnsignedRatio(Ratio::fromPercent(0)),  // 0%
+                        UnsignedLength(0),  // 0mm
+                        UnsignedLength(2000000)),  // 2mm
     // via annular ring
     mViaAnnularRing(UnsignedRatio(Ratio::fromPercent(25)),  // 25%
                     UnsignedLength(200000),  // 0.2mm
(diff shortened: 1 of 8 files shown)


ubruhin:
You are opening the same PR and posting the same comments over and over without taking my feedback into account. Suspecting heavy usage of LLM -> closing. Respect #1876 for further PRs. Thanks.

codedrifter22:
Sorry for the confusion in my previous replies — I did use an AI 
   assistant to help draft some of the responses, which clearly came 
   across as not properly addressing your feedback. That wasn't my 
   intent. If you're open to it, I'd like to actually go through your 
   comments one by one myself and address them properly. Let me know 
   if you'd reconsider reviewing this once I've done that.

ubruhin:
Thanks for clarifying. If you address my feedback, and respect the new guidelines from #1876, of course I am open to review your contributions.

PR #1879, opened 2026-08-27
BoardEditor: Add angle/line snapping while dragging traces

codedrifter22:
Automatically adjusts angles of connected line segments while dragging, similar to KiCad's interactive routing. Holding Ctrl disables this and falls back to free-form dragging.

## Description
/

While drawing, angle snapping worked correctly, but issues appeared when moving an existing trace. The angles didn't keep their degrees, and horizontal/vertical lines lost their alignment with the grid. This change adds simple angle/line adjustment while dragging traces or components. Holding Ctrl while dragging reverts to the original (free) behavior.
/

## Checklist

- [x] I am aware of the [contributing guidelines](/CONTRIBUTING.md)
- [x] I manually tested the changes extensively and am sure they work perfectly
  - [x] I tested on Windows
  - [ ] I tested on macOS
  - [ ] I tested on Linux
- [ ] I am sure the changes do not affect any other features (no regressions)

## Usage of AI/LLM/Agents

- [ ] No LLM was used at all to create this PR
- [ ] For research
- [x] For code/tests/documentation
- [ ] For something else (specify):
- [x] I reviewed **all** of the generated content in detail with my human
      brain and can confirm it has at least the same quality as if I had
      written it by myself

### DIFF
--- a/dev/fix_ci_format.sh
+++ b/dev/fix_ci_format.sh
@@ -1,0 +1,139 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+BRANCH="feature/angle-snapping"
+REMOTE="origin"
+
+# 1) ensure repo root
+REPO_ROOT="$(git rev-parse --show-toplevel 2>/dev/null || true)"
+if [ -z "$REPO_ROOT" ]; then
+  echo "Error: not inside a git repo. cd to your LibrePCB repo clone and retry."
+  exit 1
+fi
+cd "$REPO_ROOT"
+
+echo "Fetching remote..."
+git fetch "$REMOTE"
+
+echo "Checking out branch $BRANCH..."
+# create local branch if missing
+if git show-ref --verify --quiet "refs/heads/$BRANCH"; then
+  git checkout "$BRANCH"
+  git pull "$REMOTE" "$BRANCH" || true
+else
+  # try to track remote branch
+  if git ls-remote --exit-code --heads "$REMOTE" "$BRANCH" >/dev/null 2>&1; then
+    git checkout -b "$BRANCH" "$REMOTE/$BRANCH"
+  else
+    echo "Remote branch $REMOTE/$BRANCH not found. Abort."
+    exit 1
+  fi
+fi
+
+echo "Working on $(git rev-parse --abbrev-ref HEAD) @ $(git rev-parse --short HEAD)"
+
+CHANGES_MADE=0
+
+if [ -x "./dev/format_code.sh" ]; then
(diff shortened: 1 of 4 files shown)


ubruhin:
Hmm honestly, this still doesn't look mature to me. When I play around with this feature, I observe a lot of strange behavior. Thus I feel like this hasn't been tested really well yet. Just some random examples where it does strange things:

[Peek 2026-09-07 16-07.webm](https://github.com/user-attachments/assets/f25dd986-bd40-4ac0-9778-d40806be7183)

Or in some cases, it seems we get a huge performance issue, everything getting laggy:

[Peek 2026-09-07 16-10.webm](https://github.com/user-attachments/assets/302e7236-6d9d-4bab-9109-e5d0a85f57a4)

Another case where it doesn't do what I'd expect:

[Peek 2026-09-07 16-11.webm](https://github.com/user-attachments/assets/7dd3e3b8-aa29-4d6b-b1ec-130b2efad0ae)

Don't get me wrong - I'm not saying it is trivial to fix those things, or even to figure out what would be the expected behavior. I'm actually impressed about the cases where it does work well. However, as long as there is such strange behavior (which can be triggered so easy during normal usage) and performance issues, this is neither ready for merge, nor for a deeper code review.

Also, this PR again introduces changes which are not related at all to this feature (a new bash script, changing indentation of existing, otherwise untouched, code). If you just spend 5 seconds to review your own changes, you would immediately see those issues and I wouldn't have to spend 3 minutes to remind you about such simple things.

I'm sorry about this - we have high quality standards which can sometimes be frustrating for contributors, but that is required to keep LibrePCB a highly stable, reliable tool for everyone.

codedrifter22:
Hello, I apologize for the complications, but I forgot to mention a few
things.

The first thing is that my new feature only works if you select a single
section of the route, meaning just the line between two points. I wanted to
prevent the line from being moved beyond the point, and I also wanted
dragging the point itself to work only while holding CTRL, but I wasn’t
able to get that working.

I also haven’t completely managed to fix the issue where, when you drag a
component, only one line adjusts instead of the other lines adapting as
well.

I thought that, at least temporarily, this solution could still be
practical and useful, because it allows individual lines to be moved
separately with my improvement, even though it is not completely polished
yet.

Thank you for testing it and for telling me what could be improved. I’ll
try to finish and polish it properly.

Thank you!

codedrifter22:
Hello, I’ve fixed my mistakes. Could you please test it and see if everything works correctly? Thank you.

PR #1884, opened 2026-09-05
Fix compilation error when building without OpenCASCADE

### LINKED ISSUE #1883: Unit tests fail to build when USE_OPENCASCADE=OFF
```
LibrePCB Version: 2.1.2-unstable
Qt Version:       6.11.2 (built against 6.11.2)
Operating System: Windows 10 Version 22H2 (x86_64)
System Locale:    en_US (en-Latn-US, en-US)
Platform Plugin:  windows
TLS Library:      OpenSSL 3.3.0 9 Apr 2024
OCC Library:      N/A
```

##### SUMMARY
Unit tests fail to build when OpenCASCADE is not enabled (`USE_OPENCASCADE=OFF`).  This is due to the absolute inclusion of `core/3d/occmodeltest.cpp` in the `add_executable()` function in `~/tests/unittests/CMakeLists.txt`.

##### STEPS TO REPRODUCE
1) Add `-DUSE_OPENCASCADE=OFF` to the optional CMAKE declarations in the build configuration
2) Build

##### EXPECTED RESULTS
Successful build (with OCC missing)

##### ACTUAL RESULTS
An error along the lines of:
```
:-1: error: [tests\unittests\CMakeFiles\librepcb_unittests.dir\build.make:91: tests/unittests/CMakeFiles/librepcb_unittests.dir/core/3d/occmodeltest.cpp.obj] Error 1
```

##### ADDITIONAL INFORMATION
The failure actually occurs when the OCC model test file tries to `#include` an OCC header file (which won't exist when `USE_OPENCASCADE=OFF`).

One possible fix is to _conditionally_ include the OCC model test file in `CMakeLists.txt`.  Another option would be to put a `#if` guard around the entire body of the `occmodeltest.cpp` file.  Since the test file is moot when OCC is not included, the `CMakeLists.txt` solution seems the more robust approach.

AI Disclosure: Claude AI assisted in the identification of this bug.  This report was authored by a human.

AvetosDesign:
## Description

Unit tests fail to build when `USE_OPENCASCADE=OFF` due to unconditional inclusion of the `occmodeltest.cpp` file (which subsequently tries to `#include` OCC header files, which do not exist in a non-OCC build).  (See issue #1883).

This fix modifies `CMakeLists.txt` to _conditionally_ include the OCC model test file based on the build settings.

## Checklist

- [x] I am aware of the [contributing guidelines](/CONTRIBUTING.md)
- [ ] I manually tested the changes extensively and am sure they work perfectly
  - [x] I tested on Windows
  - [ ] I tested on macOS
  - [ ] I tested on Linux
- [ ] I am sure the changes do not affect any other features (no regressions)

## Usage of AI/LLM/Agents

Claude AI was used in the writing of the fix.

- [ ] No LLM was used at all to create this PR
- [ ] For research
- [x] For code/tests/documentation
- [ ] For something else (specify):
- [x] I reviewed **all** of the generated content in detail with my human
      brain and can confirm it has at least the same quality as if I had
      written it by myself

## Related Issues

Fixes #1883

### DIFF
--- a/libs/librepcb/core/3d/occmodel.cpp
+++ b/libs/librepcb/core/3d/occmodel.cpp
@@ -545,6 +545,14 @@ bool OccModel::isAvailable() noexcept {
   return (USE_OPENCASCADE != 0);
 }
 
+bool OccModel::hasRgbaSupport() noexcept {
+#if USE_OPENCASCADE
+  return (OCC_VERSION_HEX >= 0x070500);
+#else
+  return false;
+#endif
+}
+
 QString OccModel::getOccVersionString() noexcept {
   QString s = OCC_EDITION_NAME;
 #if USE_OPENCASCADE
--- a/libs/librepcb/core/3d/occmodel.h
+++ b/libs/librepcb/core/3d/occmodel.h
@@ -71,6 +71,7 @@ public:
 
   // Static Methods
   static bool isAvailable() noexcept;
+  static bool hasRgbaSupport() noexcept;
   static QString getOccVersionString() noexcept;
   static void setVerboseOutput(bool verbose) noexcept;
   static std::unique_ptr<OccModel> createAssembly(const QString& name);
--- a/tests/unittests/CMakeLists.txt
+++ b/tests/unittests/CMakeLists.txt
@@ -10,10 +10,17 @@ add_definitions(-DTEST_DATA_DIR="${CMAKE_CURRENT_SOURCE_DIR}/../data")
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
+# Conditionally-included tests (gated by build options)
+# [Claude AI assisted in authoring this solution]
+set(LIBREPCB_UNITTESTS_OPTIONAL_SOURCES)
+if(USE_OPENCASCADE)
+  list(APPEND LIBREPCB_UNITTESTS_OPTIONAL_SOURCES core/3d/occmodeltest.cpp)
+endif()
+
 # Main executable
 add_executable(
   librepcb_unittests
-  core/3d/occmodeltest.cpp
+  ${LIBREPCB_UNITTESTS_OPTIONAL_SOURCES}
   core/algorithm/airwiresbuildertest.cpp
   core/algorithm/netsegmentsimplifiertest.cpp
   core/applicationtest.cpp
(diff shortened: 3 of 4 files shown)


ubruhin:
Hi @AvetosDesign

Thanks for reporting this. Unfortunately, this is not the way I'd like to fix it since now the tests are not run at all anymore when OCC is disabled (before this issue was introduced, tests were run even without OCC). The root cause is actually very simple: #1755 added `#include <Standard_Version.hxx> ` to the test file, which should not be there. OCC headers shall only be included from `occmodel.cpp`, nowhere else.

I think the cleanest solution is to add a static method `OccModel::hasRgbaSupport()` right after `OccModel::isAvailable()`. Then this method can be used by the unit tests to determine whether RGBA support is available or not, without any kind of conditional compilation. Conditional compilation often ends up in a complex hell of a mess, thus I want to avoid it whenever possible.

Could you adjust the PR accordingly? You can just (force-)push more commits to this branch.

AvetosDesign:
Sure.  I can do that.  Do you mind if I ask a question, though?

> ...now the tests are not run at all anymore when OCC is disabled (before this issue was introduced, tests were run even without OCC).

I'm a little confused by this statement.  The only test I made conditional was the OCC test.  All the other tests should still run.  Are you saying that you expect the OCC test to run and pass with `USE_OPENCASCADE=OFF`?  Maybe there's something I don't understand about the code.

Regardless, I can respect the desire to avoid conditional compilation (it can get very messy).

ubruhin:
> Are you saying that you expect the OCC test to run and pass with USE_OPENCASCADE=OFF? 

Yes exactly. The `OccModel` class basically exposes `USE_OPENCASCADE` with its static method `isAvailable()`, which is used by the unit tests to determine the expected behavior (just have a look at those test cases). The `OccModel` class is always compiled (independent of `USE_OPENCASCADE`) so the corresponding tests shall be compiled & run as well.

AvetosDesign:
Don't accept this just yet.  I inadvertently added part of my space mouse work into the makefile.  I'll remove it.  Sorry.

AvetosDesign:
That should be clean now.  I apologize for the chaos.  Please let me know if something else needs to be adjusted.

ubruhin:
Thanks, I think this is perfect! :ok_hand: Let's see what CI says ;-)

ubruhin:
CI failure is not related to this change, I will merge now and fix CI afterwards.

Thanks for this contribution! :tada:

